PR #2, opened 2026-08-04
Add initial documentation and Fern configuration

### LINKED ISSUE #1: Add initial documentation and Fern configuration
## Summary

Add the initial `docs/` tree and `fern/` configuration derived from an internal project.

## Details

The documentation covers the full user journey:

- **Getting Started** — install, quick start, WorkloadRun quick start
- **Concepts** — architecture, catalog, workload types, health monitoring & remediation, goodput & bandwidth measurement, platform detection
- **How-to Guides** — certify a cluster, run a WorkloadRun, NCCL benchmarks, adaptive fault isolation, custom catalog entries, interpret results
- **CLI Reference** — `xcalctl` setup, certification, workloadrun commands
- **API Reference** — Certification, Workflow, Job, WorkloadRun, GoodputMeasurement, BandwidthMeasurement, LogProfile CRDs

Fern is configured to publish to `docs.nvidia.com/cluster-readiness-engine`.

## Known placeholders (will be updated as code migrates in)

- API group is currently `excalibur.nvidia.com/v1alpha1` — will change once the OSS API group is finalized
- CLI name is currently `xcalctl` — subject to change
- Install script URL assumes `NVIDIA/cluster-readiness-engine` release assets exist

pdmack:
## Summary

- Adds the full `docs/` tree (concepts, getting started, how-to guides, CLI reference, API reference) derived from an internal project
- Adds `fern/` configuration targeting `docs.nvidia.com/cluster-readiness-engine`
- API group (`cre.nvidia.com/v1alpha1`) and CLI name (`ncrectl`) updated to match the codebase
- CLI reference corrected against the actual implementation: missing command groups (`cluster`, `workflow`, `upgrade`), wrong flags removed, real flags added, correct defaults
- Catalog corrected: YAML-based entry structure, accurate domain/variant table (`nemotron5-8b`, `nemotron5-56b`, loopback variants, `dcgm-level4`)
- Getting-started and setup docs fixed: `--image-pull-secret` takes a GitHub token (not a secret name); `ncrectl setup reset` removes Kubeflow Trainer by default
- Custom catalog how-to rewritten for the YAML schema
- New pages: `cli-reference/cluster.md`, `cli-reference/workflow.md`

## Test plan

- [ ] Verify Fern renders locally with `fern docs dev`
- [ ] Confirm nav structure in `docs/index.yml` matches the page tree
- [ ] Validate CLI flag tables against `ncrectl --help` output once binary is buildable

Fixes #1


docs/how-to-guides/certify-a-cluster.md
```diff
@@ -0,0 +1,88 @@
...
+---
+{/* SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved. */}
+{/* SPDX-License-Identifier: Apache-2.0 */}
+
+
+## Before you begin
+
+- [Install ncrectl and set up the controller](../getting-started/install.md)
+- Confirm your kubeconfig points at the target cluster
+- Have an NGC API key available for image pulls
```
lalitadithya: this isn't required anymore, we can drop this since the images will all be public at launch

docs/how-to-guides/adaptive-fault-isolation.md
```diff
@@ -0,0 +1,38 @@
...
+4. This repeats until a single node (or domain boundary) is isolated.
+
+For an N-node pool this requires at most log₂(N) additional runs instead of N−1.
+
+## Enable
+
+```yaml
+spec:
+  options:
+    adaptiveFaultIsolation: true
```
dims: hallucination? i don't see `adaptiveFaultIsolation`

docs/api-reference/certification.md
```diff
@@ -0,0 +1,56 @@
...
+  categories:
+    - domain: communication
+      variant: nccl-all-reduce
+    - domain: training
+      variant: nemotron4-15b
+      options:
+        maxSteps: 50
+        nodesPerJob: 8
+  options:
+    adaptiveFaultIsolation: true
```
dims: hallucination? i don't see `adaptiveFaultIsolation`

docs/how-to-guides/adaptive-fault-isolation.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+When enabled, a failed `Certification` automatically triggers bisection on the affected node group.
```
dims: Not sure if we do `bisection`...

docs/getting-started/install.md
```diff
@@ -0,0 +1,71 @@
...
+
+- Kubernetes 1.28+ cluster with GPU nodes
+- `kubectl` configured and pointing at the target cluster
+- NVIDIA GPU Operator installed
+- Helm 3.x
+
+## Install the CLI
+
+```bash
+curl -sSL https://github.com/NVIDIA/cluster-readiness-engine/releases/latest/download/install.sh | bash
```
dims: we don't have `latest` yet. @lalitadithya Do you want to update release mechanism to do that? or can we switch to using an environment var for this?

docs/cli-reference/overview.md
```diff
@@ -0,0 +1,53 @@
...
+{/* SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved. */}
+{/* SPDX-License-Identifier: Apache-2.0 */}
+
+
+`ncrectl` is the command-line interface for the Cluster Readiness Engine. It handles installation, certification lifecycle, workload execution, and reporting — without requiring direct `kubectl` access for most operations.
+
+## Install
+
+```bash
+curl -sSL https://github.com/NVIDIA/cluster-readiness-engine/releases/latest/download/install.sh | bash
```
dims: same problem. `latest` or environment variable?

docs/concepts/health-monitoring-remediation.md
```diff
@@ -0,0 +1,78 @@
...
+Certification  status.categoryStatuses[].failedNodes  ← per-category list of failed nodes
+```
+
+### Failure reasons
+
+Each failed node entry carries a `reason`:
+
+| Reason | Meaning |
+|--------|---------|
+| `HardwareFailureDetected` | CEL health check detected an unhealthy or cordoned node during the run |
```
dims: Cordoned nodes do not reach this reason. Discovery filters them before a Job runs, and since PR #84 they appear in status.orchestration.excludedNodes and make the run INCOMPLETE.

docs/cli-reference/setup.md
```diff
@@ -0,0 +1,98 @@
...
+ncrectl setup init [flags]
+```
+
+### What it installs
+
+Runs two phases in order:
+
+| Phase | What |
+|-------|------|
+| `deps` | Kubeflow Trainer v2.2.0 |
```
dims: pkg/setup/setup.go:33 pins v2.2.1. PR #56 bumped it

pdmack:
@coderabbitai review

docs/getting-started/quick-start.md
```diff
@@ -0,0 +1,64 @@
...
+{/* SPDX-License-Identifier: Apache-2.0 */}
+
+
+This guide walks through a full cluster certification: install ncrectl, run the certification suite, and review the results.
+
+## Before you begin
+
+[Install ncrectl and set up the cluster](./install.md) before continuing.
+
+You will also need an NGC API key to pull the certification workload images.
```
kwombach12, line 17: should we link to docs on how to generate/get the key?

docs/getting-started/quick-start.md
```diff
@@ -0,0 +1,64 @@
+---
+title: Quick Start
+description: Certify a GPU cluster end-to-end in minutes.
+---
+{/* SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved. */}
+{/* SPDX-License-Identifier: Apache-2.0 */}
+
+
+This guide walks through a full cluster certification: install ncrectl, run the certification suite, and review the results.
```
kwombach12, line 9: I think we want to highlight the use case for a certification run so that we help users understand the 'certification run' vs 'workloadrun' (default tests vs custom workload)

docs/getting-started/workloadrun-quick-start.md
```diff
@@ -0,0 +1,57 @@
+---
+title: WorkloadRun Quick Start
+description: Run a distributed training or NCCL workload without writing a full Certification spec.
+---
+{/* SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved. */}
+{/* SPDX-License-Identifier: Apache-2.0 */}
+
+
+`WorkloadRun` is a simplified API for running a single workload against a set of nodes. It is useful for one-off benchmarks, smoke tests, and validating specific node groups without running the full certification suite.
```
kwombach12, line 9: similar to comment on certification quick start, I think we want a small blurb here on when/why I would need 'workloadrun' i.e. running a custom set of tests

docs/concepts/architecture.md
```diff
@@ -0,0 +1,52 @@
+---
+title: Architecture
```
kwombach12, line 2: should we include an arch diagram?

kwombach12 (dismissed):
added a couple thoughts to help polish this. wouldn't block on these

ndipebot (dismissed):

docs/README.md
```diff
@@ -0,0 +1,33 @@
+---
+title: Introduction
+description: NVIDIA Cluster Readiness Engine — GPU cluster certification, benchmarking, and hardware failure detection for Kubernetes.
+---
+{/* SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved. */}
+{/* SPDX-License-Identifier: Apache-2.0 */}
+
+
+The Cluster Readiness Engine is a Kubernetes controller for GPU cluster burn-in certification, orchestrated benchmarking, and hardware failure detection. Run real distributed workloads across topology-aware node groups, measure training throughput and interconnect bandwidth, detect hardware failures, and automatically quarantine bad nodes — before production workloads touch the cluster.
```
ndipebot: **Inaccurate: CRE never quarantines nodes.** CRE records which nodes failed and why in `status.categoryStatuses[].failedNodesRef`. It does not taint, cordon, or quarantine nodes — the Remediation controller was removed (ADR-061). External automation is expected to act on the recorded failure reasons.

docs/getting-started/quick-start.md
```diff
@@ -0,0 +1,64 @@
...
+  target:
+    nodeSelector:
+      nvidia.com/gpu.present: "true"
+  imagePullSecrets:
+    - name: ngc-secret
+  categories:
+    - domain: communication
+      variant: nccl-all-reduce
+    - domain: training
+      variant: nemotron4-15b
```
ndipebot: **Wrong catalog variant.** `nemotron4-15b` does not exist. The training catalog has `nemotron5-8b` and `nemotron5-56b`. This example will fail immediately when applied.

docs/getting-started/quick-start.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Review the results
+
+When the certification completes, ncrectl prints a pass/fail summary per node group and category. A full report is available with:
+
+```bash
+ncrectl certification report gpu-cluster-cert
+```
+
+Failed categories indicate nodes that did not meet performance thresholds. The controller automatically taints and cordons those nodes so production workloads avoid them.
```
ndipebot: **Inaccurate: CRE never taints or cordons nodes.** The controller records failed nodes in status but does not modify node state. See README comment — the same incorrect claim appears here.

docs/api-reference/logprofile.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Example
+
+```yaml
+apiVersion: cre.nvidia.com/v1alpha1
+kind: LogProfile
+metadata:
+  name: nemo-training
+spec:
+  patterns:
```
ndipebot, line 21: **Wrong spec schema.** `spec.patterns` is not a `[]Pattern` list. The actual type is `LogPatternSet` — a named struct with 8 optional `*EventPattern` fields: `trainingStep`, `checkpointSave`, `checkpointDone`, `checkpointRestore`, `checkpointLoaded`, `applicationStart`, `warmupStep`, `bandwidthResult`. Each `EventPattern` has `regex`, `example`, and `units` sub-fields. The YAML example below will not pass CRD validation.

docs/api-reference/logprofile.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Spec fields
+
+_Generated from CRD schema — coming soon._
+
+## Key spec fields (summary)
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `patterns` | []Pattern | List of named regex patterns with capture groups |
```
ndipebot: **Wrong spec table.** `patterns[].name`, `patterns[].regex`, and `patterns[].fields` do not exist. See above — the spec uses named struct fields, not an indexed list. Also: the required `spec.timestamp.layout` field is entirely absent from this document; omitting it fails CRD admission (`+kubebuilder:validation:Required`).

docs/api-reference/logprofile.md
```diff
@@ -0,0 +1,54 @@
...
+| `patterns[].fields` | []Field | Capture group names, types, and units |
+
+## Built-in profiles
+
+The following LogProfiles are installed by `ncrectl setup init`:
+
+| Name | Framework |
+|------|-----------|
+| `nccl-bandwidth` | NCCL test output (all collectives) |
+| `nemo-training` | NeMo framework training logs |
```
ndipebot: **Wrong built-in profile name.** The installed profile is `nemo-4-training`, not `nemo-training`. There are also 3 more built-in profiles not listed here: `megatron-training`, `megatron-bridge`, and `nccl-loopback`.

docs/api-reference/bandwidth-measurement.md
```diff
@@ -0,0 +1,36 @@
...
+_Generated from CRD schema — coming soon._
+
+## Key spec fields (summary)
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `jobRef` | ObjectRef | The Job to watch |
+| `logProfileRef` | string | Name of the LogProfile for NCCL output parsing |
+| `testType` | string | Collective type: `all_reduce`, `all_gather`, `alltoall` |
+| `minBusBandwidthGBps` | float | Minimum acceptable bus bandwidth in GB/s |
```
ndipebot: **Field does not exist.** `minBusBandwidthGBps` is not in `BandwidthMeasurementSpec`. Actual spec fields: `jobRef`, `logProfileRef`, `sampleInterval`, `testType`.

docs/api-reference/bandwidth-measurement.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Status fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `measuredBusBandwidthGBps` | float | Measured bus bandwidth |
```
ndipebot: **Fields do not exist.** `measuredBusBandwidthGBps` and `phase` are not in `BandwidthMeasurementStatus`. Actual status: `results []BandwidthResult`, `startTime`, `completionTime`, `conditions []metav1.Condition`.

docs/api-reference/goodput-measurement.md
```diff
@@ -0,0 +1,35 @@
...
+
+_Generated from CRD schema — coming soon._
+
+## Key spec fields (summary)
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `jobRef` | ObjectRef | The Job to watch |
+| `logProfileRef` | string | Name of the LogProfile defining parsing patterns |
+| `minGoodputRatio` | float | Minimum acceptable goodput (0.0–1.0); job fails if below this |
```
ndipebot: **Field does not exist.** `minGoodputRatio` is not in `GoodputMeasurementSpec`. Actual spec: `jobRef`, `logProfileRef`, `sampleInterval`.

docs/api-reference/goodput-measurement.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Status fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `goodputRatio` | float | Measured goodput ratio |
```
ndipebot: **Fields do not exist.** `goodputRatio` and `phase` are not in `GoodputMeasurementStatus`. Actual status uses `result string` and `conditions []metav1.Condition`.

docs/api-reference/job.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Spec fields
+
+_Generated from CRD schema — coming soon._
+
+## Status fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `phase` | string | `InProgress`, `Passed`, or `Failed` |
```
ndipebot: **Field does not exist.** `status.phase` is not in `JobStatus`. The resource uses `status.conditions []metav1.Condition` with condition types `JobInProgress`, `JobSucceeded`, `JobFailed`.

docs/api-reference/job.md
```diff
@@ -0,0 +1,36 @@
...
+| `conditions` | []Condition | InProgress, Succeeded, Failed (mutually exclusive) |
+| `workloadRef` | ObjectRef | Reference to the created workload (TrainJob, etc.) |
+| `goodputMeasurementRef` | ObjectRef | Reference to the GoodputMeasurement (if configured) |
```
ndipebot: **Fields do not exist.** `goodputMeasurementRef` and `bandwidthMeasurementRef` are not in `JobStatus`. Measurement resources reference the Job via `spec.jobRef`, not the reverse.

docs/api-reference/job.md
```diff
@@ -0,0 +1,36 @@
...
+| `bandwidthMeasurementRef` | ObjectRef | Reference to the BandwidthMeasurement (if configured) |
+| `failedNodes` | []string | Nodes identified as failed by health monitors |
```
ndipebot: **Wrong type.** `failedNodes` is `[]FailedNode` — a struct with `name string`, `reason string`, and `message string` fields — not `[]string`.

docs/api-reference/certification.md
```diff
@@ -0,0 +1,56 @@
...
+<Warning>
+`spec.categories` is **immutable** after the Certification is created (`self == oldSelf` validation). To change the category list, delete the Certification and recreate it. The minimum is 1 category.
+</Warning>
+
+## Status fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `conditions` | []Condition | InProgress, Succeeded, Failed (mutually exclusive) |
+| `categoryStatuses` | []CertificationCategoryStatus | Per-category status including `domain`, `variant`, `status`, and `failedNodes` |
```
ndipebot: **Field does not exist.** `categoryStatuses[].failedNodes` is not in `CertificationCategoryStatus`. The actual field is `failedNodesRef *corev1.TypedLocalObjectReference` — a reference to a ConfigMap containing the node list, not an inline list.

docs/api-reference/certification.md
```diff
@@ -0,0 +1,56 @@
...
+
+Each `categoryStatuses` entry includes a `failedNodes` list. Each entry in that list has a `name` (node name) and a `reason` (`HardwareFailureDetected`, `ThresholdViolation`, or `WorkloadFailed`).
```
ndipebot: **Inaccurate.** There is no inline `failedNodes` list per category. Nodes are stored in a ConfigMap referenced by `failedNodesRef`. To retrieve them, fetch the ConfigMap named in that reference.

docs/api-reference/workflow.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Spec fields
+
+_Generated from CRD schema — coming soon._
+
+## Status fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `phase` | string | `InProgress`, `Passed`, or `Failed` |
```
ndipebot: **Field does not exist.** `status.phase` is not in `WorkflowStatus`. Uses `status.conditions` with types `WorkflowInProgress`, `WorkflowSucceeded`, `WorkflowFailed`.

docs/api-reference/workflow.md
```diff
@@ -0,0 +1,35 @@
...
+| `conditions` | []Condition | InProgress, Succeeded, Failed (mutually exclusive) |
+| `iterationsCompleted` | int | Number of completed iterations |
```
ndipebot: **Wrong field name and path.** The field is `completedIterations` (not `iterationsCompleted`) and lives at `status.orchestration.completedIterations`, not at the top level.

docs/api-reference/workflow.md
```diff
@@ -0,0 +1,35 @@
...
+| `appliedOverrides` | string | Annotation listing which platform/GPU overrides matched |
```
ndipebot: **Wrong type and path.** `appliedOverrides` is `[]AppliedOverride` (a slice of structs), not a `string`. It also lives at `status.orchestration.appliedOverrides`, not at the top level.

docs/api-reference/workloadrun.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Spec fields
+
+_Generated from CRD schema — coming soon._
+
+## Status fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `phase` | string | `InProgress`, `Passed`, or `Failed` |
```
ndipebot: **Field does not exist.** `status.phase` is not in `WorkloadRunStatus`. Uses `status.conditions` with types `WorkloadRunSucceeded`, `WorkloadRunFailed`.

docs/api-reference/workloadrun.md
```diff
@@ -0,0 +1,48 @@
...
+| `conditions` | []Condition | InProgress, Succeeded, Failed (mutually exclusive) |
+| `bandwidthResult` | BandwidthResult | Measured bandwidth results (if configured) |
```
ndipebot: **Fields do not exist.** `bandwidthResult` and `goodputResult` are not in `WorkloadRunStatus`. Actual status fields include: `workflowRef`, `resolvedGpusPerNode`, `succeededNodesRef`, `failedNodesRef`, `detectedPlatform`, `detectedGPUArchitecture`.

docs/concepts/architecture.md
```diff
@@ -0,0 +1,52 @@
...
+|----------|---------|
+| `GoodputMeasurement` | Watches a Job's pod logs via a LogProfile, computes the goodput ratio |
+| `BandwidthMeasurement` | Parses NCCL log output, computes per-bus bandwidth metrics |
+| `LogProfile` | Cluster-scoped; defines regex patterns with named capture groups for log parsing |
+
+## Controller patterns
+
+- **`setExclusiveCondition()`** — enforces mutually exclusive InProgress / Succeeded / Failed conditions at every tier
+- **`Owns()` watches** — each tier watches its children for event-driven reconciliation; polling (15s in production, 1s in tests) is a safety net
+- **Adapter pattern** — the `Job` controller normalizes five training frameworks (TrainJob, MPI, PyTorch, etc.) to a `WorkloadPhase` via a `WorkloadAdapter` interface
```
ndipebot: **Inaccurate.** `pkg/workload/adapter.go` `ForSpec()` handles exactly one framework: `TrainJob`. MPI and PyTorch are user-facing input types on `WorkloadRun` (a separate resource) that get translated to TrainJob specs before the `Job` controller ever sees them. The `Job` controller's adapter only sees TrainJob.

docs/concepts/architecture.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Catalog
+
+The catalog maps `{domain, variant}` pairs to `WorkflowSpec` builders. Each entry is a Go file in `internal/catalog/entries/` that registers itself via `init()`. See [Concepts: Catalog](./catalog.md).
```
ndipebot: **Wrong path and file type.** Catalog entries are **YAML files** in `pkg/catalog/entries/<domain>/<variant>.yaml`, not Go files in `internal/catalog/entries/`. The `init()` registration is in `pkg/catalog/loader.go`, which embeds the YAML files via `go:embed`.

docs/concepts/workload-types.md
```diff
@@ -0,0 +1,38 @@
...
+
+
+The `Job` controller launches workloads via an adapter pattern that normalizes multiple training frameworks to a common `WorkloadPhase` (Running / Succeeded / Failed). The adapter is selected based on which field is set in the `WorkloadSpec`.
+
+## Supported frameworks
+
+| Framework | Spec field | Notes |
+|-----------|-----------|-------|
+| Kubeflow TrainJob | `trainJob` | Default for most catalog entries; requires Kubeflow Trainer |
+| MPI | `mpi` | Used for NCCL benchmark workloads via `mpirun` |
```
ndipebot: **Field does not exist.** `spec.mpi` is not a field on the `Job` API (`WorkloadSpec` in `job_types.go`). `WorkloadSpec` has exactly one field: `trainJob *trainerv1alpha1.TrainJobSpec`. MPI is a `WorkloadRun`-level concept that is translated to a TrainJob before Job creation.

docs/concepts/workload-types.md
```diff
@@ -0,0 +1,38 @@
...
+| PyTorch | `pytorch` | Directly launches `torchrun` |
```
ndipebot: **Field does not exist.** `spec.pytorch` is not a field in `WorkloadSpec`. See comment above — only `trainJob` exists.

docs/concepts/workload-types.md
```diff
@@ -0,0 +1,38 @@
...
+| Custom | `custom` | Arbitrary pod template; user-supplied entrypoint |
```
ndipebot: **Field does not exist.** `spec.custom` is not a field in `WorkloadSpec`. YAML using this field will be rejected by the API server.

docs/concepts/health-monitoring-remediation.md
```diff
@@ -0,0 +1,82 @@
...
+## Failed node attribution
+
+CRE does not taint, cordon, or patch nodes. Its job is to identify which nodes failed which tests and why. That signal is recorded on the Certification CR and is available for external systems (node lifecycle operators, alerting pipelines) to act on.
+
+### How failures propagate
+
+Failed nodes bubble up through the three-tier hierarchy:
+
+```
+Job            status.failedNodes              ← set when the Job fails, one entry per node with a reason
```
ndipebot: **Wrong field path.** `Workflow.status.failedNodes` does not exist. Workflow uses `status.orchestration` fields and `failedNodesRef *corev1.TypedLocalObjectReference` — a ConfigMap reference, not an inline list.

docs/concepts/health-monitoring-remediation.md
```diff
@@ -0,0 +1,82 @@
...
+      ↑ copied to
+Workflow       status.failedNodes              ← union across all failed Jobs for this category
+      ↑ copied to
+Certification  status.categoryStatuses[].failedNodes  ← per-category list of failed nodes
```
ndipebot: **Wrong field name.** `categoryStatuses[].failedNodes` does not exist. The actual field is `categoryStatuses[].failedNodesRef *corev1.TypedLocalObjectReference`.

docs/cli-reference/overview.md
```diff
@@ -0,0 +1,54 @@
...
+## Command groups
+
+| Command group | Purpose |
+|--------------|---------|
+| `ncrectl setup` | Install and uninstall the controller and its dependencies |
+| `ncrectl certification` | Run, render, report, and list-categories for Certification resources |
+| `ncrectl workloadrun` | Run, render, report, status, and cancel WorkloadRun resources |
+| `ncrectl cluster` | Inspect GPU nodes, platform, and network topology |
+| `ncrectl workflow` | Render Workflow manifests offline |
+| `ncrectl upgrade` | Upgrade ncrectl to the latest release |
```
ndipebot: **Command does not exist.** `ncrectl upgrade` is not registered. The registered top-level commands are: `certification`, `cluster`, `workflow`, `setup`, `workloadrun`.

docs/cli-reference/certification.md
```diff
@@ -0,0 +1,92 @@
...
+```
+
+### Flags
+
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--cert-file` | — | Path to a Certification YAML (mutually exclusive with `--category`) |
+| `--category` | — | Category in `domain/variant` format; repeatable (mutually exclusive with `--cert-file`) |
+| `--name` | auto | Certification name (default: `ncrectl-<timestamp>`) |
+| `--image-pull-secret` | — | GitHub token — creates a `ghcr.io` pull secret in the target namespace |
```
ndipebot: **Flag does not exist on this command.** `--image-pull-secret` is not registered on `certification run`. This flag exists on `setup init` and `workloadrun run` but not here. A fix is in progress at #167.

docs/how-to-guides/interpret-results.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Goodput results
+
+- **Goodput ratio** (0.0–1.0) — fraction of time the job was making useful forward progress
+- **Expected minimum** from the catalog entry
+- A ratio below ~0.95 suggests stalls, slow nodes, or framework overhead
+
+## Failed nodes
+
+Failed nodes are listed per category in `status.categoryStatuses[].failedNodes`, each with a `name` and `reason`:
```
ndipebot: **Wrong field path.** `status.categoryStatuses[].failedNodes` does not exist. The actual field is `failedNodesRef *corev1.TypedLocalObjectReference`. To list failed nodes, retrieve the ConfigMap referenced by `failedNodesRef` — the node list is stored there, not inline.

docs/how-to-guides/adaptive-fault-isolation.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Cross-boundary probing
+
+When the fault lies at a network domain boundary (e.g. two nodes connected via a spine switch), standard bisection may not isolate it. Cross-boundary probing extends the algorithm to test node pairs that span domain boundaries.
+
+_Configuration reference coming soon._
+
+## Reading the result
+
+After isolation the `Certification` status names the isolated node(s) in `status.categoryStatuses[].failedNodes`. See [Health Monitoring & Failed Node Attribution](../concepts/health-monitoring-remediation.md) for how to act on that data.
```
ndipebot: **Wrong field path.** `status.categoryStatuses[].failedNodes` does not exist. See interpret-results.md comment — the actual field is `failedNodesRef`.

docs/getting-started/workloadrun-quick-start.md
```diff
@@ -0,0 +1,57 @@
...
+
+```yaml
+apiVersion: cre.nvidia.com/v1alpha1
+kind: WorkloadRun
+metadata:
+  name: nccl-all-reduce
+spec:
+  image: nvcr.io/nvidia/pytorch:26.01-py3
+  framework:
+    mpi:
```
ndipebot, line 27: **Missing required field.** `spec.framework.mpi.mpirunPath` is marked `+kubebuilder:validation:Required` in the CRD (`workloadrun_types.go`). Applying this YAML will be rejected by the API server. Add `mpirunPath: /usr/mpi/bin/mpirun` (or the appropriate path for the container image).

docs/how-to-guides/nccl-benchmarks.md
```diff
@@ -0,0 +1,47 @@
...
+
+```yaml
+apiVersion: cre.nvidia.com/v1alpha1
+kind: WorkloadRun
+metadata:
+  name: nccl-all-reduce
+spec:
+  image: nvcr.io/nvidia/pytorch:26.01-py3
+  framework:
+    mpi:
```
ndipebot, line 34: **Missing required field.** Same issue as workloadrun-quick-start.md — `spec.framework.mpi.mpirunPath` is required. This MPI example will fail on `kubectl apply`.

docs/api-reference/index.md
```diff
@@ -0,0 +1,29 @@
...
+| [GoodputMeasurement](./goodput-measurement.md) | Namespaced | Log-based training throughput measurement |
+| [BandwidthMeasurement](./bandwidth-measurement.md) | Namespaced | NCCL bandwidth measurement |
+| [LogProfile](./logprofile.md) | Cluster-scoped | Regex patterns for log parsing |
+
+## Field reference
+
+The per-resource pages below are generated from the CRD OpenAPI schemas using [`crd-ref-docs`](https://github.com/elastic/crd-ref-docs). To regenerate locally:
+
+```bash
+make api-docs
```
ndipebot: **Make target does not exist.** `make api-docs` is not defined in the Makefile.

ndipebot (changes requested):
Inline comments below flag each accuracy issue with the specific line. See the summary review for the full context.

docs/README.md
ndipebot: CRE records failed nodes in status but never taints, cordons, or quarantines them — the Remediation controller was removed (ADR-061). External systems are expected to act on the recorded failure reasons. Remove "automatically quarantine bad nodes" from this description.

docs/getting-started/quick-start.md
ndipebot: This variant does not exist. The catalog has `nemotron5-8b` and `nemotron5-56b`. Update the example to use one of those.

docs/getting-started/quick-start.md
ndipebot: CRE does not taint or cordon nodes. It records which nodes failed and why in `status.categoryStatuses[].failedNodesRef`. Node remediation is the responsibility of the operator or external automation.

docs/api-reference/logprofile.md
ndipebot: The spec schema here is incorrect. `spec.patterns` is not a `[]Pattern` list — it is `LogPatternSet`, a named struct with 8 optional `*EventPattern` fields: `trainingStep`, `checkpointSave`, `checkpointDone`, `checkpointRestore`, `checkpointLoaded`, `applicationStart`, `warmupStep`, `bandwidthResult`. Each `EventPattern` has `regex`, `example`, and `units` fields. Also: `spec.timestamp.layout` is a required field not documented here — omitting it fails CRD admission validation.

docs/api-reference/logprofile.md
ndipebot: The installed profile name is `nemo-4-training`, not `nemo-training`. There are also 3 more built-in profiles not listed here: `megatron-training`, `megatron-bridge`, `nccl-loopback`.

docs/api-reference/bandwidth-measurement.md
ndipebot: `minBusBandwidthGBps` does not exist in `BandwidthMeasurementSpec`. Actual spec fields: `jobRef`, `logProfileRef`, `sampleInterval`, `testType`.

docs/api-reference/bandwidth-measurement.md
ndipebot: `measuredBusBandwidthGBps` and `phase` do not exist in `BandwidthMeasurementStatus`. Actual status: `results []BandwidthResult`, `startTime`, `completionTime`, `conditions []metav1.Condition`.

docs/api-reference/goodput-measurement.md
ndipebot: `minGoodputRatio` does not exist in `GoodputMeasurementSpec`. Actual spec: `jobRef`, `logProfileRef`, `sampleInterval`.

docs/api-reference/goodput-measurement.md
ndipebot: `goodputRatio` and `phase` do not exist in `GoodputMeasurementStatus`. Actual status uses `result string` and `conditions []metav1.Condition`.

docs/api-reference/job.md
ndipebot: These fields do not exist in `JobStatus`. Also: `failedNodes` (line 24) is `[]FailedNode` (a struct with `name`, `reason`, `message`), not `[]string`.

docs/api-reference/certification.md
ndipebot: `failedNodes` does not exist as an inline list. The actual field is `failedNodesRef *corev1.TypedLocalObjectReference` — a reference to a ConfigMap containing the node list, not an inline list of node objects.

docs/api-reference/workflow.md
ndipebot: Field name is wrong. Actual: `status.orchestration.completedIterations`. All these status fields live inside `status.orchestration.*`, not at the top level.

docs/api-reference/workloadrun.md
ndipebot: `bandwidthResult` and `goodputResult` do not exist in `WorkloadRunStatus`. Also: the MPI example earlier in this file is missing the required `mpirunPath` field (`+kubebuilder:validation:Required`). Applying that YAML will fail CRD admission. Add `mpirunPath: /usr/mpi/bin/mpirun` (or the appropriate path for the container image).

docs/concepts/architecture.md
ndipebot: Incorrect. `pkg/workload/adapter.go` `ForSpec()` handles exactly one case: `TrainJob`. The "five frameworks" description conflates the WorkloadRun CLI input types (torch/mpi/exec) with the Job-level adapter, which only ever sees TrainJob specs.

docs/concepts/architecture.md
ndipebot: Incorrect path. Catalog entries are YAML files in `pkg/catalog/entries/<domain>/<variant>.yaml`, not Go files in `internal/catalog/entries/`.

docs/concepts/workload-types.md
ndipebot: These Job spec fields do not exist. `WorkloadSpec` has exactly one field: `trainJob *trainerv1alpha1.TrainJobSpec`. YAML using `spec.mpi`, `spec.pytorch`, or `spec.custom` will be rejected by the API server. The `mpi`/`pytorch`/`custom` distinctions are WorkloadRun CLI-level concepts that get translated to TrainJob specs before being written to the CRD.

docs/concepts/health-monitoring-remediation.md
ndipebot: This field path does not exist. The actual field is `failedNodesRef *corev1.TypedLocalObjectReference` — a ConfigMap reference, not an inline node list — and it lives on Certification's `categoryStatuses[].failedNodesRef`, not as a direct `status.failedNodes` on Job or Workflow.

docs/how-to-guides/interpret-results.md
ndipebot: This jsonpath does not exist. The actual field is `failedNodesRef` — a ConfigMap reference. To get the node list, you need to first retrieve the ConfigMap referenced by `failedNodesRef`:
```bash
kubectl get certification <name> -o jsonpath='{.status.categoryStatuses[0].failedNodesRef.name}'
kubectl get configmap <ref-name> -o yaml
```

docs/getting-started/workloadrun-quick-start.md
ndipebot, line 27: This MPI example is missing the required `mpirunPath` field (`+kubebuilder:validation:Required`). Applying this YAML will fail CRD admission. Add `mpirunPath: /usr/mpi/bin/mpirun` (or the appropriate path for the container image) under `framework.mpi`.

docs/cli-reference/overview.md
ndipebot: `ncrectl upgrade` does not exist. Registered commands: `certification`, `cluster`, `workflow`, `setup`, `workloadrun`.

docs/cli-reference/certification.md
ndipebot: `--image-pull-secret` is not registered on `certification run`. This flag exists on `setup init` and `workloadrun run` but not on the certification subcommand.

ndipebot (dismissed):

pdmack:
Thanks @ndipebot — noted. We'll update `docs/cli-reference/certification.md` and `docs/cli-reference/workloadrun.md` with the three-flag expansion once #167 merges.

ndipebot (commented):
Went back through this now that the earlier comments are addressed — the CRD field/type fixes all hold up against the actual Go types and controller code, nice work closing those out.

While I was in there I found a few more accuracy issues (19 total, left as comments below), checked against the real source — CLI flags, catalog YAML, controller behavior. 8 of these would actually break something for someone following the docs: a couple of `ncrectl` commands that don't exist, a required field missing from a WorkloadRun example that would fail CRD admission, a catalog example that would fail to load, and an H200/EFA claim that isn't backed by any actual override in the catalog.

`docs/getting-started/first-certification.md` wasn't otherwise touched by this PR, so I couldn't attach comments directly to it. Two more there, for the record:

- line 18 — prerequisites say the controller only schedules on nodes without `nvidia.com/gpu.present`. Nothing in the Helm chart or setup flow enforces that (default affinity is empty, tolerations tolerate everything), so this could send someone chasing a requirement that isn't real.
- line 169 — the troubleshooting table repeats this as the root cause of a 5-minute setup hang. Worth re-checking what's actually causing that if it's from a real observed failure.

docs/concepts/goodput-bandwidth.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Goodput measurement
+
+Goodput is the fraction of elapsed time during which a training job is making useful progress (i.e. not stalled, restarting, or doing overhead work). A `GoodputMeasurement` resource watches a `Job`'s pod logs in real time, applies regex patterns from a `LogProfile`, and computes the ratio.
+
+### LogProfile
+
+A `LogProfile` (cluster-scoped) defines named regex capture groups that match log lines from a training framework. The goodput calculator uses the captured timestamps and step counts to determine when the job was making progress.
+
+Built-in LogProfiles are installed by `ncrectl setup init` for supported frameworks (NeMo, PyTorch Lightning, etc.).
```
ndipebot: PyTorch Lightning isn't actually supported — `grep -rni lightning` across the repo comes up empty. The built-in LogProfiles shipped in `helm/cluster-readiness-engine/templates/` are `nemo-4-training`, `megatron-training`, `megatron-bridge`, `nccl-bandwidth`, and `nccl-loopback`. None of them are for PyTorch Lightning, and there's no Lightning-specific parsing anywhere in `pkg/goodput`, `pkg/catalog`, or `pkg/controller`.

docs/api-reference/job.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Spec fields
+
+_Generated from CRD schema — coming soon._
+
+## Status fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `conditions` | []Condition | Mutually exclusive state: `InProgress`, `Succeeded`, `Failed`, `HardwareFailed`, `ValidationFailed` |
```
ndipebot: `HardwareFailed` and `ValidationFailed` aren't actually part of the mutually-exclusive set. `setExclusiveCondition` in `pkg/controller/job_controller.go` has a comment saying HardwareFailed is NOT part of this exclusive set — it's independent, and the exclusive-set slice only contains `JobInProgress`/`JobSucceeded`/`JobFailed`. `ValidationFailed` is set separately too (its comment calls it "additive, not exclusive"). Both can be `True` at the same time as `Succeeded`.

docs/api-reference/workflow.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Spec fields
+
+_Generated from CRD schema — coming soon._
+
+## Status fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `conditions` | []Condition | Mutually exclusive state: `InProgress`, `Succeeded`, `Failed`, `ValidationFailed` |
```
ndipebot: Same thing here — `ValidationFailed` isn't part of the mutually-exclusive set. `WorkflowValidationFailed`'s doc comment in `api/v1alpha1/workflow_types.go` says it's independent of the execution state conditions, and `setExclusiveCondition` in `workflow_controller.go` only ever includes `WorkflowInProgress`/`WorkflowSucceeded`/`WorkflowFailed`.

docs/api-reference/job.md
```diff
@@ -0,0 +1,45 @@
...
+| `workloadRef` | ObjectRef | Reference to the created workload (`TrainJob`) |
```
ndipebot: Small one — the type here should be `WorkloadReference`, not `ObjectRef`. `JobStatus.WorkloadRef` is declared as `*WorkloadReference` in `api/v1alpha1/job_types.go`, and there's no type named `ObjectRef` anywhere in the codebase.

docs/concepts/platform-detection.md
```diff
@@ -0,0 +1,70 @@
...
+## Auto-detection
+
+The controller detects two dimensions at runtime:
+
+| Dimension | Source |
+|-----------|--------|
+| Cloud platform | `spec.providerID` on the Node object (e.g. `aws://...`, `gce://...`) |
+| GPU architecture | `nvidia.com/gpu.product` node label (e.g. `NVIDIA-GB200`, `NVIDIA-H100-80GB-HBM3`) |
+
+Detection results are stamped on Workflow objects as annotations (`detected-gpu-architecture`, `detected-platform`) so you can verify what matched.
```
ndipebot: The live controller doesn't actually write these annotations. `detectPlatformConsistent`/`detectGPUArchConsistent` in `workflow_controller.go` only set `status.orchestration.detectedPlatform`/`detectedGPUArchitecture` — there's no `Annotations[...]` write in `workflow_controller.go` or `certification_controller.go`. The `detected-platform`/`detected-gpu-architecture` keys are only set client-side, in `pkg/render/render.go`'s `SetRenderAnnotations` (used by `ncrectl workflow render`), never during live reconciliation.

docs/api-reference/workloadrun.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Spec fields
+
+_Generated from CRD schema — coming soon._
+
+## Status fields
+
+| Field | Type | Description |
+|-------|------|-------------|
+| `conditions` | []Condition | Mutually exclusive state: `InProgress`, `Succeeded`, `Failed`, `ValidationFailed` |
```
ndipebot: `ValidationFailed` isn't mutually exclusive with `Succeeded`/`Failed` either. `WorkloadRunValidationFailed`'s doc comment says it's independent of the execution state conditions, and `setWorkloadRunCondition`'s exclusivity loop only covers `InProgress`/`Succeeded`/`Failed`. It's set separately and can be `True` alongside `Succeeded` — e.g. the workload finished fine but violated a performance threshold.

docs/api-reference/certification.md
```diff
@@ -0,0 +1,63 @@
...
+# Get the ConfigMap name from the category status
+kubectl get certification <name> -o jsonpath='{.status.categoryStatuses[0].failedNodesRef.name}'
+# Read the ConfigMap contents
+kubectl get configmap <ref-name> -o yaml
+```
+
+## Lifecycle
+
+1. Controller creates one `Workflow` per entry in `spec.categories`. The category list cannot be changed after creation — delete and recreate to modify it.
+2. Workflows run sequentially or in parallel depending on orchestration config.
```
ndipebot: Category Workflows never actually run in parallel — there's no config that enables it. `reconcileWorkflows` in `certification_controller.go` even says in a comment that it "ensures Workflows are created sequentially for each category," and `processNextCategory` only ever acts on the first non-terminal category per reconcile. `maxConcurrent` controls job/group parallelism *within* a single Workflow, not across categories.

docs/concepts/architecture.md
```diff
@@ -0,0 +1,76 @@
...
+[*] --> InProgress: Resource created
+InProgress --> Succeeded: Workload completed
+InProgress --> Failed: Workload failed
+Succeeded --> [*]
+Failed --> [*]
+```
+
+## Catalog
+
+The catalog maps `{domain, variant}` pairs to `WorkflowSpec` builders. Each entry is a YAML file under `pkg/catalog/entries/<domain>/<variant>/` that registers itself at startup. See [Concepts: Catalog](./catalog.md).
```
ndipebot: This doesn't quite match how it works, and it's inconsistent with `docs/concepts/catalog.md` in this same PR. Catalog entries are flat `pkg/catalog/entries/<domain>/<variant>.yaml` files (see `parsePath`'s comment in `loader.go`: `"entries/{domain}/{variant}.yaml"`), not files nested under a `<variant>/` directory. Registration is centralized too — `init()` walks the embedded FS and calls `Register()` for each file; entries don't register themselves. `catalog.md` already has the correct description — might be worth just matching it here.

docs/README.md
```diff
@@ -0,0 +1,33 @@
...
+# SPDX-License-Identifier: Apache-2.0
+---
+
+
+The Cluster Readiness Engine is a Kubernetes controller for GPU cluster burn-in certification, orchestrated benchmarking, and hardware failure detection. Run real distributed workloads across topology-aware node groups, measure training throughput and interconnect bandwidth, detect hardware failures, and record which nodes failed and why — before production workloads touch the cluster.
+
+## What it does
+
+- **Certifies GPU clusters** by running domain-specific workloads (NCCL benchmarks, distributed training) against all node groups and validating results against per-architecture thresholds
+- **Detects hardware failures** via pluggable node health monitors (CEL expressions, DCGM, custom detectors) running concurrently with workloads
```
ndipebot: DCGM isn't actually one of the pluggable node-health-monitor types. `NodeHealthMonitor` in `api/v1alpha1/job_types.go` has exactly one field, `CEL`. `pkg/nodemonitor` only has a `cel/` subpackage — there's no DCGM detector. DCGM shows up elsewhere in the repo as a diagnostics workload category (`dcgm-level4.yaml`), which is a different thing entirely. `docs/concepts/health-monitoring-remediation.md` in this same PR already gets this right — only lists CEL + custom detectors.

docs/cli-reference/overview.md
```diff
@@ -0,0 +1,50 @@
...
+
+
+`ncrectl` is the command-line interface for the Cluster Readiness Engine. It handles installation, certification lifecycle, workload execution, and reporting — without requiring direct `kubectl` access for most operations.
+
+## Install
+
+```bash
+export NCRECTL_VERSION=v0.1.0-rc.8
+curl -sSL "https://github.com/NVIDIA/cluster-readiness-engine/releases/download/${NCRECTL_VERSION}/installer" | bash
+ncrectl version
```
ndipebot: `ncrectl version` doesn't exist as a subcommand, so this will fail for anyone following the install steps. `cmd/ncrectl/main.go` only registers `certification`, `cluster`, `workflow`, `setup`, `workloadrun`. Running `ncrectl version` gives `Error: unknown command "version" for "ncrectl"`. The right command is `ncrectl --version`.

docs/getting-started/install.md
```diff
@@ -0,0 +1,76 @@
...
+export NCRECTL_VERSION=v0.1.0-rc.8
+curl -sSL "https://github.com/NVIDIA/cluster-readiness-engine/releases/download/${NCRECTL_VERSION}/installer" | bash
+```
+
+The installer automatically downloads and verifies a SHA-256 checksum before installing. On air-gapped systems, ensure `checksums.txt` from the same release is reachable alongside the binary.
+
+Verify the installation:
+
+```bash
+ncrectl version
```
ndipebot: Same problem as `cli-reference/overview.md` — `ncrectl version` fails. This is the doc's own verification step, so a first-time user hits an error right after installing. Should be `ncrectl --version`.

docs/cli-reference/overview.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+The installer also creates a `kubectl-ncre` symlink so the CLI is available as a kubectl plugin (`kubectl ncre ...`).
+
+<Warning>
+Release builds enforce a version check on every invocation. If a newer version exists, the command exits with code 1 and prints an upgrade prompt. Re-run the installer to update to a new version. CI pipelines should pin to a specific version to avoid unexpected failures.
```
ndipebot: This isn't actually implemented — it's describing an ADR (`docs/designs/052-forced-cli-upgrade.md`) that's still marked Proposed, not something that ships today. There's no `PersistentPreRunE`, no version-check HTTP call, and no `pkg/upgrade` package anywhere in the repo. Might be worth pulling this warning until it's actually built, or marking it as planned.

docs/cli-reference/workflow.md
```diff
@@ -0,0 +1,35 @@
...
+ncrectl workflow render [flags] <workflow.yaml>
+```
+
+### Flags
+
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--platform` | auto | Target platform (`aws`, `gcp`, `azure`, `oci`, `mistral`, `forge`) |
+| `--gpu-arch` | auto | Target GPU architecture (`h100`, `gb200`, `gb300`) |
+| `--nodes-file` | — | Path to a file listing node names (one per line) for topology-aware rendering |
```
ndipebot: `--nodes-file` actually expects a YAML file of full `corev1.Node` objects, not a plain list of names. `pkg/render/render.go` registers it as "Custom nodes YAML file," and `loadNodesFromFile` unmarshals straight into `[]corev1.Node`. A plain-text names-only file would fail to parse — see `pkg/render/testdata/render/custom-nodes-gcp-a100/input_nodes.yaml` for what it actually expects.

docs/cli-reference/workloadrun.md
```diff
@@ -0,0 +1,104 @@
...
+| `--controller-pull-secret` | — | Token for controller registry auth during `--setup` (e.g. GitHub PAT for `ghcr.io`) — separate from workload image credentials |
+| `--workload-registry` | — | Registry server for workload image pull (e.g. `nvcr.io`, `ghcr.io`) — required when `--workload-registry-password` is set |
+| `--workload-registry-username` | — | Registry username for workload image pull (e.g. `$oauthtoken` for NGC) — required when `--workload-registry-password` is set |
+| `--workload-registry-password` | — | Registry password or API key — creates an `ncrectl-pull-<name>` imagePullSecret in the namespace, deleted automatically when the WorkloadRun is deleted |
+| `--name` | — | Override the WorkloadRun name |
+| `--node-list` | — | Comma-separated list of nodes to target |
+| `--topology-domain` | — | Topology domain to target |
+| `--topology-key` | — | Node label key for topology grouping |
+| `--test-scale` | — | Override testScale (`intra-node`, `intra-rack`, `full-scale`) |
+| `--results-file` | — | Write results as JSON to this file path |
```
ndipebot: This is missing the `--wait` requirement that the equivalent flag on `certification run` mentions. The flag's own help text in `pkg/workloadrun/workloadrun.go` says "(requires --wait)", and the write only happens inside the `doWait`-gated branch — so `--results-file` without `--wait` silently writes nothing.

docs/how-to-guides/adaptive-fault-isolation.md
```diff
@@ -0,0 +1,43 @@
...
+```yaml
+spec:
+  categories:
+    - domain: communication
+      variant: nccl-all-reduce
+      options:
+        testScale: diagnose
+```
+
+The diagnose algorithm runs in four stages: intra-domain screening, inter-domain screening, bisection of failing groups, and confirmation of isolated suspects.
```
ndipebot: There are actually 7 diagnose stages in the code, not 4 — including a real, load-bearing one this doc skips entirely. `api/v1alpha1/workflow_types.go` defines `IntraScreening`, `IntraScreeningNoNVL`, `InterScreening`, `Bisection`, `Confirmation`, `CrossBoundary`, and `Complete`. ADR-055 documents the no-NVL screening round specifically (it separates fabric issues from NVLink issues), and the controller actively branches on it — it's just not mentioned here.

docs/how-to-guides/run-workloadrun.md
```diff
@@ -0,0 +1,67 @@
...
+apiVersion: cre.nvidia.com/v1alpha1
+kind: WorkloadRun
+metadata:
+  name: my-workload
+spec:
+  image: nvcr.io/nvidia/pytorch:26.01-py3
+  framework:
+    mpi:
+      binary: /usr/local/bin/all_reduce_perf_mpi
+      args: ["-b", "8", "-e", "32G", "-f", "2", "-n", "100"]
```
ndipebot, line 24: This example is missing the required `mpirunPath` field and would fail CRD admission as written. `MpirunPath` in `api/v1alpha1/workloadrun_types.go` is marked required with no `omitempty`, and the generated CRD lists `required: [binary, mpirunPath]`. The sibling examples in `workloadrun-quick-start.md` and `nccl-benchmarks.md` both include it — this one just needs the same line added.

docs/how-to-guides/run-workloadrun.md
```diff
@@ -0,0 +1,67 @@
...
+    logProfileRef: nccl-bandwidth
+    testType: all_reduce
+```
+
+## With goodput measurement
+
+```yaml
+spec:
+  goodputMeasurement:
+    logProfileRef: nemo-training
```
ndipebot: `nemo-training` isn't a name that actually ships — this reference wouldn't resolve. The real name is `nemo-4-training` (`helm/cluster-readiness-engine/templates/nemo4.yaml`, whose own header comment tells you to reference it as `"nemo-4-training"`). Worth noting though — even `nemo-4-training` isn't referenced by anything in the catalog today; the actual training entries use `megatron-training` instead. Might be simplest to swap this example to `megatron-training` rather than just fixing the typo.

docs/how-to-guides/certify-a-cluster.md
```diff
@@ -0,0 +1,97 @@
...
+
+```yaml
+spec:
+  target:
+    nodeSelector:
+      nvidia.com/gpu.product: NVIDIA-H200
+  enableMNNVL: false
+```
+
+H200 uses the same EFA configuration as H100 (8 GPUs per node, 8 MLNX per node).
```
ndipebot: There's no override that automatically applies H100's EFA config to H200 — this one isn't backed by anything in the catalog. `H200` shows up exactly once outside `docs/`, as a generic node-shape default in `gpu-defaults.yaml`, unrelated to AWS EFA resources. Every `gpuArchitecture` override condition matches `h100`/`gb200`/`gb300` only, never `h200`, and `ParseProduct("NVIDIA-H200")` returns a distinct `"h200"` string. As written, an H200 node wouldn't actually get the EFA resources this describes.

docs/how-to-guides/custom-catalog-entries.md
```diff
@@ -0,0 +1,130 @@
...
+    workload:
+      trainJob:
+        runtimeRef:
+          kind: TrainingRuntime
+          name: my-variant-runtime
+        trainer:
+          image: nvcr.io/nvidia/pytorch:26.01-py3
+          args:
+            - my-benchmark-command
+          numNodes: "{{ .NodesPerJob }}"
```
ndipebot: Quoting these template outputs as strings will actually break catalog loading. The rendered template gets unmarshaled straight into the typed `WorkflowSpec`, and `Trainer.NumNodes`/`NumProcPerNode` are `*int32` with no string-accepting unmarshaler — tried it, and `numNodes: "4"` fails with a `cannot unmarshal string into ... int32` error. Every real catalog entry uses the unquoted form, `numNodes: {{ .NodesPerJob }}` — worth matching that here.

ndipebot (changes requested):
Went through this again now that the earlier review threads are addressed — the CRD field/type corrections all hold up against the actual Go types and controller code, nice work closing those out.

Found a handful of new issues while I was in there, left as inline comments above (19 total). 8 of them would actually break something for someone following the docs — a couple of `ncrectl` commands that don't exist, a required field missing from a WorkloadRun example that would fail CRD admission, a catalog example that would fail to load, and an H200/EFA claim that isn't backed by any actual override in the catalog.

Two more couldn't get inline comments since that file isn't otherwise touched by this PR:

- `docs/getting-started/first-certification.md:18` — prerequisites say the controller only schedules on nodes without `nvidia.com/gpu.present`. Nothing in the Helm chart or setup flow enforces that (default affinity is empty, tolerations tolerate everything), so this could send someone chasing a requirement that isn't real.
- Same file, line 169 — the troubleshooting table repeats this as the root cause of a 5-minute setup hang. Worth re-checking what's actually causing that if it's from a real observed failure.

None of this should take long to fix, but requesting changes since a few of these would break the getting-started flow as written.

pdmack:
@ndipebot — all addressed, CI green. Summary of what the latest round fixed:

**Condition-set accuracy**
- `HardwareFailed` and `ValidationFailed` (Job), `WorkflowValidationFailed`, and `WorkloadRunValidationFailed` now documented as additive/independent — not part of the mutually-exclusive InProgress/Succeeded/Failed set
- `WorkloadRef` type corrected to `WorkloadReference`

**Controller behavior**
- Platform/GPU detection annotations clarified as client-side only (`ncrectl workflow render` / `SetRenderAnnotations`) — live controller writes to `status.orchestration.*`, not annotations
- Workflows per Certification now documented as sequential, not parallel; `maxConcurrent` scoped to within-Workflow job parallelism
- DCGM moved out of NodeHealthMonitor types — it's a catalog workload category, not a detector (only CEL exists)

**CLI accuracy**
- `ncrectl version` → `ncrectl --version` (install.md and overview.md)
- Forced-upgrade warning removed (ADR-052 is Proposed, nothing implemented)
- `--nodes-file` documented as full `corev1.Node` YAML, not a name list
- `--results-file` on `workloadrun run` now notes it requires `--wait`

**Examples**
- MPI WorkloadRun in `run-workloadrun.md` now includes required `mpirunPath`
- `nemo-training` → `megatron-training` in LogProfile examples
- H200/EFA claim removed — no `h200` override exists in the catalog
- Custom catalog template values unquoted to avoid `int32` unmarshal failure
- Adaptive fault isolation updated to 7 diagnose stages including `IntraScreeningNoNVL`
- PyTorch Lightning removed from goodput/bandwidth docs

**Upstream syncs also caught**
- `nemo-4-training` removed from all docs after #173 deleted the Helm template
- `nvidia.com/gpu.present` scheduling prerequisite removed from `first-certification.md` after #170 emptied the controller affinity; related troubleshooting rows updated

Thanks for the thorough second pass.

ndipebot (approved):
/lgtm
/approve

PR #21, opened 2026-08-06
docs: add a first-certification getting started guide

dims:
## Summary

This PR adds `docs/getting-started/first-certification.md`, an end-to-end guide for a new user: install the CLI, check the cluster, install CRE, run one catalog category, read the report, and clean up. It ends with a troubleshooting table for the nine most likely first-run failures.

Every command and claim was verified against the code on main:

- The install command uses the `installer` asset with `-p`, and the plugin name is `kubectl ncre`.
- `certification run` is shown without `--image-pull-secret`, which df413a2 removed. The workload images are public.
- The prerequisites include three that no other document states: the controller needs a node without GPUs, the chart requires the Prometheus Operator CRDs (the ServiceMonitor has no opt-out), and the GPU Operator must exist before setup.
- The failure section says CRE records failed nodes with a reason and does not modify them, which matches ADR-061 and the manager ClusterRole.
- The report section explains the namespace behavior: `run` creates `ncrectl-<timestamp>` and `report` defaults to `default`, so `-n` is required.

The file follows the conventions of the docs structure that #2 introduces (a `docs/getting-started/` directory, MDX frontmatter, SPDX comments), so the two changes compose. When #2 lands, this page needs one nav entry in `docs/index.yml`.

## Type of Change

Documentation. No code changes.

## Testing

- The markdown renders correctly in preview.
- All relative links point to files that exist on main (ADR-001, ADR-002, ADR-050, ADR-059, README).

## Risk

Low. One new file.

## Summary by CodeRabbit

* **Documentation**
  * Added a complete guide for completing your first certification.
  * Covers prerequisites, CLI setup, cluster validation, CRE installation, NCCL certification, report interpretation, cleanup, troubleshooting, and next steps.

### DIFF
--- a/docs/getting-started/first-certification.md
+++ b/docs/getting-started/first-certification.md
@@ -1,0 +1,174 @@
+---
+title: Your first certification
+description: Install CRE on a GPU cluster, run one certification category, read the report, and clean up.
+---
+
+{/* SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved. */}
+{/* SPDX-License-Identifier: Apache-2.0 */}
+
+# Your first certification
+
+This guide takes you from an empty GPU cluster to a completed certification report. You install the `ncrectl` CLI, check that your cluster is a valid target, install CRE, run one communication test, and read the result. Plan for 30 to 60 minutes. Most of that time is the workload itself.
+
+## Before you start
+
+You need:
+
+- A Kubernetes cluster with NVIDIA GPU nodes. Every target GPU node must carry the same `nvidia.com/gpu.product` label value. The [NVIDIA GPU Operator](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/) provides these labels. CRE does not install the GPU Operator.
+- At least one node without GPUs. The CRE controller schedules only on nodes that do not have the `nvidia.com/gpu.present` label.
+- `kubectl` access with permission to create CRDs, cluster roles, and namespaces. The setup step needs this. Later certification runs need less.
+- `helm` on your PATH. The setup step calls it.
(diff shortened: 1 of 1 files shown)


docs/getting-started/first-certification.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+`ncrectl` uses your current kubeconfig context. Pass `--kubeconfig` or `--context` on any command to point somewhere else.
+
+## Step 2: check your cluster
+
+Run the preflight commands before you install anything:
+
+```bash
+ncrectl cluster info
```
lalitadithya: can we document use of the kubectl plugin to match what is used in the readme? it would be good to keep it consistent

docs/getting-started/first-certification.md
```diff
@@ -0,0 +1,174 @@
...
+What happens:
+
+1. The CLI discovers the GPU nodes and prints the detected product.
+2. It creates a namespace named `ncrectl-<timestamp>` and a Certification in it. **Note both names from the output.** You need them later.
+3. The controller creates one Workflow for the category, the Workflow creates a Job, and the Job runs the NCCL test through Kubeflow Trainer across all target nodes.
+4. With `--wait`, the CLI prints a status line on every change and a heartbeat every 15 seconds, then prints the report.
+
+The default `--timeout` is 30 minutes. The command above raises it to 60. Training categories need more; give them hours, not minutes.
+
+## Step 5: read the report
```
lalitadithya, line 117: we have a report command that will show a nice table, should we use that instead here?

dims:
Both addressed in 368aef9.

- Commands now use `kubectl ncre`, matching the README. The install step still shows both forms once and says the guide uses the plugin form from there on.
- Step 5 now leads with `kubectl ncre certification report <name> -n <namespace>`, so the reader can print the table on demand rather than only reading what `--wait` printed. The namespace warning moved up next to the command, since that is where people hit it.

dims:
Rebased on latest main.

/ok to test dce54d734bc10a2810800e01bf230d44b0b82199

docs/getting-started/first-certification.md
dims: Done in dce54d73.

docs/getting-started/first-certification.md
dims: Done in dce54d73

lalitadithya (approved):

PR #27, opened 2026-08-06
test: remove the e2e scaffolding and vet the uat suite

dims:
## Summary

Removes the e2e test suite, per review feedback on this PR.

The suite was kubebuilder scaffolding. It held two smoke tests, one asserting the controller pod reaches Running and one asserting the metrics endpoint serves, and no business logic. `test/e2e/e2e_test.go` still carried the generated `// TODO: Customize the e2e test suite with scenarios specific to your project`.

It also did not compile:

```
$ go vet -tags=e2e ./test/e2e/
vet: test/e2e/e2e_suite_test.go:47:14: undefined: utils.LoadImageToKindClusterWithName
```

It called seven helpers that `test/utils` never defined. The initial repository squash dropped the functions and kept the constants they used. Nobody noticed because no CI job compiles tagged test code: `make test-ci` skips it, and the UAT workflow compiles only `./test/uat/`.

The UAT suite now covers this ground properly, with a real Kind cluster, a real Helm install, and real Kubeflow Trainer and JobSet controllers.

## What changed

- Delete `test/e2e` and `test/utils`, and the four `*-test-e2e` make targets.
- Move ginkgo and gomega to indirect in `go.mod`, and drop `ginkgolinter` from `.golangci.yml`, since no code uses them now.
- `make vet` also vets the `uat` tagged package, so tagged code cannot rot unnoticed again. The CI Build job runs vet through `make build`, which makes this a required check.

Net: 605 lines removed, 4 added.

## Type of Change

Test code and build configuration. No controller changes.

## Testing

- `make vet` passes and now covers both invocations.
- `make lint` reports 0 issues.
- `go build ./...` and the unit tests pass.

## Note

ADR-049 considered replacing these tests and chose to keep them, on the grounds that deployment health and metrics are a separate concern from UAT. That call predates the discovery that the suite never compiled. Worth a short amendment to ADR-049 if we want the record to match; happy to add it here or in a follow-up.

One thing this does give up: the metrics spec was the only test that exercised the HTTPS `:8443/metrics` endpoint with its RBAC. Integration and UAT both read metrics from the in-process registry instead. If we want that covered, the natural home is the UAT suite, which already has a live cluster.

## Summary by CodeRabbit

- **Testing**
  - Validation now includes both default and UAT-tagged test packages.
  - Removed the legacy end-to-end test suite and its setup, execution, and cleanup commands, simplifying the available test workflow.

- **Chores**
  - Updated linting configuration by removing an obsolete linter.
  - Cleaned up unused development dependencies and test utilities.
  - Streamlined project maintenance configuration without changing application functionality.


lalitadithya:
@dims , do you think we can just drop the e2e tests since that is just scaffolding and doesn't container any business tests right now?

dims:
@lalitadithya let me convert into draft, digging into something, will come back to this.

dims:
@lalitadithya agreed, dropped them in 9a131eb. The PR now deletes `test/e2e` and `test/utils` instead of restoring the helpers, and I updated the title and description to match.

Also removed along with them: the four `*-test-e2e` make targets, and ginkgo, gomega, and ginkgolinter, since nothing uses them now. 605 lines out, 4 in.

I kept one piece of the original change: `make vet` now also vets the `uat` tagged package. That is what would have caught this in the first place, since no CI job compiles tagged test code today. The Build job runs vet, so it is a required check.

Two things worth flagging for the record:

- ADR-049 explicitly considered replacing these tests and decided to keep them. That call predates knowing the suite never compiled. Happy to add a short amendment here or in a follow-up.
- The metrics spec was the only test that hit the HTTPS `:8443/metrics` endpoint with its RBAC. Integration and UAT both read from the in-process registry. If we want that path covered, UAT is the natural home since it already has a live cluster.

dims:
Rebased on latest main.

/ok to test adfa4b76b4ad7e6362315804143452af00906530

dims:
> @dims , do you think we can just drop the e2e tests since that is just scaffolding and doesn't container any business tests right now?

 Agreed, dropped.

lalitadithya (approved):

PR #28, opened 2026-08-06
fix: let the installer pin a version and fix the release install commands

dims:
## Summary

The install command in every release body cannot work today.

```bash
curl -sSL .../releases/latest/download/installer | bash
```

Without `-p`, the installer resolves the version through `gh release view` or the `/releases/latest` API. Both skip pre-releases. Every release so far (`v0.1.0-rc.1` through `v0.1.0-rc.7`) is a pre-release, so the command fails to find a version and exits. Anyone copying it from a release page gets an error.

The second command in the same section has a deeper problem:

```bash
curl -sSL .../releases/download/<tag>/installer | bash
```

This downloads the installer script from a specific tag, but the installer always resolved the version itself and had no way to accept one. So the "install this exact version" command installed whatever the newest release was, not the requested tag. It was silently wrong rather than broken.

The fix:

- Add `-v TAG` to the installer. When set, the installer skips version resolution and installs that tag. The usage text documents it.
- Release notes now pass `-p` for the newest release, with a comment explaining that every release is a pre-release until v1.0.0, and `-v <tag>` for the exact-version command.

## Type of Change

Bug fix. Installer script and the release-notes template; no controller code.

## Testing

Ran against the real release:

```
$ bash installer -v v0.1.0-rc.7 -d /tmp/inst-test
OS:   darwin
Arch: arm64
Using requested version v0.1.0-rc.7
Installing ncrectl v0.1.0-rc.7
Downloading ncrectl-darwin-arm64...
ncrectl installed to /tmp/inst-test/ncrectl

$ /tmp/inst-test/ncrectl --version
ncrectl version v0.1.0-rc.7
```

The `kubectl-ncre` symlink is created as expected. `bash -n installer` passes and the workflow YAML parses.

## Risk

Low. `-v` is additive and off by default, so the existing resolution path is unchanged. The release-notes change only affects the text of future release bodies.

## Note

Existing release bodies keep the broken command; only releases cut after this merges get the fix. Worth editing the `v0.1.0-rc.7` body by hand if people are using it now.

## Summary by CodeRabbit

* **New Features**
  * Added support for installing a specific release version with the `-v` option.
  * Added support for installing pre-release versions with the `-p` option until v0.1.0.

* **Bug Fixes**
  * Improved release selection and platform-specific installer downloads for more reliable installations.

* **Documentation**
  * Clarified authentication requirements and installation instructions for exact releases and pre-releases.


.github/workflows/release.yml
```diff
@@ -129,14 +129,16 @@ jobs:
             ## Install ncrectl
 
             ```bash
-            # Requires gh CLI authenticated, or GITHUB_TOKEN set
-            curl -sSL https://github.com/NVIDIA/cluster-readiness-engine/releases/latest/download/installer | bash
+            # Requires gh CLI authenticated, or GITHUB_TOKEN set.
+            # -p includes pre-releases. Every release is a pre-release
+            # until v1.0.0, so keep -p until then.
```
lalitadithya: until v0.1.0 right?

```suggestion
            # until v0.1.0, so keep -p until then.
```

dims:
Added a second fix to this branch, found while installing on a real cluster.

**The installer installed itself as `ncrectl` on any machine without the `gh` CLI.**

The API fallback picked the first asset url in the release JSON and never matched the platform binary name:

```bash
grep -o '"url" *: *"...releases/assets/[^"]*"' | grep -v "upload_url" | head -1
```

Assets are sorted by name, so `installer` always won over `ncrectl-linux-amd64`. The name-matching block below it was unreachable, because it only ran when the first lookup returned nothing.

The failure is silent. The script prints a green "installed successfully" and leaves a 9 KB shell script at `/usr/local/bin/ncrectl`. The next command fails with `illegal option -- -` and the installer's own usage text.

This affects servers, CI runners, and fresh clusters, which is where people install from a script rather than with `gh` present.

Verified on a fresh Ubuntu 24.04 host with no `gh` CLI, installing from this repo:

```
before: /usr/local/bin/ncrectl: Bourne-Again shell script, 9126 bytes
after:  /usr/local/bin/ncrectl: ELF 64-bit LSB executable, x86-64, statically linked
        ncrectl version v0.1.0-rc.7
```

dims:
Applied in 0328a7b. You are right, pre-releases end at v0.1.0.

dims:
Rebased on latest main.

/ok to test e4586bac4a9ee5982c3960b67f50bc5d2140d9e6

.github/workflows/release.yml
dims: Applied in e4586bac

lalitadithya (approved):

PR #31, opened 2026-08-06
fix: parse pod log timestamps from nodes that are not on UTC

dims:
## Summary

CRE reports "no bandwidth data" on any cluster whose nodes are not set to UTC, even when the workload wrote a full result table. Nothing logs an error.

Kubelet prefixes each log line with an RFC3339Nano timestamp. The zone is `Z` on a UTC node and an offset such as `-07:00` on any other node, which makes the prefix 35 characters instead of 31.

`stripK8sTimestamp` only searched for the separating space between positions 20 and 31:

```go
for i := 20; i < len(line) && i < 32; i++ {
```

So it returned the line with the prefix still attached. The bandwidth regex is anchored with `^\s*(?P<size>\d+)`, so it then tried to read `2026` as the size, hit the `-`, and failed. Every data line was skipped silently. The goodput timestamp regex ended in a literal `Z` and missed the same lines.

## How it was found

A single-GPU `nccl-loopback` certification on an A100 cluster whose nodes run in PDT. The workload succeeded and wrote a complete table:

```
2026-08-06T09:30:31.128034512-07:00   1073741824   268435456   float   sum   -1   1558.51   688.95   0.00   0 ...
```

The BandwidthMeasurement completed with `reason: JobSucceeded` and an empty results list, and the report printed "no bandwidth data".

## The fix

- `stripK8sTimestamp` cuts at the first space instead of assuming a maximum prefix length, so it handles both zone forms and any fractional-second precision.
- The goodput timestamp regex accepts `Z` or a numeric offset.

## Type of Change

Bug fix.

## Testing

- New test `TestParseBandwidthLogsNonUTCNode` uses the real log lines above, including a positive offset. It fails on main with `expected 3 results, got 0` and passes with this change.
- Existing tests still pass; they only used `Z` timestamps, which is why this was never caught.
- `make lint` reports 0 issues.

## Why no test caught this

The integration fixtures are hand-written with `Z` timestamps, and the UAT suite runs on KWOK nodes, which run no containers and produce no logs. The parsers had never seen output from a real kubelet on a non-UTC node.

## Risk

Low. The change only widens what the parsers accept.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kubernetes log timestamp parsing for UTC and numeric timezone offsets.
  * Added support for timestamps with variable-length fractional seconds.
  * Improved NCCL bandwidth log parsing for nodes using non-UTC time zones.
  * Preserved accurate parsing of logged sizes and bandwidth values across supported timestamp formats.
  * Improved reliability when processing A100 loopback and multi-node all-reduce bandwidth logs, including zero-bandwidth results and varied performance measurements.


dims:
Added the real pod log as a test fixture.

`pkg/nccl/testdata/real_a100_loopback_pdt.log` is the output of an `nccl-loopback` certification on an A100 node set to PDT, exactly as `kubectl logs --timestamps` returns it: the NCCL banner, the table header, 56 data rows across 28 message sizes and two cycles, and the footer.

`TestParseRealA100LoopbackLog` parses it and checks the first and last rows. It fails on main and passes with this change, so the parser now has a regression test against real hardware output rather than hand-written lines.

One detail the fixture pins down: the last row reads `688.95` algBW and `0.00` busBW. A single-rank all-reduce gives a bus bandwidth of zero, because the value scales by 2*(n-1)/n. The parser must still record the row, and the test asserts that.

dims:
Rebased on latest main.

/ok to test d52b885cce730b3cf683069d1871221ef5f801a5

dims:
Reproduced this end to end on real hardware, not just in a fixture.

Ran a two node NCCL all-reduce certification on a cluster with A100 nodes set to PDT. The run passed and printed 24 rows of real bandwidth data. The `BandwidthMeasurement` recorded **0 of the 24 rows**, and the results file held no bandwidth data at all:

```
=== bandwidth table (launcher log) ===
2026-08-06T14:10:27.180224784-07:00   256    64  float sum -1  583.79 0.00 0.00 ...
2026-08-06T14:11:42.931043300-07:00  67108864 16777216 float sum -1 2649049 0.03 0.03 ...
# Avg bus bandwidth : 0.0119653

=== BandwidthMeasurement status ===
results count: 0
keys: [completionTime, conditions]

=== results file ===
"result": "PASSED"
```

Two points this adds:

1. The failure is silent. There are no numbers, not zero valued numbers. Without a threshold, a run with no measurement data looks the same as a good run. The user reads `PASSED`.
2. The same happened on a single node loopback run in the same session, so it is not specific to one entry.

Added the captured launcher log as a second fixture in 0d1f22a. It covers what the loopback fixture cannot: loopback has one rank, so its bus bandwidth column is 0 on every row, while two ranks give a value above 0 that the test can assert.

Both fixtures return 0 rows against the old parser and the full count against the new one:

```
=== with OLD parser ===
--- FAIL: TestParseRealA100LoopbackLog: expected 56 results, got 0
--- FAIL: TestParseRealA100AllReduce2NodeLog: expected 24 results, got 0
=== fix restored ===
ok  github.com/NVIDIA/cluster-readiness-engine/pkg/nccl
```

lalitadithya (approved):

PR #37, opened 2026-08-07
docs: add an index of the architecture decision records

dims:
## Problem

`docs/designs/` holds 58 records and no index. The only way to find one is to read the file names. `README.md` and `AGENTS.md` both send readers here, and both assume you already know which record you want.

## Change

Adds `docs/designs/README.md`: a table of all 58 records, each title linked to its file, in file-name order. GitHub renders a directory's `README.md` as its landing page, so every existing link to `docs/designs/` now arrives at the index. No existing link changes.

**No record is edited.** The index reports what is there.

## What the index records

The number in the table is the one each record gives itself in its own heading. Three records give themselves a number their file name does not match:

| Record calls itself | File |
|---|---|
| ADR-0022 | `000-adr.md` |
| ADR-026 | `027-kustomize-override-ux.md` |
| ADR-027 | `031-platform-aware-nccl-config.md` |

So `ADR-027` reaches a different record depending on whether you follow the heading or the file name. The two records that cite the override record disagree for the same reason: `032-orchestration-overrides.md` calls it ADR-027 and `031-platform-aware-nccl-config.md` calls it ADR-026, and both mean `027-kustomize-override-ux.md`.

The README carries that lookup table and the explanation, so a reader who searches by either number lands somewhere useful. The three rows are also marked in the main table.

Twelve numbers are unclaimed by any heading: 000, 006, 011, 020, 028, 029, 030, 031, 033, 036, 037 and 040. The README notes that `README.md` and `AGENTS.md` describe the set as "ADR-000 to ADR-069" and name `000-adr.md` as the architecture record, both of which read by file name rather than by heading.

I checked the history: none of the unclaimed numbers ever belonged to a record, so they are not deletions.

## Alternative considered

Aligning the three headings to their file names instead. That removes the ambiguity outright and is a five-line change, but it rewrites the identity of records that other documents already cite. Reporting the current state was preferred. If you would rather fix the headings, say so and I will follow up; the index regenerates from whatever the headings say.

## Verification

- Every one of the 58 rows re-read against its file's `# ADR-NNN:` heading; all agree.
- All 58 links resolve.
- Every record file appears exactly once.
- `make verify` passes and leaves no generated file dirty.

### DIFF
--- a/docs/designs/README.md
+++ b/docs/designs/README.md
@@ -10,7 +10,6 @@ Read the relevant record before you change the behaviour it describes. `CLAUDE.m
 ## Conventions
 
 - The file name carries the number, and each record's heading matches it.
-- The numbers are not contiguous. Twelve are unused: 006, 011, 020, 022, 026, 028, 029, 030, 033, 036, 037, 040. No record ever held them.
 - ADR-000 is the full architecture record. ADR-001 is an abridged version of the same
   decision, kept for readers who want the short form.
 - New records follow the structure in `CLAUDE.md`: Context, Decision, Implementation,
--- a/docs/designs/README.md
+++ b/docs/designs/README.md
@@ -9,39 +9,18 @@ Read the relevant record before you change the behaviour it describes. `CLAUDE.m
 
 ## Conventions
 
-- The number below is the one the record gives itself in its own heading.
-- The rows are in file-name order, which is the order the directory lists them in.
+- The file name carries the number, and each record's heading matches it.
+- The numbers are not contiguous. Twelve are unused: 006, 011, 020, 022, 026, 028, 029, 030, 033, 036, 037, 040. No record ever held them.
+- ADR-000 is the full architecture record. ADR-001 is an abridged version of the same
+  decision, kept for readers who want the short form.
 - New records follow the structure in `CLAUDE.md`: Context, Decision, Implementation,
   Rationale, Consequences, Alternatives Considered, Notes, References.
-
-### Numbers that differ from the file name
-
-Three records give themselves a number that their file name does not match. Search by
-either and use this table to find the other.
-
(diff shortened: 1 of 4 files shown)


docs/designs/README.md
```diff
@@ -0,0 +1,103 @@
...
+`AGENTS.md` both point here for that reason.
+
+## Conventions
+
+- The number below is the one the record gives itself in its own heading.
+- The rows are in file-name order, which is the order the directory lists them in.
+- New records follow the structure in `CLAUDE.md`: Context, Decision, Implementation,
+  Rationale, Consequences, Alternatives Considered, Notes, References.
+
+### Numbers that differ from the file name
```
lalitadithya: can we fix these and make them consistent?

lalitadithya:
love the readme being the index, let's fix the numbering so that we don't carry debt

docs/designs/README.md
dims: on it

docs/designs/README.md
```diff
@@ -0,0 +1,80 @@
...
+the decision costs. They are written before the code, not after it, and they are the
+fastest way to learn why this repository is shaped the way it is.
+
+Read the relevant record before you change the behaviour it describes. `CLAUDE.md` and
+`AGENTS.md` both point here for that reason.
+
+## Conventions
+
+- The file name carries the number, and each record's heading matches it.
+- The numbers are not contiguous. Twelve are unused: 006, 011, 020, 022, 026, 028, 029, 030, 033, 036, 037, 040. No record ever held them.
```
lalitadithya: can fix and drop this?

dims: on it!

lalitadithya (approved):

PR #55, opened 2026-08-07
fix: make failureLog usable for triage

dims:
## What was wrong

`status.failureLog` is the one field built for triage. It was unreliable in both directions. Three real failures on an A100 cluster produced two records an operator could not use.

### Present, but misleading

`failureLogTailLines` was 30. That is right for a workload whose last line is the error, which is why it worked well the first time it fired. It is wrong for a workload that prints a structured document.

`dcgmi diag --json` ends with closing braces, an `entity_groups` block and a `metadata` block. A failed level 4 run stored 392 bytes ending in:

```
"test_summary" : { "status" : "Pass" }
...
"metadata" : { "Driver Version Detected" : "580.126.20", "version" : "4.5.2" }
```

**The failure record of a failed job read `Pass`,** and named no failing test.

I measured this again while preparing the fix, running the same image and command the catalog uses (`nvcr.io/nvidia/cloud-native/dcgm:4.5.2-1-ubuntu22.04`, `dcgmi diag --json --verbose`) with `--run 1` for speed:

| | |
|---|---|
| whole document | 50 lines, 705 bytes |
| last 30 lines | **392 bytes** |
| lines holding `"status"` | 18 and 23, i.e. the first half |

So the tail is structurally the wrong end of the document.

### Absent entirely

When the pod was already deleted, `captureFailureLog` found no pod and returned silently, as its best-effort comment allowed. The Job was `Failed` with no diagnostics at all. Seen twice: once after the launcher hit `BackoffLimitExceeded` and JobSet reaped the pod, once after a deliberate worker delete.

## What changed

All in `pkg/controller`:

1. **Request 800 lines rather than 30, and cap the stored text at 32 KB.**
2. **Keep the last bytes when the cap is hit.** The old code used `io.ReadAll(io.LimitReader(stream, 8*1024))`, which keeps the *first* 8 KB and so drops the most recent output, which is the part a failure record needs. `readLogTail` slides a fixed window instead, so the byte cap trims from the front.
3. **Always record something.** When no pod is available, store `PodNotFound` or `NoTerminatedContainer` with a line saying why, rather than leaving the field unset. A failed Job now never carries an empty record.

The timeout capture in `workflow_controller.go` used the same constant and the same `LimitReader`, so it uses the shared helper now.

No schema change. `make manifests` only rewrites two description strings.

## Tests

`pkg/controller/failure_log_test.go`:

- `readLogTail` keeps the end, not the start, including across many small chunked reads such as a live log stream produces
- a structured document keeps its failing entry under the new cap, and loses it under a small one, which is the bug
- `captureFailureLog` records `PodNotFound` when the pod is gone, and `NoTerminatedContainer` when pods exist but none failed

I checked these catch regressions rather than passing by construction. Reverting the tail direction to `io.LimitReader` fails 2 tests. Removing the stub record fails 2 more.

## Verification

`make lint` 0 issues, `make build` clean, `make test` all packages pass. No golden file changed.

## Note on the numbers

800 lines and 32 KB are chosen, not derived. 32 KB per failed Job in status is comfortable against the etcd object limit, and covers the measured document many times over. I could not size against a real `--run 4 --verbose` failure without a 70 minute run on known-bad hardware. If a maintainer would rather see that measured first, say so and I will capture it.

### DIFF
--- a/api/v1alpha1/job_types.go
+++ b/api/v1alpha1/job_types.go
@@ -298,6 +298,8 @@ type JobStatus struct {
 	// failureLog captures the tail of pod logs from the most recent workload failure.
 	// Populated when the Job transitions to Failed. Only the pod that caused the
 	// failure (earliest non-zero exit code) is captured. Overwritten on each retry.
+	// Always set on failure: when no pod was available to read, reason and tail
+	// record why rather than leaving the field unset.
 	// +optional
 	FailureLog *FailureLog `json:"failureLog,omitempty"`
 }
@@ -313,7 +315,8 @@ type FailureLog struct {
 	// reason is the termination reason (e.g., "OOMKilled", "Error").
 	// +optional
 	Reason string `json:"reason,omitempty"`
-	// tail is the last ~30 lines of the failed container's logs.
+	// tail is the end of the failed container's logs, up to 32 KB.
+	// When no pod was available to read, this explains why instead.
 	// +optional
 	Tail string `json:"tail,omitempty"`
 }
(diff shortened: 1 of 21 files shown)


ndipebot (dismissed):
/approve
/lgtm

lalitadithya (approved):

PR #57, opened 2026-08-07
fix: say which nodes a certification left untested, and make discovery deterministic

dims:
## What was wrong

A Certification reports `PASSED` after silently excluding nodes.

When the target set holds more than one GPU architecture, the Workflow keeps the primary architecture, drops the rest, and succeeds. Reproduced on a 2-node A100 cluster by relabelling one node:

```
$ ncrectl cluster info
Error: heterogeneous GPUs: NVIDIA-A100-PCIE-40GB (1), NVIDIA-H100-80GB-HBM3 (1)     <- correct

$ ncrectl certification run --cert-file ... --wait
  GPU:       h100
  Nodes:     1
  Status:    Succeeded
  Categories:   1/1 passed
  Failed Nodes: none
  Result:       PASSED                                                              <- the A100 node is gone
```

The Certification status says only `Succeeded=True — All certification categories completed successfully`. The report says only `GPU: h100`, which reads as "this is an h100 cluster", not "your A100 went untested".

A `HeterogeneousGPU` Warning event does exist on the Workflow. Nothing surfaces it, so `certification run --wait` never shows it. On a large fleet, odd-labelled nodes stay uncertified indefinitely while every report says PASSED.

## Commit 1 — record what was excluded

- `OrchestrationStatus` gains `excludedNodes` and `exclusionReason`.
- The Workflow controller fills them when it filters, and the existing Warning event now names the dropped nodes.
- The report prints an `Excluded:` line under `Nodes:`, and both fields reach the JSON report.

```
  Platform:  onprem
  GPU:       h100
  Nodes:     1
  Excluded:  1 (node002)
             target set has more than one GPU architecture; certified h100 only
```

Both fields are `omitempty`, so a run that excludes nothing is byte-identical to before. No existing golden file changed.

**The pass/fail verdict is deliberately untouched.** Whether a mixed target set should fail outright — the way a mixed *platform* already does in `detectPlatformConsistent` (`workflow_controller.go:216-224`) — is a separate decision, and I did not want to make it inside a visibility fix.

## Commit 2 — give discovery a stable order

While writing the test for commit 1, I found something worse: **which nodes get certified was not deterministic.**

`detectGPUArchConsistent` keeps `nodes[0]`'s architecture, and the node list was never sorted — the only `sort.Slice` calls are on groups, after partitioning. Two envtest runs over identical input (two H100 nodes and one A100) produced:

| Run | Certified | Excluded |
|---|---|---|
| A | `h100`, **2 nodes** | `gpu-node-03-odd` |
| B | `a100`, **1 node** | `gpu-node-01`, `gpu-node-02` |

Same cluster, same config, different subset certified — and on run B the single minority node won over the two H100s. That is a coin flip deciding which hardware gets tested.

`discoverTargetNodes` now sorts by name before returning, which fixes it for every caller rather than only for architecture detection. Name order also matches how `pkg/orchestration` already chunks nodes into groups.

**Behaviour change worth noting:** on a mixed target set the certified subset may differ from before, but only because the previous winner was arbitrary. It is now the architecture of the first node by name.

## Tests

`pkg/controller/excluded_nodes_test.go` — the exclusion helper, plus a test pinning the current first-node-wins rule so changing it has to be deliberate.

`pkg/report/excluded_test.go` — the report line renders, is omitted when nothing is excluded, and both fields reach the JSON.

`pkg/controller/discover_order_test.go` — feeds `discoverTargetNodes` a reader that returns nodes in a deliberately wrong order. **The controller-runtime fake client sorts by name**, so a test built on it passes with or without the change and proves nothing; my first attempt did exactly that. With the stub reader, 2 of 3 subtests fail without the sort and the already-sorted case still passes.

`cmd/integration/testdata/reconcile/workflow-heterogeneous-gpu/` — end to end: three nodes in, two certified, the odd one named in `status.excludedNodes`. This case was written before the sort, found to be genuinely flaky, and removed; with discovery ordered it is stable. I ran it 5 times to confirm.

## Verification

`make lint` 0 issues, `make build` clean, `make test` all packages pass. Hardware reproduction on a 2-node A100 cluster; cluster restored to baseline afterwards.


pkg/controller/workflow_detect.go
```diff
@@ -114,6 +115,23 @@ func detectGPUArchitecture(nodes []corev1.Node) string {
...
+	for i := range kept {
+		keptNames[kept[i].Name] = true
+	}
+	var out []string
+	for i := range all {
+		if !keptNames[all[i].Name] {
+			out = append(out, all[i].Name)
+		}
+	}
+	sort.Strings(out)
```
ndipebot: This sort is redundant. `discoverTargetNodes` now returns nodes sorted by name, so `all` is already in order and the loop produces excluded names in name order naturally.

pkg/report/report.go
```diff
@@ -284,6 +289,8 @@ func Build(ctx context.Context, c client.Client, cert *burninv1alpha1.Certificat
 			report.Platform = wf.Status.Orchestration.DetectedPlatform
 			report.GPU = wf.Status.Orchestration.DetectedGPUArchitecture
 			report.TotalNodes = wf.Status.Orchestration.TotalNodes
+			report.ExcludedNodes = wf.Status.Orchestration.ExcludedNodes
+			report.ExclusionReason = wf.Status.Orchestration.ExclusionReason
```
ndipebot, line 293: No test for Build populating exclusion from workflow status. The report tests construct CertReport directly. There's nothing that feeds real `OrchestrationStatus.ExcludedNodes` through Build and verifies it reaches the report.

pkg/controller/discover_order_test.go
```diff
@@ -0,0 +1,107 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ndipebot: Tests should use `testutil.TestCaseParser`

pkg/controller/excluded_nodes_test.go
```diff
@@ -0,0 +1,93 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ndipebot: Tests should use `testutil.TestCaseParser`

pkg/report/excluded_test.go
```diff
@@ -0,0 +1,61 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ndipebot: Tests should use `testutil.TestCaseParser`

pkg/controller/workflow_detect.go
dims: Correct. Dropped in 01a295f.

pkg/report/report.go
dims: Agreed, and this was the real gap. I tested the rendering and the controller write. Nothing tested the step between them.

Added in 01a295f: testdata/build-excluded-report runs a real OrchestrationStatus through Build and compares the rendered text. A control case has nothing excluded. I removed the assignments in Build to check it fails.

pkg/controller/discover_order_test.go
dims: Converted in 01a295f. Now discover_node_order_test.go with testdata/discover-node-order.

pkg/controller/excluded_nodes_test.go
dims: Converted in 01a295f. Now gpu_arch_exclusion_test.go with testdata/gpu-arch-exclusion.

pkg/report/excluded_test.go
dims: Converted in 01a295f. Now build_excluded_test.go with testdata/build-excluded-report.

lalitadithya (approved):

PR #58, opened 2026-08-08
fix: give a WorkloadRun a safe way to request a GPU off a DGX

dims:
## What was wrong

A WorkloadRun had no safe way to ask for a GPU off a DGX. Both paths through `pkg/platform/runtime.go` were broken, and the failure modes point away from the cause.

```go
if cfg.Resources != nil {
    container["resources"] = cfg.Resources                            // used exactly as written
} else {
    container["resources"] = defaultWorkerResources(cfg.GpusPerNode)  // GPUs + memory: 800Gi
}
```

**Omit `spec.resources`** and you get `memory: 800Gi`. Only a DGX-sized node can satisfy that, so anywhere else the pod stays `Pending` forever.

**Set `spec.resources`** to dodge that, and your block is used verbatim. Kubernetes only attaches a GPU to a container that names `nvidia.com/gpu`, so unless you remembered to write it yourself, the pod schedules onto a node it cannot use.

The second case is the nastier one. `numProcPerNode` is set from `gpusPerNode` *outside* that branch:

```go
"mlPolicy": map[string]any{
    "numNodes": cfg.NodesPerJob,
    "torch": map[string]any{
        "numProcPerNode": cfg.GpusPerNode,   // always runs
    },
},
```

So the runtime is told to start one process per GPU while the container has none. The run dies inside CUDA reporting something like *driver version is insufficient*, which sends whoever is debugging it to check the NVIDIA driver, the GPU Operator and the node image. The actual cause is a missing line in their own YAML.

Both branches were confirmed on a 2-node A100 cluster. It is written up as fact F6 in the test plan because we got caught by it there.

## Commit 1 — keep the GPU request when the user sets `resources`

`withGPURequest` fills `nvidia.com/gpu` in from `gpusPerNode` when the user did not name it.

Deliberately conservative: if the resource appears in **either** `limits` or `requests`, the block is left entirely alone. So an explicit value is never overwritten, including an explicit `0` — asking for no GPU stays expressible. It copies rather than editing the caller's block, which a test checks.

Applies to both the torch and MPI runtimes, which had the same either-or shape (`:98-101` and `:229-232`).

**No golden file changes in this commit.**

## Commit 2 — stop guessing 800Gi

`defaultWorkerResources` no longer invents a memory number.

The 800Gi came from the training catalog entries (`nemotron5-8b`, `nemotron5-56b`), which set their own resources and are unaffected. CRE cannot know what an arbitrary WorkloadRun needs. The GPU count is the part it does know, and stays the only part it fills in.

Regenerates the only two goldens that reach this default:

```
 cmd/integration/testdata/reconcile/workloadrun-mpi/expected.json   | 2 --
 cmd/integration/testdata/reconcile/workloadrun-torch/expected.json | 2 --
```

Both inputs omit `spec.resources`. Each loses two `"memory": "800Gi"` lines; the `nvidia.com/gpu` request is unchanged. Catalog-driven goldens do not move.

## Behaviour change worth calling out

Anyone omitting `spec.resources` today gets a pod that cannot schedule off a DGX, so in practice they are already broken. But anyone relying on the 800Gi reservation **on** a DGX loses it. That seemed better than keeping a number CRE has no basis to pick, and it is the half of the report that says "omit resources → unschedulable".

If maintainers would rather keep a memory default, commit 2 can be dropped on its own and commit 1 still makes `spec.resources` safe to use.

## Tests

`pkg/platform/resources_test.go`:

- memory-only block gets the GPU filled in, and the user's own values survive
- an explicit GPU count is kept; an explicit `0` is kept
- naming it in `requests` only leaves the block alone
- an empty block still gets a GPU; `nil` passes through
- the caller's block is not mutated
- the default asks for GPUs and nothing else
- both branches end up requesting a GPU

## Verification

### DIFF
--- a/pkg/platform/resources_test.go
+++ b/pkg/platform/resources_test.go
@@ -1,0 +1,97 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package platform
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func rl(pairs map[string]string) corev1.ResourceList {
+	out := corev1.ResourceList{}
+	for k, v := range pairs {
+		out[corev1.ResourceName(k)] = resource.MustParse(v)
+	}
+	return out
+}
+
+// A WorkloadRun that sets spec.resources used to lose its GPU request, because
+// the block was taken exactly as written. The pod then landed on a node it
+// could not use and failed inside CUDA.
+func TestWithGPURequest(t *testing.T) {
+	tests := []struct {
+		name        string
+		res         *corev1.ResourceRequirements
+		gpusPerNode int32
+		wantGPU     string // "" means the resource must be absent
+		wantMemory  string
+	}{
+		{
+			name:        "memory only gets the GPU filled in",
+			res:         &corev1.ResourceRequirements{Limits: rl(map[string]string{"memory": "64Gi"})},
+			gpusPerNode: 1,
+			wantGPU:     "1",
+			wantMemory:  "64Gi",
+		},
+		{
+			name:        "an explicit GPU count is kept",
+			res:         &corev1.ResourceRequirements{Limits: rl(map[string]string{"nvidia.com/gpu": "2"})},
+			gpusPerNode: 8,
+			wantGPU:     "2",
+		},
+		{
+			name:        "an explicit zero is kept, so asking for no GPU still works",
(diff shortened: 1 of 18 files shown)


pkg/platform/resources_test.go
```diff
@@ -0,0 +1,130 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ndipebot, line 1: Tests should use `testutil.TestCaseParser` with testdata directories and golden files, same pattern as `pkg/catalog/gpu_defaults_test.go`. Each case gets an input.yaml and expected.json.

ndipebot (changes requested):

pkg/platform/resources_test.go
dims: Converted in 21cbea6, following gpu_defaults_test.go. Seven cases under testdata/worker-resources.

ndipebot (approved):
/approve
/lgtm

PR #64, opened 2026-08-08
fix: bound a WorkloadRun Job so it cannot run forever

dims:
## What was wrong

`isJobTimedOut` returns false when `Execution.TimeoutPerJob` is nil:

```go
func (r *WorkflowReconciler) isJobTimedOut(workflow *..., g *...) bool {
	timeout := workflow.Spec.Orchestration.Execution.TimeoutPerJob
	if timeout == nil || g.StartTime == nil {
		return false
	}
	return time.Since(g.StartTime.Time) > timeout.Duration
}
```

So a WorkloadRun that set no timeout could never time out. Observed on hardware: a run sat `InProgress` for **4h10m** against a node with no allocatable GPUs, and only stopped because someone was watching.

Catalog certifications were never exposed to this — the loader defaults `timeoutPerJob` to 1h, or 15m for diagnose. Only the WorkloadRun path left it unset.

## Three ways it stayed nil, not one

The bug report described one. Reading the code found three:

1. **An empty `timeoutPerJob`** — the obvious one.
2. **`spec.orchestration` absent altogether**, which skips the whole block. This is the common case for a small WorkloadRun, and the original note missed it.
3. **A discarded parse error.** `d, parseErr := time.ParseDuration(...)`, and on failure nothing was set and nothing was said, so a typo left the Job unbounded silently.

`resolveWRTimeout` covers all three, in both copies of this logic (`pkg/controller/workloadrun_controller.go` and `pkg/workloadrun/workloadrun.go`).

## Why 24h and not 1h

Deliberately far longer than the catalog's default. A WorkloadRun is arbitrary user work — training through a WorkloadRun is a documented use case — and cutting a legitimate multi-hour run short would be worse than the hang it prevents.

This is a backstop against never finishing, not a performance bound. Set `spec.orchestration.timeoutPerJob` for anything tighter.

Happy to change the number if maintainers would rather match the catalog at 1h; it is one constant, `catalog.DefaultWorkloadRunTimeoutPerJob`.

## Tests

`pkg/controller/wr_timeout_test.go` covers all four inputs — no orchestration block, orchestration without a timeout, an unparseable value, and the user's own value winning — plus a check that `isJobTimedOut` actually fires on the resolved timeout, so the bound is wired through rather than merely stored.

## Golden files

`workloadrun-torch` and `workloadrun-mpi` are the only cases that reach this path. Their execution block goes from `{}` to carrying the default:

```
-        "execution": {},
+        "execution": {
+          "timeoutPerJob": "24h0m0s"
+        },
```

## Verification


ndipebot (approved):
/approve
/lgtm

PR #169, opened 2026-08-18
fix(podlogs): bound pod log stream duration

### LINKED ISSUE #137: BandwidthMeasurement log stream has no context timeout — goroutine leaks on log server failure
### Prerequisites

### Code of Conduct

### Bug Description
`pkg/controller/bandwidthmeasurement_controller.go:2073` (approximately) calls `Stream(ctx)` to stream pod logs without attaching a deadline to `ctx`. If the Kubernetes log server stalls (e.g. the node is network-partitioned), the streaming goroutine blocks indefinitely. The parent reconcile returns, but the goroutine holding the stream connection is never cancelled. Over time, enough stalled nodes accumulate leaked goroutines to exhaust the controller's goroutine budget.

**Location:** `pkg/controller/bandwidthmeasurement_controller.go` (Stream call)

**Suggested fix:** Wrap the stream context with `context.WithTimeout(ctx, 2*time.Minute)` before calling `Stream`.

#

### Component

Controller / Reconcilers

### Steps to Reproduce
1. Simulate a log server stall: block the kubelet log port (10250) on a node while a BandwidthMeasurement is reading logs.
2. Observe goroutine count rising in the controller process (`/debug/pprof/goroutine`).
3. After enough stalls, the controller becomes unresponsive.

#### Environment
- CRE version: all
- Kubernetes version: all
- GPU architecture: all
- Deployment method: all

kaynetu:
## Summary

- Bound the complete lifecycle of pod log requests to two minutes, covering both stream creation and response-body reads.
- Reuse the bounded stream helper for measurement collection, Job failure-log capture, and Workflow timeout-log capture.
- Add regression tests for header stalls, body-read stalls, earlier parent deadlines, the default timeout, and cancellation on close.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected

- [ ] API / CRDs
- [ ] Catalog / Workloads
- [ ] CLI (ncrectl)
- [ ] Helm / Deployment
- [ ] Documentation / CI
- [x] Other: Pod log fetching

## Testing

- [x] `make manifests generate`
- [x] `make lint-fix`
- [x] `make build`
- [x] `make test`
- [x] Post-rebase `go test ./pkg/podlogs ./pkg/controller`
- [x] No breaking changes

## Risk Assessment

Low. A stalled log request can still occupy a reconcile worker for up to two minutes, but it can no longer block indefinitely. Measurement fetch errors skip the current sample without advancing the log-fetch anchor, so the next sample retries the window subject to the existing maximum-lookback clamp.

## Checklist

- [x] Self-review completed
- [x] Golden files not required; integration output is unchanged
- [x] Documentation updates not required; no API or user-facing configuration changed
- [x] Ready for review

Fixes #137

### DIFF
--- a/pkg/podlogs/fetcher.go
+++ b/pkg/podlogs/fetcher.go
@@ -69,9 +69,10 @@ func (f *kubernetesLogFetcher) FetchLogs(ctx context.Context, namespace, podName
 	return ScanLines(io.LimitReader(stream, limit))
 }
 
-// OpenStream opens a pod log request with DefaultStreamTimeout. The returned
-// stream keeps the timeout active until it is closed, so stalled response-body
-// reads are bounded as well as the initial request.
+// OpenStream opens a pod log request bounded by DefaultStreamTimeout or the
+// parent context's deadline, whichever is earlier. The returned stream keeps
+// that deadline active until it is closed, so stalled response-body reads are
+// bounded as well as the initial request.
 func OpenStream(ctx context.Context, req *rest.Request) (io.ReadCloser, error) {
 	return openStreamWithTimeout(ctx, req, DefaultStreamTimeout)
 }
--- a/pkg/podlogs/fetcher.go
+++ b/pkg/podlogs/fetcher.go
@@ -9,11 +9,17 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 )
+
+// DefaultStreamTimeout bounds the complete lifecycle of a pod log request,
+// including opening the stream and consuming its response body.
+const DefaultStreamTimeout = 2 * time.Minute
 
 // PodLogFetcher abstracts fetching raw log lines from a pod.
 type PodLogFetcher interface {
@@ -52,7 +58,7 @@ func (f *kubernetesLogFetcher) FetchLogs(ctx context.Context, namespace, podName
 	podLogOpts.LimitBytes = &limit
 
 	req := f.clientset.CoreV1().Pods(namespace).GetLogs(podName, podLogOpts)
-	stream, err := req.Stream(ctx)
+	stream, err := OpenStream(ctx, req)
 	if err != nil {
 		return nil, fmt.Errorf("opening log stream: %w", err)
 	}
@@ -61,6 +67,37 @@ func (f *kubernetesLogFetcher) FetchLogs(ctx context.Context, namespace, podName
 	// LimitBytes is enforced by the API server, but a misbehaving or proxied
 	// endpoint could ignore it — bound the client side too.
 	return ScanLines(io.LimitReader(stream, limit))
+}
+
+// OpenStream opens a pod log request with DefaultStreamTimeout. The returned
+// stream keeps the timeout active until it is closed, so stalled response-body
+// reads are bounded as well as the initial request.
+func OpenStream(ctx context.Context, req *rest.Request) (io.ReadCloser, error) {
+	return openStreamWithTimeout(ctx, req, DefaultStreamTimeout)
+}
+
+func openStreamWithTimeout(ctx context.Context, req *rest.Request, timeout time.Duration) (io.ReadCloser, error) {
+	streamCtx, cancel := context.WithTimeout(ctx, timeout)
+	stream, err := req.Stream(streamCtx)
+	if err != nil {
+		cancel()
+		return nil, err
+	}
+
+	return &cancelReadCloser{
+		ReadCloser: stream,
+		cancel:     cancel,
+	}, nil
+}
+
+type cancelReadCloser struct {
+	io.ReadCloser
+	cancel context.CancelFunc
+}
+
+func (r *cancelReadCloser) Close() error {
+	r.cancel()
+	return r.ReadCloser.Close()
 }
 
 // DefaultMaxLogBytes caps a single log fetch at 8 MB. Sized well above a normal
--- a/pkg/podlogs/fetcher_test.go
+++ b/pkg/podlogs/fetcher_test.go
@@ -1,0 +1,147 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package podlogs
+
+import (
+	"context"
+	"io"
+	"net/http"
+	"strings"
+	"testing"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestOpenStreamTimesOutBeforeResponseHeaders(t *testing.T) {
+	transport := roundTripperFunc(func(req *http.Request) (*http.Response, error) {
+		<-req.Context().Done()
+		return nil, req.Context().Err()
+	})
+
+	req := newPodLogRequest(t, transport)
+	stream, err := openStreamWithTimeout(context.Background(), req, 50*time.Millisecond)
+
+	require.Nil(t, stream)
+	require.Error(t, err)
+	assert.ErrorIs(t, err, context.DeadlineExceeded)
+}
+
+func TestOpenStreamTimesOutWhileReadingResponseBody(t *testing.T) {
+	transport := roundTripperFunc(func(req *http.Request) (*http.Response, error) {
+		return &http.Response{
+			StatusCode: http.StatusOK,
+			Header:     make(http.Header),
+			Body: &contextBlockingBody{
+				ctx:    req.Context(),
+				reader: strings.NewReader("complete line\n"),
+			},
+			Request: req,
+		}, nil
+	})
+
+	req := newPodLogRequest(t, transport)
+	stream, err := openStreamWithTimeout(context.Background(), req, 50*time.Millisecond)
+	require.NoError(t, err)
+	t.Cleanup(func() { require.NoError(t, stream.Close()) })
+
+	lines, err := ScanLines(stream)
+
+	require.Error(t, err)
+	assert.ErrorIs(t, err, context.DeadlineExceeded)
+	assert.Equal(t, []string{"complete line"}, lines)
+}
+
+func TestOpenStreamHonorsEarlierParentDeadline(t *testing.T) {
+	transport := roundTripperFunc(func(req *http.Request) (*http.Response, error) {
+		<-req.Context().Done()
+		return nil, req.Context().Err()
+	})
+
+	req := newPodLogRequest(t, transport)
+	ctx, cancel := context.WithTimeout(context.Background(), 50*time.Millisecond)
+	t.Cleanup(cancel)
+
+	stream, err := openStreamWithTimeout(ctx, req, time.Minute)
+
+	require.Nil(t, stream)
+	require.Error(t, err)
+	assert.ErrorIs(t, err, context.DeadlineExceeded)
+}
+
+func TestOpenStreamUsesDefaultTimeoutAndCancelsOnClose(t *testing.T) {
+	requestContexts := make(chan context.Context, 1)
+	transport := roundTripperFunc(func(req *http.Request) (*http.Response, error) {
+		requestContexts <- req.Context()
+		return &http.Response{
+			StatusCode: http.StatusOK,
+			Header:     make(http.Header),
+			Body:       io.NopCloser(strings.NewReader("")),
+			Request:    req,
+		}, nil
+	})
+	clientset, err := kubernetes.NewForConfig(&rest.Config{
+		Host:      "http://pod-logs.test",
+		Transport: transport,
+	})
+	require.NoError(t, err)
+
+	started := time.Now()
+	stream, err := OpenStream(context.Background(), clientset.CoreV1().Pods("default").GetLogs("worker-0", &corev1.PodLogOptions{}))
+	require.NoError(t, err)
+
+	requestCtx := <-requestContexts
+	deadline, ok := requestCtx.Deadline()
+	require.True(t, ok)
+	assert.WithinDuration(t, started.Add(DefaultStreamTimeout), deadline, time.Second)
+
+	require.NoError(t, stream.Close())
+	select {
+	case <-requestCtx.Done():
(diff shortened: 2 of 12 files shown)


ndipebot (commented):
Reviewed the timeout change. The core fix is right: wrapping the stream so the deadline stays live through response-body reads closes the real gap in #137 (`req.Stream(ctx)` alone only bounds the request, not the read). All three `GetLogs(...).Stream(ctx)` call sites in the repo are converted, so there is no straggler. `go test ./pkg/podlogs -race -count=8` is green and stable, so the 1s `WithinDuration` tolerance is not flaky.

Main request is the test format — see the inline comment on `fetcher_test.go`. These are hand-rolled table-free tests, while the repo's pattern for this exact feature area (`TestFailureLogTail` in `pkg/controller/failure_log_test.go`) uses `testutil.TestCaseParser` with a `testdata/` dir and golden files. Plus one real ordering bug in `Close()` and a testability note on the hardcoded timeout.

pkg/podlogs/fetcher.go
```diff
@@ -63,6 +69,37 @@ func (f *kubernetesLogFetcher) FetchLogs(ctx context.Context, namespace, podName
...
+	}, nil
+}
+
+type cancelReadCloser struct {
+	io.ReadCloser
+	cancel context.CancelFunc
+}
+
+func (r *cancelReadCloser) Close() error {
+	r.cancel()
```
ndipebot, line 100: > [!WARNING]
> **WITHDRAWN — measured and refuted; see my reply below. The current `cancel(); Close()` order is correct. Original text kept for the record:**

<details><summary>Original comment</summary>

**Cancel-before-close defeats HTTP connection reuse.**

Cancelling the request context makes `net/http`'s transport tear down the underlying connection, so closing the body afterwards can never return it to the idle pool. Since `FetchLogs` normally reads to EOF (the logs are far below the 8&nbsp;MB `LimitBytes`), this connection *would* otherwise be reusable — so every goodput/bandwidth sample now forces a fresh TCP + TLS handshake against the API server. With per-pod sampling on a large cluster that adds up.

Closing first and cancelling second is equally safe — `cancel` still runs, and it becomes a no-op for a connection that has already been released cleanly:

```go
func (r *cancelReadCloser) Close() error {
	err := r.ReadCloser.Close()
	r.cancel()
	return err
}
```

</details>

pkg/podlogs/fetcher.go
```diff
@@ -9,12 +9,18 @@ import (
...
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 )
 
+// DefaultStreamTimeout bounds the complete lifecycle of a pod log request,
+// including opening the stream and consuming its response body.
+const DefaultStreamTimeout = 2 * time.Minute
```
ndipebot, line 22: > [!WARNING]
> **WITHDRAWN — refuted by @kaynetu below; see the thread. Callers can already shorten the bound by wrapping the parent context. Original text kept for the record:**

<details><summary>Original comment</summary>

**The timeout isn't injectable, so nothing outside this package can shorten it.**

`openStreamWithTimeout` is unexported, so the two controller call sites can only reach the hardcoded 2 minutes via `OpenStream`. That runs against the repo's convention of making these durations configurable (CLAUDE.md: "Configurable requeue intervals: tests use 1s, production uses 15s"), and it matters here because the integration suite's test timeout is 10s — a stalled fetch in an envtest run would blow the whole suite rather than fail fast.

Suggest either exporting `OpenStreamWithTimeout`, or adding a package-level `var streamTimeout = DefaultStreamTimeout` that tests can override.

</details>

pkg/podlogs/fetcher.go
```diff
@@ -63,6 +69,37 @@ func (f *kubernetesLogFetcher) FetchLogs(ctx context.Context, namespace, podName
 	return ScanLines(io.LimitReader(stream, limit))
 }
 
+// OpenStream opens a pod log request with DefaultStreamTimeout. The returned
```
ndipebot: **The doc comment is slightly misleading — and the test that should pin this down doesn't.**

Wording first: "opens a pod log request with `DefaultStreamTimeout`" reads as if the 2 minutes always applies. Because this is `context.WithTimeout(ctx, timeout)`, an earlier parent deadline wins — the 2 minutes is a ceiling, not a floor. That's the right behaviour and worth stating: *"...bounded by `DefaultStreamTimeout` or the parent context's deadline, whichever is earlier."*

The more substantive part: **`TestOpenStreamHonorsEarlierParentDeadline` does not actually verify that.** I mutation-tested it — an implementation that ignores the parent context entirely still passes. It just takes 60s instead of 0.05s.

The reason is that both deadlines produce the same error. With parent=50ms and stream=1m, the transport blocks on `req.Context().Done()` and returns `ctx.Err()`; whether the 50ms parent or the 1m stream deadline fires, the result is `context.DeadlineExceeded`, and `assert.ErrorIs(err, context.DeadlineExceeded)` can't tell them apart. Nothing asserts on elapsed time, so the only observable difference — 0.05s vs 60s — goes unchecked.

An elapsed-time bound is what makes it a real test:

```go
start := time.Now()
stream, err := openStreamWithTimeout(ctx, req, time.Minute)
elapsed := time.Since(start)

require.Nil(t, stream)
assert.ErrorIs(t, err, context.DeadlineExceeded)
assert.Less(t, elapsed, time.Second, "parent deadline should win over the 1m stream timeout")
```

When this moves to `testutil` (see my comment on `fetcher_test.go`), that becomes an `elapsedUnder` field in the golden rather than an inline assertion — which also stops this case and `stall-before-headers` from producing byte-identical goldens.

pkg/podlogs/fetcher_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"testing"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
```
ndipebot, line 24: **These don't follow the repo's test format.**

Every unit test in this repo uses `testutil.TestCaseParser` with a `testdata/<subdir>/<case>/` directory and golden files — 40 of the 71 test files use it today, and it is the required pattern for new tests, without exception. It also already covers this feature area: `TestFailureLogTail` at `pkg/controller/failure_log_test.go:31` tests `readLogTail`, the function that consumes these streams:

```go
p := testutil.TestCaseParser{Subdir: "failure-log-tail", ExpectedSuffix: ".json"}
p.TestDir(t, func(tc *testutil.TestCase) error { ... tc.Actual = ... })
```

**All four tests in this file should be converted to that shape**, e.g. under `testdata/open-stream/<case>/` with `input.yaml` selecting the scenario and `expected.json` holding the serialized outcome:

| case | `input.yaml` | `expected.json` |
|---|---|---|
| `stall-before-headers` | `stall: headers`, `parentTimeout: 0`, `streamTimeout: 50ms` | `{"lines": [], "errorKind": "DeadlineExceeded", "streamOpened": false}` |
| `stall-reading-body` | `stall: body`, `parentTimeout: 0`, `streamTimeout: 50ms` | `{"lines": ["complete line"], "errorKind": "DeadlineExceeded", "streamOpened": true}` |
| `parent-deadline-earlier` | `stall: headers`, `parentTimeout: 50ms`, `streamTimeout: 1m` | `{"lines": [], "errorKind": "DeadlineExceeded", "streamOpened": false, "elapsedUnder": "1s"}` |
| `default-timeout-cancel-on-close` | `stall: none`, `streamTimeout: default` | `{"deadlineFromDefault": true, "cancelledOnClose": true}` |

Two notes on making that conversion actually pull its weight:

1. **The scenarios are less uniform than a quick read suggests**, so don't force one shared harness. `TestOpenStreamTimesOutBeforeResponseHeaders` and `TestOpenStreamHonorsEarlierParentDeadline` share a transport that blocks on `req.Context().Done()`, but `TestOpenStreamTimesOutWhileReadingResponseBody` returns a real `*http.Response` wrapping a bespoke `contextBlockingBody` and additionally asserts on the returned lines. The transport behaviour itself can't live in `testdata` — only a name for it can — so expect a small `switch` on a `stall:` string that selects the `RoundTripper`, with everything else driven from the case files.

2. **Give each case a distinguishing field.** A naive `{lines, errorKind}` golden makes `stall-before-headers` and `parent-deadline-earlier` byte-identical, so the goldens would stop discriminating exactly where the behaviour differs. That's why `streamOpened` / `elapsedUnder` are in the table above. This matters more than it looks — see my separate note on `fetcher.go:72`: as written, `TestOpenStreamHonorsEarlierParentDeadline` passes even against an implementation that ignores the parent context entirely, because both deadlines yield `DeadlineExceeded` and nothing asserts on elapsed time.

*(Amending an earlier version of this comment. It cited `failure_log_test.go:32` — the correct line is :31 — described the three tests as "near-identical", which overstates it, and suggested `TestOpenStreamUsesDefaultTimeoutAndCancelsOnClose` was reasonable to leave hand-written. That exemption was wrong: testutil is required for all tests here, and that case serializes fine as shown above.)*

pkg/podlogs/fetcher.go
kaynetu: It might not be a good idea to have the package-level var. It is mutable global state, and parallel tests can race it.

`context.WithTimeout` takes the **earlier** of the parent deadline and the new one. The 2-minute cap is a ceiling, not a floor.

```go
func openStreamWithTimeout(ctx context.Context, req *rest.Request, timeout time.Duration) (io.ReadCloser, error) {
	streamCtx, cancel := context.WithTimeout(ctx, timeout)
	stream, err := req.Stream(streamCtx)
```

If the parent already dies at 50ms, the stream dies at 50ms even when the helper is asked for a minute. That is exactly what this test does:

```go
func TestOpenStreamHonorsEarlierParentDeadline(t *testing.T) {
	// ...
	ctx, cancel := context.WithTimeout(context.Background(), 50*time.Millisecond)
	t.Cleanup(cancel)

	stream, err := openStreamWithTimeout(ctx, req, time.Minute)

	require.Nil(t, stream)
	require.Error(t, err)
	assert.ErrorIs(t, err, context.DeadlineExceeded)
}
```

Same thing from a caller outside the package — no export needed:

```go
ctx, cancel := context.WithTimeout(ctx, time.Second)
defer cancel()

stream, err := podlogs.OpenStream(ctx, req) // still "2 minutes", but this call fails in 1s
```

Callers that require a shorter bound can wrap the parent context before calling OpenStream; production otherwise uses the two-minute safety ceiling.. A test that needs fail-fast does the same: pass a short-lived `ctx` into `OpenStream` / `FetchLogs` instead of adding a package-level `var`.

ndipebot: Agreed on both counts — withdrawing this comment.

The package-level `var` was bad advice: mutable global state that races under `t.Parallel()`, which is a worse problem than the one it solves.

And the premise was wrong. `context.WithTimeout` taking the earlier deadline means every caller can already shorten the bound by wrapping the parent — no export, no new API. So "nothing outside this package can shorten it" is simply false; a caller that wants 1s passes a 1s ctx and gets 1s. That also contradicted my own comment on `fetcher.go:72`, where I said an earlier parent deadline wins. Both can't be true, and yours is the one that is.

Nothing left of the finding — the 2 minutes is a ceiling callers can lower at will, which is the right design. Sorry for the noise.

pkg/podlogs/fetcher.go
ndipebot: **Retracting this comment — I measured it and I was wrong. The current order is fine; no change needed.**

I claimed cancelling before `Close()` prevents the connection returning to the idle pool. It doesn't, because `FetchLogs` reads the body to EOF, and net/http returns the connection to the pool **at EOF — before `Close()` is ever called**. The subsequent `cancel()` is a no-op for pooling.

Measured with this exact code (`fetcher.go` copied verbatim, real client-go `Clientset`, `httptest` keep-alive server counting `StateNew`), 20 sequential `FetchLogs` calls, ~200KB body:

| ordering | distinct TCP conns / 20 fetches |
|---|---|
| `cancel(); Close()` — current | 1 |
| `Close(); cancel()` — what I suggested | 1 |
| no cancel at all (control) | 1 |

Mechanism confirmed directly: read request 1 to EOF, then *without* closing or cancelling, issue request 2 — `httptrace` reports `GotConn.Reused=true`. The connection was already parked in the idle pool at EOF.

Two more reasons the suggestion was wrong:

- **Partial reads.** If the body *isn't* drained, reuse is lost — but identically in all three variants (20 conns each). The undrained body is what costs the connection, not the cancel, so my reordering wouldn't have fixed the one case where reuse actually breaks.
- **HTTP/2, which makes it moot on a real cluster.** client-go negotiates HTTP/2 by default (`apimachinery` `SetTransportDefaults` → `configureHTTP2Transport`, unless `DISABLE_HTTP2`). Measured over TLS: `proto=HTTP/2.0`, 1 connection for 20 fetches — and there `cancel` sends `RST_STREAM` on one multiplexed stream rather than tearing down TCP. Even the partial-read case stayed at 1 connection.

So "every sample forces a fresh TCP + TLS handshake" was wrong on the mechanism and wrong on the protocol. Apologies for the noise — `cancel(); r.ReadCloser.Close()` is correct as written.

One sub-claim did hold, for the record: `rest.Request.Stream` returns the raw response body unwrapped (client-go v0.36.3 `rest/request.go:931`), and both controller call sites drain to EOF via `readLogTail` with `defer stream.Close()` — so there's no leaked cancel either.

kaynetu: Yes, makes sense

pkg/podlogs/fetcher.go
kaynetu: Addressed in `e8610a2`. The `OpenStream` comment now states that the earlier deadline wins, and the `parent-deadline-earlier` fixture records `elapsedUnder: "1s"` to distinguish the 50 ms parent deadline from the 1 minute stream timeout. `go test -race -count=8 ./pkg/podlogs` passed.

pkg/podlogs/fetcher_test.go
kaynetu: Addressed in `e8610a2`. All four scenarios now use `testutil.TestCaseParser` with `input.yaml` and `expected.json` fixtures, including `streamOpened`, `elapsedUnder`, `deadlineFromDefault`, and `cancelledOnClose` evidence. `make test` and `go test -race -count=8 ./pkg/podlogs` passed.

ndipebot (approved):
/lgtm
/approve

PR #174, opened 2026-08-19
fix(controller): enable concurrent reconciliation

### LINKED ISSUE #138: No MaxConcurrentReconciles on any controller — all reconciles are sequential on large fleets
### Prerequisites

### Code of Conduct

### Bug Description
None of the six CRE reconcilers (Job, Workflow, Certification, GoodputMeasurement, BandwidthMeasurement, WorkloadRun) configure `MaxConcurrentReconciles` in their `SetupWithManager` calls. The controller-runtime default is 1, meaning all reconciles are processed sequentially. On a fleet with 64 nodes and 8 Workflows, the Workflow reconciler processes each Workflow one at a time. Tail latency grows linearly with fleet size.

**Location:** All `SetupWithManager` functions in `pkg/controller/`

**Suggested fix:** Add `controller.Options{MaxConcurrentReconciles: <N>}` to each `SetupWithManager` call. A sensible starting value is `10` for Job/Workflow/Certification and `5` for measurement reconcilers. Expose as a Helm value so operators can tune per-cluster.

#

### Component

Controller / Reconcilers

### Steps to Reproduce
1. Deploy CRE with 100 active Workflows.
2. Observe in the controller logs that Workflow reconciles are serialised — each waits for the previous to complete.
3. `time.Since(reconcile start)` in logs shows linear growth with workflow count.

#### Environment
- CRE version: all
- Kubernetes version: all
- GPU architecture: all
- Deployment method: all

kaynetu:
## Summary

- Enable concurrent reconciliation across all six CRE controllers instead of relying on controller-runtime's sequential default: 10 workers for Job, Workflow, Certification, and WorkloadRun, and 5 workers for GoodputMeasurement and BandwidthMeasurement.
- Expose the two concurrency tiers through validated manager CLI flags and Helm values so operators can tune them for cluster size and API-server capacity.
- Add manager flag tests and Helm rendering coverage for default and overridden values, install Helm in the unit-test CI job, and document the settings in the README.
- Update the Kind/KWOK UAT infrastructure for Kubernetes 1.36 ResourceClaim binding behavior, and keep the manager and DRA stub on the real Kind control-plane node instead of simulated KWOK nodes.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [x] 🔨 Build/CI

## Component(s) Affected

- [ ] API / CRDs
- [ ] Catalog / Workloads
- [ ] CLI (ncrectl)
- [x] Helm / Deployment
- [x] Documentation / CI
- [x] Other: Kind/KWOK UAT infrastructure

## Testing

- [x] `make manifests generate`
- [x] `make lint-fix`
- [x] `make build`
- [x] `make test`
- [x] Local Kubernetes 1.36 Kind/KWOK UAT: `make test-uat-run`
- [x] No breaking changes

## Risk Assessment

Low to moderate. Higher reconcile concurrency can increase Kubernetes API-server load, but the defaults are bounded, measurement controllers use a lower limit because they fetch and parse Pod logs, and both limits are operator-tunable. There are no API or CRD changes. The Kubernetes 1.36 compatibility and node-placement changes are limited to the UAT environment.

## Checklist

- [x] Self-review completed
- [x] Golden files not required; integration output is unchanged
- [x] Documentation updated
- [x] Ready for review

Fixes #138

## Summary by CodeRabbit

* **New Features**
  * Added configurable concurrency limits for standard and measurement reconciliations, with validation and Helm support.
  * Added support for selecting custom Kubernetes node images when creating UAT clusters.
  * Improved UAT workload scheduling and added allocation timestamps for resource claims.

* **Bug Fixes**
  * Improved measurement completion handling for consistent timestamps, accurate sampling, and atomic status updates.

* **Documentation**
  * Documented reconciliation tuning and custom Kubernetes node image configuration.

* **Chores**
  * Updated Kind and Helm versions used by development and CI workflows.


cmd/manager/main_test.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestControllerConcurrencyOptionsValidate(t *testing.T) {
+	tests := []struct {
```
ndipebot: These are table-driven tests, which #172 just converted the whole repo away from -- this reintroduces the pattern a day after we removed the last of it. Both funcs map cleanly onto `testutil.TestCaseParser`:

- `TestControllerConcurrencyOptionsValidate` -> `cmd/manager/testdata/concurrency-validate/<case>/` with an `input.yaml` holding the two ints. The error cases can use the harness's `error` file (substring match on the returned error -- `pkg/workload/testdata/for-spec/empty-spec/error` is the precedent), and the valid case gets a small `expected.json`.
- `TestControllerConcurrencyFlagDefaults` has an exact converted precedent from #172: `TestNewRunCommandFlagDefaults` in `pkg/certification/certification_test.go`, which snapshots the full flag set via `cmd.Flags().VisitAll` into a golden. Doing the same here is also a strictly stronger test -- a newly added flag shows up as a golden diff instead of silently passing.

helm/cluster-readiness-engine/tests/render_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+const helmTemplateTimeout = 30 * time.Second
+
+func TestHelmTemplateRendersConcurrencyArgs(t *testing.T) {
+	requireHelm(t)
+	chartDir := chartDir(t)
+
+	tests := []struct {
```
ndipebot: Same convention issue as `cmd/manager/main_test.go` -- this should be a `testutil.TestCaseParser` test rather than an inline table. Shelling out to helm is not a blocker: the harness only cares that the test body produces `tc.Actual`, so the `helmTemplate`/`managerArgs` helpers can stay exactly as they are. One testdata case per scenario (`defaults/`, `overrides/`) with the `--set` list in `input.yaml`, and `expected.json` holding the full rendered args array. That last part is also a better assertion than the two `slices.Contains` checks: any unintended change to the manager args becomes a visible diff instead of passing silently.

helm/cluster-readiness-engine/tests/render_test.go
```diff
@@ -0,0 +1,125 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), helmTemplateTimeout)
+	defer cancel()
+
+	args := make([]string, 0, 3+2*len(set))
+	args = append(args, "template", "render-test", chartDir)
+	for _, value := range set {
+		args = append(args, "--set", value)
+	}
+
+	cmd := exec.CommandContext(ctx, "helm", args...)
```
ndipebot: Two mechanical problems with the exec approach:

The go test cache can't see files that only a subprocess reads, so after one green run, edits to `deployment.yaml` or `values.yaml` don't invalidate anything. I reproduced this: delete both concurrency args from the template, rerun `go test ./helm/cluster-readiness-engine/tests/`, and you get `(cached)` PASS. It's not just local `make test` either -- on go1.26 `-coverprofile` is in the cacheable flag set, so the CI invocation can also serve a stale result on a chart-only change. Easiest fix is to have the test read the chart inputs in-process (`os.ReadFile` the template and values files, even just to discard them) so the cache keys on their contents.

Separately, `CombinedOutput()` feeds helm's stderr into the YAML decoder below. helm writes non-fatal warnings (deprecations, coalesce notices) to stderr while still exiting 0, so the first such warning from a future helm version turns into a confusing decode error here. `cmd.Output()` for the stream, keep stderr for the failure message.

helm/cluster-readiness-engine/tests/render_test.go
```diff
@@ -0,0 +1,125 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package tests
```
ndipebot: Could this live in `test/helm/` instead? There's no Go code anywhere under `helm/` today, and putting a Go package inside the chart is what forced the `.helmignore` carve-out in this PR. From `test/helm/` the `chartDir()` hop just becomes `../../helm/cluster-readiness-engine`, `go list ./...` still picks it up, and the `.helmignore` entry goes away. It also avoids any confusion with helm's own `templates/tests/` convention.

helm/cluster-readiness-engine/templates/deployment.yaml
```diff
@@ -32,6 +32,8 @@ spec:
         - --metrics-bind-address=:{{ .Values.metrics.port }}
         - --leader-elect
         - --health-probe-bind-address=:8081
+        - --max-concurrent-reconciles={{ .Values.manager.maxConcurrentReconciles }}
```
ndipebot: If someone nulls one of these (`--set manager.maxConcurrentReconciles=null` -- the same idiom our own Tiltfile uses for `imagePullSecrets`), helm drops the key and this renders `--max-concurrent-reconciles=` with an empty value. pflag then fails to parse the empty int before RunE ever runs, so the manager exits on a strconv error and crash-loops, and the friendly `validate()` message never gets a chance to fire. Worth a guard: either `{{ required }}` so it fails at render time with a real message, or wrap the two args in `{{- with }}` so a null falls back to the binary's built-in defaults.

test/uat/tilt/dra-stub/deploy.sh
```diff
@@ -68,6 +68,9 @@ rules:
 - apiGroups: ["resource.k8s.io"]
   resources: ["resourceclaims/status"]
   verbs: ["update", "patch"]
+- apiGroups: ["resource.k8s.io"]
+  resources: ["resourceclaims/binding"]
```
ndipebot, line 72: This rule is legit for 1.36 -- `resourceclaims/binding` is the new beta (on-by-default) authorization subresource, and the stub's `status.allocation` patch would come back Forbidden without it (https://kubernetes.io/docs/concepts/security/hardening-guide/dynamic-resource-allocation/). The problem is nothing in this repo actually runs 1.36: `install-uat-tools` pins kind v0.30.0, `setup-test-uat` calls `kind create cluster` with no `--image`, and kind v0.30.0 defaults to `kindest/node:v1.34.0`. So the headline "Kubernetes 1.36 ResourceClaim binding" fix in this PR only gets validated on dev machines whose local kind is newer.

Can we pin the node image (or bump KIND_VERSION to v0.32+, whose default is v1.36.1) in this same PR so CI covers the path we're claiming to fix? Otherwise a regression here stays invisible until someone's local run breaks.

test/uat/tilt/dra-stub/main.go
```diff
@@ -140,6 +140,7 @@ func reconcileResourceClaims(ctx context.Context, client dynamic.Interface) {
 		patch := map[string]any{
 			"status": map[string]any{
 				"allocation": map[string]any{
+					"allocationTimestamp": metav1.Now().Format(time.RFC3339Nano),
```
ndipebot, line 146: As far as I can tell this field is decorative. `allocationTimestamp` only matters as the timeout anchor for allocations whose devices carry BindingConditions (KEP-5007), and the stub never sets any, so nothing reads it. On the 1.34 cluster CI runs, the field is feature-gated off and the API server drops it from the write entirely -- harmless, since the idempotency check keys on the whole allocation map, but it means this line does nothing anywhere we test. I'd either drop it or add a comment saying it's informational only, so nobody later assumes the scheduler depends on it.

ndipebot (changes requested):
Went through this carefully since it flips every controller from serial to concurrent. The core of the change holds up: I dug through all six reconcilers for shared state that would start contending at 10/5 workers and came up empty. The goodput/bandwidth maps are all mutex-guarded, `getLogFetcher()` returns its fallback without writing the field (so no lazy-init race), the job controller's detector cache takes a lock, and every child Create path tolerates AlreadyExists, which matters now that two Workflows can race on a shared dependency resource. The zero-value passthrough is fine too: controller-runtime falls back to the manager-level default when the option is <= 0, and the integration suite already sets a global MaxConcurrentReconciles of 5, so envtest behavior doesn't change.

Two things I'd like addressed before merge, plus smaller comments inline:

1. Both new test files are table-driven tests. We just finished converting the whole repo off that pattern in #172 (merged the day before this review, so the branch probably predates it), and both files convert cleanly to `testutil.TestCaseParser` + testdata golden files. Specifics inline.

2. The "Kubernetes 1.36 ResourceClaim binding" fixes are never exercised by our CI. The UAT job pins kind v0.30.0 and `setup-test-uat` passes no node image, so CI runs a 1.34 cluster where the new RBAC rule is unused and the `allocationTimestamp` field gets silently dropped by the API server. The RBAC rule itself is correct for 1.36, so this is a coverage problem rather than a correctness one -- details on the deploy.sh comment.

(Your local `make test-uat-run` result squares with this, for what it's worth: recent kind releases default to a 1.36 node image, so the fix is real on your machine. I just want CI to see the same thing.)

cmd/manager/main_test.go
kaynetu: Addressed in 92824ac: both manager tests now use testutil.TestCaseParser, with substring error fixtures and a full sorted flag-default golden.

helm/cluster-readiness-engine/tests/render_test.go
kaynetu: Addressed in 92824ac: the Helm test now uses testutil.TestCaseParser and compares the complete rendered manager argument list for default and override cases.

helm/cluster-readiness-engine/tests/render_test.go
kaynetu: Addressed in 92824ac: requireChartInputs reads Chart.yaml, values.yaml, and deployment.yaml in-process, while helmTemplate uses Output with a separate stderr buffer.

helm/cluster-readiness-engine/tests/render_test.go
kaynetu: Addressed in 92824ac: the package now lives under test/helm, and the chart-specific .helmignore exception was removed.

helm/cluster-readiness-engine/templates/deployment.yaml
kaynetu: Addressed in 92824ac: required now rejects null values during rendering, and null-core/null-measurement golden cases verify the errors while ordinary installs retain the chart defaults.

test/uat/tilt/dra-stub/deploy.sh
kaynetu: Addressed in 92824ac: CI now installs Kind v0.32.0, whose default is Kubernetes 1.36.1, and KIND_NODE_IMAGE documents local 1.34.8/1.36.1 selection: https://github.com/kubernetes-sigs/kind/releases/tag/v0.32.0

test/uat/tilt/dra-stub/main.go
kaynetu: Retained with evidence in 92824ac: Kubernetes v1.36.1 PreBind sets a missing AllocationTimestamp before UpdateStatus, which reproduced the immutable-allocation failure, so the stub sets it initially: https://github.com/kubernetes/kubernetes/blob/v1.36.1/pkg/scheduler/framework/plugins/dynamicresources/dynamicresources.go#L1487-L1499

ndipebot: Confirmed -- you're right, withdrawing this one. I read the v1.36.1 source: bindClaim fills a missing AllocationTimestamp and calls UpdateStatus (it's at L1603-1615 in that blob, your link lands on the defer a few lines up in the same function), and since a populated status.allocation is whole-struct immutable, that write gets rejected and PreBind fails. With both gates default-on in 1.36, setting the timestamp at allocation time like the stub does is the only order that works -- dropping it like I suggested would have broken scheduling on exactly the cluster CI now runs. The comment you added explains it well. Thanks for the source pointer.

cmd/manager/main_test.go
```diff
@@ -0,0 +1,86 @@
...
+	})
+}
+
+func TestControllerConcurrencyFlagDefaults(t *testing.T) {
+	p := testutil.TestCaseParser{
+		Subdir:         "concurrency-flag-defaults",
+		ExpectedSuffix: ".json",
+	}
+	p.TestDir(t, func(tc *testutil.TestCase) error {
+		cmd := newRootCommand()
```
ndipebot: Non-blocking, but worth knowing about: this panics the moment anyone adds a second case dir under `testdata/concurrency-flag-defaults/`. `newRootCommand()` binds the zap flags onto the global `flag.CommandLine`, so the second `TestDir` iteration dies with `flag redefined: zap-devel` -- I reproduced it by just copying the `default` case. Passes today only because there's exactly one case. Hoisting `newRootCommand()` out of the closure (or binding zap to a fresh FlagSet) defuses the trap for whoever adds a case later.

cmd/manager/main_test.go
```diff
@@ -0,0 +1,86 @@
...
+
+		opts := controllerConcurrencyOptions{
+			maxConcurrentReconciles:            input.MaxConcurrentReconciles,
+			measurementMaxConcurrentReconciles: input.MeasurementMaxConcurrentReconciles,
+		}
+		if err := opts.validate(); err != nil {
+			return err
+		}
+
+		data, err := json.MarshalIndent(input, "", "  ")
```
ndipebot: Tiny one: `tc.Actual` here is the parsed input marshaled straight back out, so `defaults/expected.json` is a byte-for-byte echo of `input.yaml` -- the golden can't fail unless the input changes, and regenerating it is a no-op. Encoding the outcome instead (even just `{"valid": true}`) would make the expected file actually assert something.

test/helm/render_test.go
```diff
@@ -0,0 +1,140 @@
...
+		t.Fatal("runtime.Caller failed")
+	}
+	return filepath.Clean(filepath.Join(filepath.Dir(filename), "..", "..", "helm", "cluster-readiness-engine"))
+}
+
+// requireChartInputs reads the chart files that determine the manager args in
+// process. Go's test cache cannot observe files read only by the Helm subprocess.
+func requireChartInputs(t *testing.T, chartDir string) {
+	t.Helper()
+	for _, name := range []string{"Chart.yaml", "values.yaml", filepath.Join("templates", "deployment.yaml")} {
```
ndipebot: These three files cover everything the manager args depend on, so this does fix the case I flagged. One gap remains: the test's pass/fail also depends on the rest of the chart rendering at all (`managerArgs` decodes every doc), and the cache can't see those files -- I broke `_helpers.tpl` and a warm `go test ./test/helm/` kept serving the cached pass. `filepath.WalkDir` over the chart dir instead of the fixed list would close it completely. Fine as a follow-up.

ndipebot (dismissed):
Verified everything against 92824ac on the merged head: reran the converted tests (all green), re-rendered the chart for the default/override/null cases (nulls now fail at render time with the right messages, `0` still falls through to the binary's validate() as intended, so the README claim stays coherent), and checked the UAT run on this head actually pulled `kindest/node:v1.36.1` and passed -- the 1.36 path is genuinely covered now, not just claimed. Nice touch documenting KIND_NODE_IMAGE in CONTRIBUTING; the version claims there check out against the kind v0.32.0 release.

The allocationTimestamp thread was yours to win -- I read the 1.36.1 scheduler source and conceded on the thread. Good catch reproducing that.

Left three nits on the new tests. None of them block the merge.

ndipebot (commented):
some nits

cmd/manager/main_test.go
kaynetu: Fixed in 05fffe6: newRootCommand() is now created before TestDir, so additional cases reuse one command and do not re-register the global zap flags.

cmd/manager/main_test.go
kaynetu: Fixed in 05fffe6: the successful case now emits {"valid": true}, so expected.json asserts the validation outcome instead of echoing input.yaml.

test/helm/render_test.go
kaynetu: Fixed in 05fffe6: requireChartInputs now uses filepath.WalkDir and os.ReadFile on every chart file, so changes to helpers and other templates invalidate the Go test cache.

ndipebot (approved):
/lgtm
/approve

PR #207, opened 2026-08-21
fix(cli): report certification state on wait timeout

### LINKED ISSUE #79: certification run --wait: a timeout prints no report, writes no --results-file, and does not say the run continues
When `ncrectl certification run --wait --timeout` hits its timeout, three things happen at once, all in the same code path.

Observed on a two-node A100 cluster, `--timeout 15m --results-file /root/p71.json`:

```
[watch] communication/nccl-all-reduce: InProgress (15m0s)
Error: certification did not complete within 15m0s (ran for 15m0s)

$ ls /root/p71.json
ls: cannot access ...: No such file or directory

$ kubectl -n p71 get pods
...launcher-0-0   1/1   Running   0   15m
...node-0-0       1/1   Running   0   15m
...node-0-1       1/1   Running   0   15m
```

## 1 and 2 — no report, and no results file

`watchCertification` returns `nil` for the cert on timeout, and the whole report phase is gated on that being non-nil (`pkg/certification/certification.go:821-826`):

```go
finalCert, waitErr := watchCertification(ctx, wc, cfg.cert.Name, cfg.namespace, cfg.timeout, out)

// --- Report phase (only when we have a cert; nil on timeout or watch failure) ---
if finalCert != nil {
    handleReport(ctx, wc, finalCert, cfg.resultsFile, waitErr, out)
}
```

So a timeout produces no printed report **and** no `--results-file`, even though the Certification exists in the cluster and its partial status is readable.

This looks unintended rather than by design: `handleReport` already takes `waitErr` and branches on it, sending output to stderr when the wait failed. The code is set up to report a failed wait; it just never gets a cert to report on.

The flag help says only `requires --wait`, and `--wait` was supplied. A script reading that path cannot distinguish "timed out" from "never ran" — it gets no file either way.

**Suggested fix:** on timeout, `Get` the Certification and pass it to `handleReport` with the existing `waitErr`. The partial report is the useful artifact — it shows which categories finished. If a partial report is not wanted, having `--results-file` write something explicit (a `RUNNING`/`TIMEOUT` result) would still beat writing nothing.

## 3 — the timeout stops the watch, not the run

The certification keeps running; pods keep holding GPUs. That is defensible — the CR is the source of truth, and `--cleanup` exists — but nothing in the message says so, and `--cleanup` is not the default.

An operator who scripts `--wait` and moves on after the error leaves a run consuming GPUs with nothing watching it. On a small cluster that blocks the next run.

**Suggested fix:** documentation-level. Have the timeout message say the run continues and name the two follow-ups, in the style of the existing `WaitingForNodes` message:

```
certification did not complete within 15m0s (ran for 15m0s). The certification
is still running; check with `kubectl get certification p71 -n p71`, or stop it
with `kubectl delete ns p71`.
```

## Why one issue

All three are the same moment in the same path, and 1 and 2 share a fix. Splitting them would mean three issues pointing at adjacent lines. Happy to split if maintainers prefer.

## Found

Hardware test pass on 2026-08-08. The run itself overran because of test parameters, not a CRE defect — but the timeout handling is what made it hard to tell.

kaynetu:
## Summary

- Retrieve the latest Certification when `nvcrectl certification run --wait` reaches its timeout, then print the partial report and honor `--results-file` before returning the timeout error.
- Explain that a run without `--cleanup` continues in the cluster, including commands to watch it, print an updated report, or stop it.
- Print the report before optional cleanup, bound post-timeout API reads, and handle terminal, deleted, and temporarily unreachable Certification states explicitly.
- Add focused lifecycle tests, golden files for operator-facing timeout output, and CLI documentation.

## Type of Change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected

- [ ] API / CRDs
- [ ] Catalog / Workloads
- [x] CLI (nvcrectl)
- [ ] Helm / Deployment
- [x] Documentation / CI
- [ ] Other: ____________

## Testing

- [x] `make ci`
- [x] Focused timeout and cleanup-ordering regression tests
- [x] No breaking changes
- [ ] Manual cluster testing repeated after implementation; behavior is covered with fake-client and command-level tests

## Risk Assessment

Low. Timeout still returns a nonzero error and the default resource lifecycle is unchanged: without `--cleanup`, the Certification remains the source of truth and continues running. With `--cleanup`, the partial report is emitted before deletion. Post-timeout snapshot reads are capped at 30 seconds and degrade to explicit guidance if the Certification cannot be retrieved.

## Checklist

- [x] Self-review completed
- [x] `make manifests generate` run; no `*_types.go` files changed
- [x] New unit-test golden files added; integration golden output is unchanged
- [x] Documentation updated
- [x] Ready for review

Closes #79


pkg/certification/certification_test.go
```diff
@@ -447,6 +506,247 @@ spec:
...
+
+	gotErr := finishCertificationWait(context.Background(), wc, cfg, nil, waitErr)
+
+	assert.Same(t, waitErr, gotErr)
+	assert.Empty(t, out.String())
+	_, err := os.Stat(resultsFile)
+	assert.True(t, os.IsNotExist(err))
+}
+
+func TestFinishCertificationWaitReportsRetrievalFailure(t *testing.T) {
```
ndipebot: This one and `TestFinishCertificationWaitReportsUnknownState` below are snapshot-shaped -- they probe operator-facing output with Contains/NotContains -- and they exercise the same function whose output this PR already goldens in `finish-certification-wait-timeout/`. Two more case dirs on that parser would cover them: extend the input struct with knobs the callback already has machinery for (`certExists: false` -> empty fake client, `getError: "API server unavailable"` -> wrap in certificationGetErrorClient) and let expected.txt pin the full output. That also makes the NotContains checks unnecessary -- a golden asserts absence by exact match. The resultsFile-not-written check can stay as an in-callback assertion, same as your other golden tests here.

For the record, the other five plain funcs look fine to me under the convention's carve-outs (concurrency guard, bool pin, single-value asserts, empty-output, ordering).

pkg/certification/certification.go
```diff
@@ -995,14 +1001,91 @@ func executeCertificationRun(cfg *certRunConfig) (pipelineErr error) {
 		_, _ = fmt.Fprintf(out, "Waiting up to %s (derived from category timeouts; pass --timeout to override).\n", cfg.timeout)
 	}
 	finalCert, waitErr := watchCertification(ctx, wc, cfg.cert.Name, cfg.namespace, cfg.timeout, out)
+	pipelineErr = finishCertificationWait(ctx, wc, cfg, finalCert, waitErr)
+	return pipelineErr
+}
+
+// finishCertificationWait reports the best available state after the watch
```
ndipebot, line 1008: Flagging so it doesn't get lost rather than as a blocker: `workloadrun run --wait` still has the exact bug #79 describes -- `watchWorkloadRun` returns nil on timeout (workloadrun.go:884) and the report/results-file block is gated on `finalRun != nil` (workloadrun.go:846-856), so its users get no partial report, no results file, and no guidance while the run keeps holding GPUs. Now that certification handles timeout well, the sibling command's silence is a visible inconsistency. A follow-up issue pointing at this function as the template would do; also happy if you'd rather port it in this PR.

pkg/certification/certification_test.go
```diff
@@ -25,6 +35,55 @@ import (
...
+		lines = append(lines, line)
+	}
+	return strings.TrimSpace(strings.Join(lines, "\n")) + "\n"
+}
+
+func (c *certificationDeadlineClient) Get(
+	ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption,
+) error {
+	if _, ok := ctx.Deadline(); ok {
+		c.sawDeadline = true
```
ndipebot: This records sawDeadline whenever the ctx has *a* deadline -- but an already-expired context has one too, so the test can't tell a live 30s-bounded ctx from a dead one. And the fake client never consults the context at all (I checked controller-runtime v0.24.1 -- no ctx.Err/Done anywhere in fake/client.go), so if a future refactor ever passes the expired watch-scoped ctx into finishCertificationWait, production degrades to the 'could not retrieve' warning while this test and the pipeline test stay green. Cheap fix: also assert `ctx.Err() == nil` and that the deadline is in the future. That pins the liveness invariant the whole feature hangs on, not just boundedness.

pkg/certification/certification.go
```diff
@@ -1195,6 +1307,9 @@ func processWatchEvents(
 				return cert, true, fmt.Errorf("certification failed")
 			}
 
+		case <-ctx.Done():
```
ndipebot, line 1316: Narrow race worth a code comment: once the deadline passes, this case is permanently ready, and select picks uniformly among ready cases -- so a terminal Succeeded event in flight on the watch channel at that instant can lose the coin flip. Pre-PR the event always got processed and the command exited 0. Post-PR you recover the cert (the post-timeout Get finds it terminal and prints the PASSED report with the 'completed while the timeout was being handled' line) but still return waitErr, so a CI gate keying on exit code fails a certification that actually passed at the boundary. If nonzero-on-timeout is the intended contract even then -- defensible, #79 asks for a nonzero exit -- a short comment saying so would stop someone from 'fixing' it later. Otherwise consider returning nil when the retrieved cert is terminal-Succeeded.

pkg/certification/certification.go
```diff
@@ -1072,6 +1155,34 @@ func generateCertNamespace() string {
...
+}
+
+func certificationWaitContextError(ctx context.Context, timeout time.Duration, start time.Time) error {
+	if errors.Is(ctx.Err(), context.DeadlineExceeded) {
+		return &certificationWaitTimeoutError{
+			timeout: timeout,
+			elapsed: time.Since(start).Truncate(time.Second),
+		}
+	}
+	return ctx.Err()
```
ndipebot, line 1189: With --setup or --cleanup the ctx is signal-bound, so Ctrl-C during the watch now surfaces as `Error: context canceled`. The split itself is an improvement -- the old path printed the misleading did-not-complete-within message for an interrupt -- but the raw Go error is unpolished, and workloadrun prints `interrupted` for the same case (workloadrun.go:881-882). Returning fmt.Errorf("interrupted") here for context.Canceled would match.

docs/getting-started/first-certification.md
```diff
@@ -170,7 +170,7 @@ kubectl nvcre setup reset
...
 | `no GPU nodes match target` | All candidate nodes are cordoned, or the `nvidia.com/gpu.product` label is missing (GPU Operator not running). | `kubectl uncordon` the nodes you want tested, or install the GPU Operator. |
-| The watch stops at the timeout | The `--timeout` (derived from category timeouts when not set, 30 minutes minimum) expired. | Raise `--timeout`. Training runs with custom settings need hours. The Certification keeps running in the cluster; re-attach with `kubectl get certification <name> -n <namespace> -w`. |
+| The watch stops at the timeout | The `--timeout` (derived from category timeouts when not set, 30 minutes minimum) expired. | Read the partial report, then raise `--timeout` if needed. Without `--cleanup`, the Certification keeps running; monitor it with `kubectl get certification <name> -n <namespace> --watch`, print a new snapshot with `nvcrectl certification report <name> -n <namespace>` (which exits nonzero while the run is active), or stop it with `kubectl delete certification <name> -n <namespace>`. |
```
ndipebot: This guide declares at line 48 that it uses the `kubectl nvcre` form throughout, and step 6 shows this exact command as `kubectl nvcre certification report ...` -- the new row is now the one place giving an instruction in the `nvcrectl` spelling. (The CLI's printed hint rightly uses nvcrectl since the binary can't know how it was invoked; this is just the doc.)

pkg/certification/certification_test.go
```diff
@@ -25,6 +35,55 @@ import (
...
+) error {
+	return c.err
+}
+
+type certificationDeadlineClient struct {
+	client.WithWatch
+	sawDeadline bool
+}
+
+func normalizeCertificationReportOutput(output string) string {
```
ndipebot, line 73: Two small things here. The normalizer TrimSpaces every line, so the goldens store the guidance commands unindented while the code prints them with two-space indentation -- a spacing or separator regression in the new block would pass these tests. Print output is fully deterministic for these fixtures (boxWidth is a constant, and you already substitute the temp path), so preserving leading whitespace -- or skipping normalization -- would pin what users actually see. Also, this function landed between certificationDeadlineClient and its Get method; moving it out from between the type and its method would help the next reader.

ndipebot (changes requested):
The core fix is solid. I went hunting for the classic version of this bug -- passing the already-expired watch context into the post-timeout Get, which would make the whole feature silently degrade to the 'could not retrieve' warning in production while fake-client tests stay green -- and it's not there: the deadline is shadowed inside watchCertification and finishCertificationWait gets the live parent ctx. Report-before-cleanup ordering holds on all paths, the --results-file JSON is the same shape scripts get on success, the printed kubectl/nvcrectl commands all check out against the CRD and the report subcommand (including the exits-nonzero-while-running claim), and the package passes with -race. All three asks from #79 are covered.

One real ask before merge, the rest is smaller:

1. The two timeout-output tests that probe with assert.Contains should be golden cases on the parser this PR already built -- details inline. The other new test funcs are fine convention-wise.
2. `workloadrun run --wait` still has the exact bug #79 describes, one command over. Not this PR's scope, but now that certification behaves well on timeout the sibling's silence is a visible inconsistency -- comment inline near finishCertificationWait.

Nits inline on the deadline test's blind spot, a narrow exit-code race at the deadline boundary, the Ctrl-C message, one doc-convention slip, and the golden normalizer.

pkg/certification/certification_test.go
kaynetu: Addressed in 6c7a2e0: `certification-missing` and `state-unknown` are now cases in `TestFinishCertificationWaitTimeout`, with exact golden output and an in-callback results-file absence assertion.

pkg/certification/certification.go
kaynetu: Tracked separately in #219, which points to PR #207 as the bounded post-timeout retrieval and reporting pattern.

pkg/certification/certification_test.go
kaynetu: Addressed in 6c7a2e0: the test client now requires `ctx.Err() == nil` and a future deadline, and `TestFinishCertificationWaitBoundsPostTimeoutReads` asserts it.

docs/getting-started/first-certification.md
kaynetu: Addressed in 6c7a2e0: the troubleshooting command now uses `kubectl nvcre certification report`.

pkg/certification/certification.go
kaynetu: Documented in 6c7a2e0: the refreshed report may show a terminal state, but the command intentionally retains the timeout error so deadline expiry always exits nonzero.

pkg/certification/certification.go
kaynetu: Addressed in 6c7a2e0: `context.Canceled` now returns `interrupted`, covered by `TestCertificationWaitContextErrorCanceled`.

pkg/certification/certification_test.go
kaynetu: Addressed in 6c7a2e0: the deadline client type and method are contiguous, normalization preserves leading whitespace outside report boxes, and the goldens pin the two-space command indentation.

ndipebot (approved):
Verified 6c7a2e0 the same way as the first round -- against the pushed code, with mutation checks wherever the fix was about test strength. Everything holds:

- The two Contains-based tests are gone, fully subsumed by the `certification-missing` and `state-unknown` golden cases. Since golden comparison is exact-match, the old NotContains assertions are pinned even harder now, and state-unknown gained a results-file absence check the old test didn't have. Net coverage went up.
- The deadline-client fix has real teeth: I injected an expired reportCtx into finishCertificationWait and the test now fails exactly there (it passed before this change -- that was the blind spot).
- Same for the normalizer: de-indenting the guidance block in the code makes run-remains-active fail on precisely that line. And all four golden diffs in this commit are pure indentation-preservation -- nothing behavioral hiding in the churn.
- The exit-code comment sits right above the `return waitErr` and captures both halves of the trade-off, and the terminal-at-timeout cases pin the behavior with assert.Same, so nobody can accidentally 'fix' that race later.
- Ctrl-C now prints `interrupted`, matching workloadrun, and nothing anywhere consumes errors.Is(err, context.Canceled), so dropping the sentinel is safe. The nil-Err edge cases all check out -- every call site guards on ctx.Err() != nil first.
- The doc row now uses the `kubectl nvcre` form, the table is still well-formed, and #219 captures the workloadrun follow-up accurately with this PR as the template.

Nothing left from me. Nice round of fixes.

PR #287, opened 2026-09-02
feat: enable NCCL path-spread tuning for OCI GB300

mlsorensen:
Add NCCL_IB_ADAPTIVE_ROUTING=1 and set NCCL_IB_SPLIT_DATA_ON_QPS=1 in nccl/oci-gb300-roce-env.yaml, overriding the shared nccl/oci-roce-env.yaml default of 0. This fragment is always included after the shared one, and mpirun -x takes the last value, so the GB300 value wins. Applies to every OCI+GB300 consumer of the fragment: the communication entries and the nemotron training entries. Other platforms and non-GB300 OCI are unchanged.

Each GB300 node has four 800 Gbps RoCE rails. Untuned, all-reduce plateaued at roughly half the per-node fabric ceiling, which is the signature of traffic concentrating on a subset of ECMP paths rather than spreading across the rails. Enabling adaptive routing and splitting data across QPs raised peak bus bandwidth by about 70% on a multi-rack all-reduce, and neither setting regressed any configuration measured. NCCL_IB_QPS_PER_CONNECTION=8 was also evaluated and gave no improvement over the existing 4, so it is not adopted.

Verified with ncrectl certification render --platform oci: the OCI+GB300 entries emit adaptive routing once and the GB300 split value last.

## Related Issue

Closes #

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] API / CRDs
- [x] Catalog / Workloads
- [ ] CLI (nvcrectl)
- [ ] Helm / Deployment
- [ ] Documentation / CI
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

## Checklist
- [ ] Self-review completed
- [ ] Commits are signed off for the DCO (`git commit -s`)
- [ ] `make manifests generate` run (if `*_types.go` was modified)
- [ ] Golden files updated (if integration test output changed)
- [ ] Documentation updated (if needed)
- [ ] Ready for review


ndipebot (approved):
/lgtm
/approve

PR #288, opened 2026-09-02
feat(mcp): add a read-only MCP server for certification state

### LINKED ISSUE #242: [Feature]:  Add MCP Server
### Prerequisites

### Code of Conduct

### Feature Summary

Expose NVCRE test status and results to agents through an MCP server.

### Problem/Use Case

Answering "did this certification pass, and which nodes failed?" today means a person running the CLI and reading CRD status. It's a repetitive lookup loop that an agent could run — but only against a typed interface, not scraped CLI output. Operators increasingly have agents in their workflow and want readiness state available there.

### Proposed Solution

Ship an MCP server in-repo exposing read operations: list the catalog, get certification status, fetch a report, list failed nodes. Auth flows through the user's kubeconfig so an agent can never exceed the permissions of whoever runs it. Whether agents can trigger runs is a separate question — runs consume real GPU time, so we'd want to start read-only.

### Component

New Component

ntheanh201:
## Summary

Adds `nvcrectl mcp serve`: a read-only MCP server over stdio exposing four tools — `list_categories`, `get_certification_status`, `get_certification_report`, `list_failed_nodes` — so an agent can answer "did this certification pass, and which nodes failed?" against a typed interface instead of scraped CLI output.

The load-bearing decision is that **every certification verdict is projected from `report.Build`** rather than re-derived from the CR. An earlier draft of this change walked the Certification itself, and drifted from the report in three ways before anything noticed:

- it never applied the `PASSED` → `INCOMPLETE` downgrade `report.Build` makes when a Workflow excluded nodes ([report.go:308](https://github.com/NVIDIA/cluster-readiness-engine/blob/main/pkg/report/report.go)), and carried no `excludedNodes` field — so a run that left 8 of 40 nodes untested was reported to an agent as `PASSED`, in the cheaper tool an agent reaches for first;
- it returned the raw `InProgress` category status where the report says `Running`;
- it deduplicated failed nodes on a different key than `report.CertFailedNodes`.

Projecting from the report makes agreement structural rather than a convention two code paths have to maintain. `TestStatusAgreesWithReport` pins it across every fixture — the golden files could not, since they record each tool independently and a divergence sits unnoticed in two blocks 60 lines apart.

Stdio only: the agent spawns the process, which avoids introducing a network listener and the bearer-token/OAuth design that would come with it. No tool creates, mutates, or deletes anything, and none triggers a run — `TestListTools` fails if a fifth or mutating tool is ever added. ADR-075 records the reasoning.

## Related Issue

Closes #242

## Type of Change
- [x] ✨ New feature
- [x] 📚 Documentation

## Component(s) Affected
- [x] CLI (nvcrectl)
- [x] Documentation / CI

## New dependency — needs a maintainer call

This adds **`github.com/modelcontextprotocol/go-sdk v1.7.0`**, the first MCP dependency in `go.mod`. It lands in the `nvcrectl` binary, not the controller.

- License is Apache-2.0 / MIT (the project is mid-relicense from MIT; both are recorded verbatim in `THIRD_PARTY_NOTICES.md`).
- New transitive modules: `google/jsonschema-go`, `golang-jwt/jwt/v5`, `segmentio/asm`, `segmentio/encoding`, `yosida95/uritemplate/v3`.
- `THIRD_PARTY_NOTICES.md` appears to be hand-maintained (no generator target found), so I edited it by hand. If you have a `go-licenses`-style generator, my entry will want reformatting.

Happy to drop the SDK and hand-roll the JSON-RPC framing instead if a new dependency is unwelcome here.

## Also in this PR

Two corrections that came out of reviewing the above:

- **A security claim that was not true.** The docs, cobra help, and package doc each said the server "never reads in-cluster service account tokens". `pkg/kubeconfig` uses client-go's standard loading rules, which end in an in-cluster fallback — run `mcp serve` in a pod with no kubeconfig and it authenticates as that pod's ServiceAccount. The tools stay read-only, so this is a confidentiality claim rather than privilege escalation, but a guarantee that only holds outside a pod is worse than none in a feature aimed at agents. Now stated accurately, with the RBAC the tools need.
- **The docs name a sharp edge rather than hide it:** `pkg/report` returns empty results rather than errors when it cannot read a Workflow or the node-results ConfigMap, so a caller lacking that permission is told `"failedNodes": []` when the truth is "not allowed to look". Fine for a CLI a human reads; worth knowing for a tool an agent quotes. I did not change that shared behaviour in this PR.

## Testing

- [x] Tests pass locally

`make ci` passes on the branch: 25 packages, `golangci-lint` 0 issues, including the envtest integration suite.

Unit and golden coverage: `TestMCPTools` drives a real MCP session over in-memory transports (initialize → tools/list → tools/call) rather than mocking; `TestNotFound` asserts a tool error with the nvcrectl-style message; `TestListTools` pins the four-tool read-only surface; `TestStatusAgreesWithReport` asserts the two tools describe a Certification identically. A new `excluded-nodes` fixture covers the `INCOMPLETE` path — reintroducing the bug fails both that assertion and the golden.

**Validated against a live cluster** (Kubernetes v1.35.3, 2 nodes, GPU Operator present), not just envtest:

- With no NVCRE CRDs installed, both certification tools return a tool error naming the missing resource, rather than an empty result.
- With NVCRE v0.1.0 installed and real `Certification`/`Workflow` objects:

| Case | Result |
|---|---|
| Succeeded run with `excludedNodes` | `INCOMPLETE` from **both** tools, `excludedNodes` surfaced |
| Category `InProgress` | normalised to `Running` in **both** tools |
| Failed nodes from the gzip ConfigMap | decoded, with per-node reason (`HardwareFailureDetected`, `ThresholdViolation`) |
| `status.result` == `report.result` | agrees for every certification |

The cluster was returned to its prior state afterwards (`setup reset` plus manual removal of the namespaces and CRDs).

## Risk

Low, and additive. New package plus one new `nvcrectl` command group; no controller, CRD, or reconciler changes; nothing in the install path. The surface is read-only by construction and pinned by a test. The real risk is the new dependency, called out above.

Two things deliberately out of scope: no HTTP/SSE transport (it needs its own auth design), and no ability to trigger runs — runs consume real GPU time and that deserves its own decision about consumption and preemption.

## Checklist
- [x] Self-review completed
- [ ] `make manifests generate` run (if `*_types.go` was modified) — n/a, no API types changed
- [x] Ready for review

### DIFF
--- a/docs/cli-reference/mcp.md
+++ b/docs/cli-reference/mcp.md
@@ -1,0 +1,64 @@
+---
+title: nvcrectl mcp
+description: Serve read-only NVCRE certification state to MCP agents.
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+---
+
+
+## nvcrectl mcp serve
+
+Exposes NVCRE certification state to [Model Context Protocol](https://modelcontextprotocol.io) (MCP) agents over the stdio transport, so an agent can answer "did this certification pass, and which nodes failed?" without scraping CLI output.
+
+```bash
+nvcrectl mcp serve [flags]
+```
+
+### Tools
+
+The server is strictly read-only: no tool creates, mutates, or deletes a resource, and nothing triggers a run — runs consume real GPU time. Every tool carries the MCP `readOnlyHint` annotation.
+
+| Tool | Description |
+|------|-------------|
+| `list_categories` | List the certification catalog: every registered `domain/variant` category a Certification can run |
+| `get_certification_status` | Overall result (`PASSED`/`FAILED`/`RUNNING`), conditions, per-category state, and failed node names for one Certification |
+| `get_certification_report` | The full report that `nvcrectl certification report` prints — categories with metrics, bandwidth, cliques, diagnose results, and per-node results |
+| `list_failed_nodes` | Failed nodes for one Certification with per-node failure reason and message |
+
+The three certification-scoped tools accept `name` and `namespace` (default `default`).
+
+### Authentication
+
+All cluster access uses the kubeconfig of whoever launches the server, resolved with the standard client-go rules: the `--kubeconfig`/`--context` flags, then the `KUBECONFIG` environment variable, then `~/.kube/config`. The server holds no credentials of its own, never reads in-cluster service account tokens, and can therefore never exceed the permissions of the user who runs it.
+
+### Client configuration
+
+Most MCP clients spawn the server themselves over stdio. For example, in Claude Desktop or any client with a similar config format:
+
+```json
+{
+  "mcpServers": {
+    "nvcre": {
+      "command": "nvcrectl",
+      "args": ["mcp", "serve"]
+    }
+  }
+}
+```
+
+### Flags
+
+| Flag | Default | Description |
+|------|---------|-------------|
+| `--kubeconfig` | `~/.kube/config` | Path to the kubeconfig file to use for requests |
+| `--context` | current context | Name of the kubeconfig context to use |
+
+### Example
+
+```bash
+# Serve with the default kubeconfig (typical: launched by the MCP client)
+nvcrectl mcp serve
+
+# Serve against a specific kubeconfig and context
+nvcrectl mcp serve --kubeconfig /path/to/kubeconfig --context gpu-cluster
+```
--- a/docs/cli-reference/mcp.md
+++ b/docs/cli-reference/mcp.md
@@ -21,11 +21,15 @@ The server is strictly read-only: no tool creates, mutates, or deletes a resourc
 | Tool | Description |
 |------|-------------|
 | `list_categories` | List the certification catalog: every registered `domain/variant` category a Certification can run |
-| `get_certification_status` | Overall result (`PASSED`/`FAILED`/`RUNNING`), conditions, per-category state, and failed node names for one Certification |
-| `get_certification_report` | The full report that `nvcrectl certification report` prints — categories with metrics, bandwidth, cliques, diagnose results, and per-node results |
+| `get_certification_status` | Overall result (`PASSED`/`INCOMPLETE`/`FAILED`/`RUNNING`), conditions, per-category state, any nodes excluded from the run, and the unique names of failed nodes for one Certification |
+| `get_certification_report` | The full report that `nvcrectl certification report` prints — categories with metrics, bandwidth, cliques, and diagnose results |
 | `list_failed_nodes` | Failed nodes for one Certification with per-node failure reason and message |
 
 The three certification-scoped tools accept `name` and `namespace` (default `default`).
+
+`INCOMPLETE` means the run passed but left some targeted nodes untested — the Workflow excluded them, and `excludedNodes`/`exclusionReason` say which and why. Treat it as "not certified": nothing was observed about those nodes either way. Both `get_certification_status` and `get_certification_report` report it, because both derive the verdict from the same builder.
+
+`get_certification_status.failedNodes` is the unique node names, deduplicated across categories. `list_failed_nodes` returns one row per distinct failure reason instead, so a node that failed in several categories appears more than once — use the former for a count.
 
 ### Authentication
 
--- a/docs/cli-reference/mcp.md
+++ b/docs/cli-reference/mcp.md
@@ -29,7 +29,13 @@ The three certification-scoped tools accept `name` and `namespace` (default `def
 
 ### Authentication
 
-All cluster access uses the kubeconfig of whoever launches the server, resolved with the standard client-go rules: the `--kubeconfig`/`--context` flags, then the `KUBECONFIG` environment variable, then `~/.kube/config`. The server holds no credentials of its own, never reads in-cluster service account tokens, and can therefore never exceed the permissions of the user who runs it.
+All cluster access uses the kubeconfig of whoever launches the server, resolved with the standard client-go rules: the `--kubeconfig`/`--context` flags, then the `KUBECONFIG` environment variable, then `~/.kube/config`. The server holds no credentials of its own and adds no privilege of its own — it is exactly as authorized as the identity client-go resolves.
+
+<Warning>
+Client-go's loading rules end in an in-cluster fallback. Run `nvcrectl mcp serve` inside a pod with no kubeconfig and it authenticates as that pod's ServiceAccount, which may be broader than the operator running the agent. When you deploy the server in-cluster, bind its ServiceAccount to a role that grants no more than the reads below.
+</Warning>
+
+The tools need `get`/`list` on `certifications` and `workflows` in the target namespace, and `get` on the `configmaps` holding failed-node results. A caller missing the ConfigMap read still gets a successful response with an empty `failedNodes` list rather than an error, so grant the ConfigMap read explicitly — otherwise an agent can read "no nodes failed" from what is really "not allowed to look".
 
 ### Client configuration
 
--- a/pkg/mcpserver/server.go
+++ b/pkg/mcpserver/server.go
@@ -6,13 +6,9 @@
 //
 // The server answers "did this certification pass, and which nodes failed?"
 // from the same typed data sources the nvcrectl report command reads —
-// catalog.List, report.Build, and the failed-nodes ConfigMaps — so an agent
-// never has to scrape CLI output.
-//
-// Authentication flows strictly through the caller's kubeconfig (standard
-// client-go loading rules: --kubeconfig/--context flags, KUBECONFIG env, then
-// ~/.kube/config). The server can therefore never exceed the permissions of
-// whoever runs it, and holds no credentials of its own.
+// catalog.List, report.Build, and the failed-nodes ConfigMaps. Every
+// certification verdict is projected from report.Build rather than re-derived
+// from the CR, so a tool can never disagree with the report the CLI prints.
 //
 // The tool set is read-only by design (issue #242): no tool creates, mutates,
 // or deletes a resource, and nothing triggers a run — runs consume real GPU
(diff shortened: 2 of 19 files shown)


docs/cli-reference/mcp.md
```diff
@@ -0,0 +1,70 @@
...
+```
+
+### Tools
+
+The server is strictly read-only: no tool creates, mutates, or deletes a resource, and nothing triggers a run — runs consume real GPU time. Every tool carries the MCP `readOnlyHint` annotation.
+
+| Tool | Description |
+|------|-------------|
+| `list_categories` | List the certification catalog: every registered `domain/variant` category a Certification can run |
+| `get_certification_status` | Overall result (`PASSED`/`FAILED`/`RUNNING`), conditions, per-category state, and failed node names for one Certification |
```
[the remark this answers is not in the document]
ntheanh201: Fixed in e8a08ab — verified against the code first, and it was stale: `server.go:111` already returns `PASSED/INCOMPLETE/FAILED/RUNNING` while this table listed three of the four.

Rather than just add the word, the doc now says what `INCOMPLETE` *means* — the run passed but left targeted nodes untested, with `excludedNodes`/`exclusionReason` naming which and why — since a reader who sees the value without that context will read it as a pass, which is the mistake it exists to prevent.

While confirming the finding I found the adjacent row stale in the same way: it promised per-node results, but `report.Build` never populates `NodeResults` (only `pkg/workloadrun` does). The tool description in code was already corrected for that; the table had not been. Both rows now match the code, and I documented the deliberate difference between `get_certification_status.failedNodes` (unique names) and `list_failed_nodes` (one row per distinct reason), which is an easy over-count for a caller.

`make ci` green: 25 packages, lint 0 issues, `verify-doc-links` OK across 252 links.

ndipebot (commented):
Read through this carefully. The design call to project every verdict from `report.Build` instead of re-deriving from the CR is the right one, and it is the part that was most likely to go wrong. `TestStatusAgreesWithReport` holds that agreement structurally instead of leaving it as a convention two code paths have to remember, which the golden files could not have done on their own. All four operations from #242 are here, nothing extra, and the read-only surface is pinned by a test rather than by intent.

Two things I would like fixed before this merges. Both are about what an agent reads back, which is the whole point of the feature.

Also worth flagging separately: the `Instructions` string in `server.go` still says authentication uses the kubeconfig of whoever launched the server. That is the exact claim this PR corrects in the docs, the cobra help, and the package doc, and `Instructions` is the one string a client actually reads at session start. Small fix, but it is the copy that matters most.

docs/cli-reference/mcp.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Authentication
+
+All cluster access uses the kubeconfig of whoever launches the server, resolved with the standard client-go rules: the `--kubeconfig`/`--context` flags, then the `KUBECONFIG` environment variable, then `~/.kube/config`. The server holds no credentials of its own and adds no privilege of its own — it is exactly as authorized as the identity client-go resolves.
+
+<Warning>
+Client-go's loading rules end in an in-cluster fallback. Run `nvcrectl mcp serve` inside a pod with no kubeconfig and it authenticates as that pod's ServiceAccount, which may be broader than the operator running the agent. When you deploy the server in-cluster, bind its ServiceAccount to a role that grants no more than the reads below.
+</Warning>
+
+The tools need `get`/`list` on `certifications` and `workflows` in the target namespace, and `get` on the `configmaps` holding failed-node results. A caller missing the ConfigMap read still gets a successful response with an empty `failedNodes` list rather than an error, so grant the ConfigMap read explicitly — otherwise an agent can read "no nodes failed" from what is really "not allowed to look".
```
ndipebot, line 42: This RBAC list is incomplete for `get_certification_report`, and the gap lands an operator in exactly the trap the next sentence warns about.

`report.Build` reads more than certifications, workflows, and configmaps. It does `Get` on the nvcre `Job` and on `batch/v1` Jobs (`pkg/report/report.go:339`, `:568`), and `List` on `GoodputMeasurement` and `BandwidthMeasurement` (`:623`, `:637`) and on `Job` (`:1366`, `:1399`).

So if someone binds a role with precisely what this paragraph says, the report tool still returns a successful response, just with metrics, bandwidth, and diagnose data quietly absent. Same failure mode as the ConfigMap case called out right below, and arguably worse because the missing data is not a single obvious field.

The list also over-specifies in one place: `certifications` is only ever fetched by name, so it needs `get` and not `list`.

pkg/mcpserver/server.go
```diff
@@ -0,0 +1,335 @@
...
+	// Surfaced here so the cheaper tool cannot hide them from an agent.
+	TotalNodes      int             `json:"totalNodes,omitempty"`
+	ExcludedNodes   []string        `json:"excludedNodes,omitempty"`
+	ExclusionReason string          `json:"exclusionReason,omitempty"`
+	Conditions      []conditionInfo `json:"conditions,omitempty"`
+	Categories      []categoryState `json:"categories,omitempty"`
+	// FailedNodes is the unique node names that failed, deduplicated across
+	// categories. Use this for a node count; list_failed_nodes returns one
+	// row per distinct failure reason and so can repeat a name.
+	FailedNodes []string `json:"failedNodes,omitempty"`
```
ndipebot, line 284: Three tools describe "no failed nodes" three different ways, and an agent has no way to tell which one means zero.

- Here, `omitempty` means the `failedNodes` key is simply absent. See the `cert-pass` block in `testdata/mcp-tools/basic/expected.txt`.
- `report.CertReport.FailedNodes` has no `omitempty` and `CertFailedNodes` returns nil, so `get_certification_report` emits `"failedNodes": null`. See `testdata/mcp-tools/excluded-nodes/expected.txt`.
- `list_failed_nodes` emits `[]`.

The tool description a few lines up tells the agent to use this field for the unique node count, and a key that is not there reads as "unknown", not "zero". That is the one misreading this feature exists to prevent.

`TestStatusAgreesWithReport` cannot catch this, since a missing key and an explicit null both decode to a nil slice. Picking one shape and asserting the serialized bytes would.

Same thing applies to `categories` and `conditions` just above: both disappear entirely for a Certification that has not populated `CategoryStatuses` yet.

pkg/mcpserver/server.go
```diff
@@ -0,0 +1,335 @@
...
+}
+
+// getCertStatusOutput summarizes a Certification's overall and per-category
+// state without pulling measurement data. Every field except Conditions is
+// projected from report.Build, so it agrees with get_certification_report.
+type getCertStatusOutput struct {
+	Name      string `json:"name"`
+	Namespace string `json:"namespace"`
+	Result    string `json:"result"` // "PASSED", "INCOMPLETE", "FAILED", or "RUNNING"
+	// ExcludedNodes lists nodes that matched the target but were left
```
ndipebot, line 273: Minor: this comment describes `ExcludedNodes` but sits directly above `TotalNodes`, so godoc attaches it to the wrong field. Moving it down one line fixes it.

pkg/mcpserver/server.go
```diff
@@ -0,0 +1,335 @@
...
+			FailedNodes:     rep.FailedNodes,
+		}
+		for _, c := range cert.Status.Conditions {
+			out.Conditions = append(out.Conditions, conditionInfo{
+				Type:    c.Type,
+				Status:  string(c.Status),
+				Reason:  c.Reason,
+				Message: c.Message,
+			})
+		}
```
asivanadi0, line 177: result and conditions can disagree on the INCOMPLETE path, and the excluded-nodes golden already shows it: result is INCOMPLETE while conditions still carry Succeeded=True / AllCategoriesSucceeded.

That is faithful to how report.Build works (projection over CR state), but it is a sharp edge for an agent. The cheaper tool an agent reaches for first now returns both signals in one payload; anything that keys off conditions (or treats Succeeded=True as terminal pass) will report a clean pass for a run that left nodes untested — the exact failure mode the PR body is trying to close.

Two options that keep the projection story intact:
1. Document in the tool description that result is authoritative and must not be inferred from conditions (especially when excludedNodes is non-empty).
2. Or omit / annotate conditions when they disagree with the projected result so the cheaper tool cannot contradict itself.

TestStatusAgreesWithReport would not catch this today because it does not compare conditions at all.

pkg/mcpserver/server.go
```diff
@@ -0,0 +1,335 @@
...
+	return func(ctx context.Context, req *mcp.CallToolRequest, in certRef) (*mcp.CallToolResult, any, error) {
+		cert, err := store.certification(ctx, in)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		seen := map[string]bool{}
+		details := []failedNodeDetail{}
+		for _, cat := range cert.Status.CategoryStatuses {
+			for _, n := range report.FailedNodesFromRef(ctx, store.Client, cert.Namespace, cat.FailedNodesRef) {
```
asivanadi0, line 213: list_failed_nodes is the one certification tool that still walks CategoryStatuses + FailedNodesFromRef itself instead of going through report.Build (or a shared detail helper that Build also uses).

Today that walk matches CertFailedNodes, so names agree. The load-bearing claim in the ADR is that agreement with the report is structural rather than conventional — TestStatusAgreesWithReport pins that for status vs report, but not for this tool. If CertFailedNodes later grows a filter (skip recovered nodes, change the dedupe key, etc.), list_failed_nodes can drift without a failing test.

Would it be worth either (a) extracting a shared FailedNodeDetails helper that both CertFailedNodes and this tool call, or (b) extending the agreement test so the unique names from list_failed_nodes equal report.FailedNodes across the fixtures?

asivanadi0 (commented):
Review focused on the agent-facing contract and the report-projection invariant. The Build-backed status/report path and TestStatusAgreesWithReport are the right shape for this surface.

One numbering collision to resolve before merge: this PR and #309 both introduce ADR-075 (MCP vs scheduling stalls). One of them needs a new number or the design index will fork.

PR #294, opened 2026-09-02
build: track Makefile Go tool pins in a nested tools module

### LINKED ISSUE #279: Bring Makefile and workflow tool pins under automated updates
Part of #262. **Phase 0 — Foundation.** Depends on: #264. Size: S.

## Problem

#264 pinned every tool that contributes to a release signature, but only some of those pins are watched by an updater. Dependabot's `github-actions` ecosystem covers the SHA-pinned `anchore/sbom-action` and `cosign-installer`. It does **not** cover:

- the `*_VERSION` variables in the `Makefile` — `CONTROLLER_TOOLS_VERSION`, `ADDLICENSE_VERSION`, `GOLANGCI_LINT_VERSION` (pre-existing) plus `GOVULNCHECK_VERSION`, `GOTESTSUM_VERSION`, `GOCOVER_COBERTURA_VERSION` (added by #264)
- `COSIGN_VERSION` in `publish.yml`

Dependabot parses neither Makefiles nor workflow `env:` values. A pin nobody bumps is a pin that rots: `govulncheck` in particular is a scanner whose whole value is knowing about advisories published *after* we pinned it, so a stale pin quietly stops finding new vulnerabilities while the job stays green.

This gap is currently documented in a comment in `.github/dependabot.yml` and nowhere else. This issue is that comment's tracking artifact — filed so the gap does not disappear when #264 closes.

## Scope

Pick one and apply it consistently:

- **A `tools.go` with a `//go:build tools` tag**, blank-importing the Go tool packages so their versions live in `go.mod` and fall under Dependabot's existing `gomod` ecosystem, which already has cooldown rules configured. This is the idiomatic Go answer and needs no new tooling. It adds direct dependencies to `go.mod`, which `CLAUDE.md` gates behind maintainer approval — get that first.
- **Renovate with regex managers**, which can match `*_VERSION ?= vX.Y.Z` in the `Makefile` and `COSIGN_VERSION:` in a workflow directly. Handles every pin including cosign, but introduces a second updater alongside Dependabot.

`COSIGN_VERSION` needs an answer either way — `tools.go` does not reach it, since cosign is installed by an action rather than `go install`. Note that ADR-074 moves signing into a reusable `attest.yml` (#265), so decide where that value lives before wiring an updater to its current location.

## Acceptance criteria

- [ ] Every tool version pinned in `Makefile` or a workflow `env:` is raised automatically by an updater, or is explicitly documented as deliberately manual with the reason.
- [ ] A deliberately stale pin produces an update PR within one scheduled cycle.
- [ ] The `.github/dependabot.yml` comment listing the uncovered surface is updated or removed to match reality.
- [ ] Cooldown settings match the existing `gomod` policy so a malicious release is not adopted within hours.

asivanadi0:
Closes NVIDIA/cluster-readiness-engine#279

Draft of the tools.go / Dependabot work. Issue-first: waiting on maintainer ack (especially the go.mod expansion). See https://github.com/asivanadi0/cluster-readiness-engine/pull/1 for the full test plan.


go.mod
```diff
@@ -83,30 +272,51 @@ require (
...
+	mvdan.cc/unparam v0.0.0-20260823230713-2fa3d841b0c8 // indirect
 	sigs.k8s.io/apiserver-network-proxy/konnectivity-client v0.36.0 // indirect
 	sigs.k8s.io/jobset v0.12.0 // indirect
 	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
 	sigs.k8s.io/randfill v1.0.0 // indirect
 	sigs.k8s.io/structured-merge-diff/v6 v6.4.2 // indirect
 	volcano.sh/apis v1.15.1 // indirect
 )
+
+tool (
```
mchmarny: **MAJOR:** Putting the tool chain in the root module raises the require count from 103 to 305 — golangci-lint's tree, almost all `// indirect`. This module is importable: consumers use [`api/v1alpha1`](https://github.com/NVIDIA/cluster-readiness-engine/tree/991b91490f9475c5f018cf6bc4f7989f03b08378/api/v1alpha1) for the CRD types, so every consumer's module graph inherits those 200 entries and MVS considers their versions when resolving shared dependencies.

That cost is not needed to get the automation. A nested `tools/go.mod` keeps tool dependencies out of the root graph entirely, and Dependabot reaches it with a second `gomod` entry using `directory: /tools`. The Makefile's `gomodver` call changes only which file it reads.

Worth deciding before merge rather than after: once a consumer pins a version of this module, the published graph is what they resolved against.

Minimum correction: either move the tool requires to a nested `tools/go.mod` with a matching Dependabot directory entry, or state in the PR why the root-module graph is the accepted trade-off.

tools/tools.go
```diff
@@ -0,0 +1,28 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build tools
```
mchmarny: **MINOR:** With the `tool (` block present in go.mod, this build-tagged blank-import file looks redundant. `tool` directives are the Go 1.24+ mechanism and `go mod tidy` retains the requires they name; the `//go:build tools` pattern is what they replaced. This file's own comment says the `tool` lines are what keep tidy from dropping the requires, which argues the same thing.

Minimum correction: delete the file, run `go mod tidy`, and confirm the six requires survive. If they do, this also reverts the `tools` addition to the addlicense path list in the Makefile.

mchmarny (commented):
Comment: 1 MAJOR, 1 MINOR against e62cc58. Reviewed against a base of ce70f5f; the workflow comment hunks still apply to main.

go.mod
asivanadi0: Moved the tool dependencies into a nested tools/go.mod and added a Dependabot gomod entry at directory: /tools. Root module is back to 103 requires, so the tool graph stays out of consumers importing api/v1alpha1.

tools/tools.go
asivanadi0: Deleted tools/tools.go, ran go mod tidy in the tools module, and confirmed all six tool requires survive as indirect requirements. Also removed tools from the addlicense path list.

.github/dependabot.yml
```diff
@@ -34,17 +34,53 @@ updates:
...
+  # skip). Check during release prep:
+  #   - cosign_version / crane_version defaults on attest.yml workflow_call
+  #     inputs. Dependabot does not update workflow_call input defaults; cosign
+  #     governs the attestation bundle format, and crane's download is
+  #     digest-pinned for one version only.
+  #   - SYFT_VERSION + SYFT_SHA256 in release.yml: version and digest must move
+  #     together, and there is no atomic pair update for that.
+  #   - YQ_VERSION in publish-fern-docs.yml, KIND_VERSION / TILT_VERSION in
+  #     .github/actions/install-uat-tools, KWOK_VERSION in the Makefile --
+  #     GitHub-release binaries outside any module graph.
```
kaynetu, line 92: The manual-pin list is not the full surface #279 asked to track. Dependabot's `github-actions` ecosystem updates `uses:` SHAs only. These values are on this branch and will not move:

- `COSIGN_VERSION` in `release.yml` (`verify-release` env, passed to `cosign-installer` as `cosign-release`). This is the workflow `env:` pin the issue named. It left `publish.yml`, and it is still an env value.
- `cosign-release: v3.1.3` as a literal `with:` input in `docs-verify.yml`.
- `CRANE_VERSION` + `CRANE_SHA256` in `docs-verify.yml`. Same digest-pair constraint as crane in `attest.yml` and as syft, in a second copy.
- `version: v3.19.2` on `azure/setup-helm` in `ci.yml`, `release.yml`, and `uat.yml` (twice on current `main`).

`attest.yml`'s defaults are the signer. `release.yml` and `docs-verify.yml` are the verifiers. Bumping only the documented default leaves verification on a different cosign or crane. The Helm CLI pin can drift the same way across those three workflows.

Minimum correction: add each pin to this list with the reason it stays manual. For cosign and crane, better to have the verifier jobs read the same default `attest.yml` already documents, so there is one pin.

asivanadi0: Addressed in 5d68e846f123fa7685cb55d84e3b1dbb5af23568.

**Cosign / crane — unified (preferred path):** `attest.yml` remains the single pin (`workflow_call` defaults + `CRANE_SHA256`). `release.yml` `verify-release` and `docs-verify.yml` now read those values via `.github/actions/read-attest-pins`, so there is no second `COSIGN_VERSION`, no literal `cosign-release: v3.1.3`, and no duplicate crane digest pair on the verifiers.

**Helm CLI — centralized:** `ci.yml`, `release.yml`, and `uat.yml` install through `.github/actions/setup-helm`. Dependabot still moves the `azure/setup-helm` SHA; the CLI `version:` input lives in one place.

**Manual-pin list:** `.github/dependabot.yml` now documents the full #279 surface you named (attest.yml cosign/crane and that verifiers consume it, Helm `version:` in `setup-helm`, plus the existing GitHub-release / digest pins) with why each stays manual.

`test/releasepolicy/tool_pins_test.go` fails if a verifier re-embeds those versions, if helm is called beside the composite action, or if a pin drops off the dependabot comment.

PR #295, opened 2026-09-02
test: enforce release workflow policy invariants (#274)

### LINKED ISSUE #274: Workflow policy regression tests
Part of #262. **Phase 4 — Assure.** Depends on: #265, #270. Size: M.

> **Rescoped 2026-09-02.** Substantial policy-test infrastructure landed inside the PRs it protects (#282, #292, #293) rather than as this issue. `test/releasepolicy` now holds eight tests, each mutation-tested. But those mostly pin invariants this issue never listed. Measured against this issue's own scope, **four of five invariants remain uncovered** — they are the whole of the work below.

## Problem

Everything this epic builds lives in YAML and can be deleted by a well-meaning refactor. The failure is silent: signing steps that vanish do not break the build, they just stop signing. The release-time gate (#270) catches it — but at release time, on a red release, under time pressure. It should be caught in the PR that removes it.

## Already covered

Landed with the changes they protect; listed so this issue is not re-litigated:

| Test | Pins |
|---|---|
| `TestAttestValidationAccepts` / `Rejects` | `attest.yml`'s input validation, 33 input sets |
| `TestAttestWorkflowIsGatedToThisRepository` | the signer cannot run from a fork |
| `TestJobOutputReferencesResolve` | every `needs.<job>.outputs.<field>` resolves |
| `TestNoExpressionInterpolationInRunBlocks` | no `${{ }}` inside a `run:` body |
| `TestVerificationUsesExactIdentity` | no `--certificate-identity-regexp`, no bare `gh attestation verify` |
| `TestPublishedVerifyCommandsAreExact` | the commands we publish, not only the workflows |
| `TestPredicateCheckToleratesDuplicateAttestations` | the release gate's predicate parser |

That covers exactly one of this issue's original invariants: the exact-identity rule.

## Remaining scope

Four invariants, all currently true and none currently enforced. Each must be mutation-tested — reintroducing the defect has to fail the test, or the test is decoration.

1. **`attest.yml` is the sole signer.** Nothing else may invoke `cosign sign`, `cosign attest`, `cosign attest-blob`, or `actions/attest-build-provenance`. This is what makes the published identity contract true: users pin one workflow, so a second signer anywhere silently widens what that pin accepts. Also assert `attest.yml` stays `workflow_call`-only — any other trigger makes it directly runnable and its identity forgeable from a branch.

2. **Signing jobs request `id-token: write` and nothing more, and non-signing jobs do not request it at all.** The OIDC token is the credential the whole contract rests on; a job that does not sign has no business being able to mint one.

3. **No action is referenced by a mutable tag.** A `@v4` that moves is an unreviewed change to the release path. Every `uses:` outside a local `./` reference must be a 40-character SHA.

4. **The expected-asset list matches what the release actually uploads.** Four independent hardcoded lists exist — the `attest-binaries` matrix, the pre-publish bundle check, the digest map, and #270's gate — while assets are staged and uploaded by glob and derive from `NVCRECTL_PLATFORMS` in the Makefile. Adding a platform ships an unsigned binary and an unsigned SBOM with all four gates green, while the release notes state that every asset carries a Sigstore bundle. ADR-074 commits to this test by name. Found by review of #293 and deliberately not fixed there.

## Acceptance criteria

- [ ] Adding a second signing invocation outside `attest.yml` fails `make test`.
- [ ] Giving `attest.yml` a trigger other than `workflow_call` fails `make test`.
- [ ] Granting `id-token: write` to a job that does not sign fails `make test`.
- [ ] Replacing any pinned action SHA with a tag fails `make test`.
- [ ] Adding a platform to `NVCRECTL_PLATFORMS` without updating every asset list fails `make test`.
- [ ] Runs in the existing `ci.yml`; needs no network and no cluster.

asivanadi0:
Closes NVIDIA/cluster-readiness-engine#274

Draft of the remaining test/releasepolicy mutation tests (sole signer, id-token, SHA-pinned actions, NVCRECTL_PLATFORMS vs asset lists). Issue-first: waiting on maintainer ack. Fork PR: https://github.com/asivanadi0/cluster-readiness-engine/pull/2


test/releasepolicy/workflow_policy_test.go
```diff
@@ -0,0 +1,494 @@
...
+)
+
+const (
+	makefilePath       = "../../Makefile"
+	releaseWorkflow    = "../../.github/workflows/release.yml"
+	attestWorkflowName = "attest.yml"
+)
+
+// workflowFiles lists every workflow definition under .github/workflows.
+func workflowFiles(t *testing.T) []string {
```
mchmarny: **BLOCKER:** `workflowFiles` is already declared in this package. [`shell_scope_test.go:37`](https://github.com/NVIDIA/cluster-readiness-engine/blob/991b91490f9475c5f018cf6bc4f7989f03b08378/test/releasepolicy/shell_scope_test.go#L37) on `main` defines `func workflowFiles(t *testing.T) []string` with the same signature and an equivalent body, so adding this one makes `package releasepolicy` fail to compile and takes `make test` down with it.

The base here is `ce70f5f`, which predates #293 — that PR added `shell_scope_test.go` along with the helper.

Minimum correction: after rebasing onto `main`, delete this copy and call the existing helper. `onReleasePath` and the `releasePathWorkflows` set are also already there and may cover what the new tests need.

.github/workflows/release.yml
```diff
@@ -846,19 +846,25 @@ jobs:
         run: |
           tag="${TAG}"
           rm -rf verify-release && mkdir verify-release
+          # Static expected-asset list (ADR-074). Built assets come from
+          # NVCRECTL_PLATFORMS by glob; this list is what the gate actually
+          # checks. test/releasepolicy fails `make test` when they drift.
+          BINARIES="nvcrectl-linux-amd64 nvcrectl-linux-arm64 nvcrectl-darwin-amd64 nvcrectl-darwin-arm64"
```
mchmarny, line 998: **MAJOR:** This hunk edits the `Verify published release assets` step, which #293 deleted. It was replaced by the `verify-release` job, so the change cannot apply after a rebase.

The requirement it adds is already satisfied on `main`, which carries exactly one `BINARIES=` line listing all four binaries ([release.yml at 991b91490f94](https://github.com/NVIDIA/cluster-readiness-engine/blob/991b91490f9475c5f018cf6bc4f7989f03b08378/.github/workflows/release.yml)). I ran each of this PR's four extractors against `main`'s current `release.yml` and `Makefile`:

| Extractor | Result on main |
|---|---|
| `releaseGateBinaries` | 1 match, all 4 binaries |
| `attestBinariesMatrixSubjects` | matches the `- subject:` entries |
| `digestMapSubjects` | `for want in` present |
| `prePublishBundleSubjects` | `for f in nvcrectl-` present |

`NVCRECTL_PLATFORMS` is `linux/amd64,linux/arm64,darwin/amd64,darwin/arm64`, so the expected set matches.

Minimum correction: drop the `release.yml` change and rebase. `TestExpectedAssetListsMatchNVCRECTLPlatforms` should pass against `main` unchanged — worth confirming, since that makes this PR test-only.

mchmarny (commented):
Comment: 1 BLOCKER, 1 MAJOR against 2aefb51. Reviewed against a base of ce70f5f, which predates #292, #293, #297 and #298 on main.

test/releasepolicy/workflow_policy_test.go
asivanadi0: Deleted the duplicate `workflowFiles` and now call the one #293 added in `shell_scope_test.go`. The rebase also turned up a second redeclaration: `releaseWorkflow` was already in `predicate_parse_test.go` with the same value, so that const is gone too.

On `onReleasePath` / `releasePathWorkflows`: these checks are meant to scan every workflow, not only the release path, so narrowing them to that set would weaken what they enforce. Happy to change that if you would rather they be release-path-scoped.

.github/workflows/release.yml
asivanadi0: Dropped the release.yml hunk. #293 already removed the step it edited, and TestExpectedAssetListsMatchNVCRECTLPlatforms passes against unmodified main (single BINARIES= line, all four binaries, matching NVCRECTL_PLATFORMS). This PR is now test-only.

asivanadi0: Dropped the release.yml hunk. #293 already removed the step it edited, and TestExpectedAssetListsMatchNVCRECTLPlatforms passes against unmodified main (single BINARIES= line, all four binaries, matching NVCRECTL_PLATFORMS). This PR is now test-only.

test/releasepolicy/workflow_policy_test.go
```diff
@@ -0,0 +1,481 @@
...
+			} `json:"jobs"`
+		}
+		if err := yaml.Unmarshal(raw, &doc); err != nil {
+			t.Fatalf("parse %s: %v", path, err)
+		}
+
+		for jobName, job := range doc.Jobs {
+			for _, step := range job.Steps {
+				if attestProvenanceAction.MatchString(step.Uses) {
+					t.Errorf("%s: job %q step %q uses %s; provenance must be emitted by attest.yml via cosign, not actions/attest-build-provenance",
```
mchmarny: `lll`: 133 chars, limit is 120. Same fix needed at lines **106** (135), **184** (126), **321** (130) and **367** (131) — five in total, all `t.Errorf` format strings that wrap cleanly at the `+` concatenation the rest of this package uses.

test/releasepolicy/workflow_policy_test.go
```diff
@@ -0,0 +1,481 @@
...
+// workflowTriggers returns the `on:` block, tolerating YAML 1.1 turning a bare
+// `on:` key into the boolean true (the same hazard workflowCallOutputs faces).
+func workflowTriggers(raw []byte, t *testing.T) map[string]any {
+	t.Helper()
+
+	var doc map[string]any
+	if err := yaml.Unmarshal(raw, &doc); err != nil {
+		t.Fatalf("parse workflow triggers: %v", err)
+	}
+	for _, key := range []string{"on", "true"} {
```
mchmarny: `goconst`: the literal `"true"` appears 3 times and the package already defines `boolTrue` for it. Worth using here — this is the YAML 1.1 `on:` → boolean-true gotcha, and naming it is what makes that readable.

(`workflow_graph_test.go` hits the same trap in `workflowCallOutputs`, if a cross-reference is useful.)

test/releasepolicy/workflow_policy_test.go
```diff
@@ -0,0 +1,481 @@
...
+	if m == nil {
+		re = regexp.MustCompile(`(?m)^NVCRECTL_PLATFORMS\s*=\s*(.+)$`)
+		m = re.FindSubmatch(raw)
+	}
+	if m == nil {
+		t.Fatal("Makefile has no NVCRECTL_PLATFORMS assignment")
+	}
+
+	var out []string
+	for _, p := range strings.Split(strings.TrimSpace(string(m[1])), ",") {
```
mchmarny: `modernize`: `stringsseq` — `strings.SplitSeq` is preferred over ranging `strings.Split`. Same at line **404**, where it wants `strings.FieldsSeq`. Go 1.24+, already used elsewhere in this package.

test/releasepolicy/workflow_policy_test.go
```diff
@@ -0,0 +1,481 @@
...
+	}
+	return out
+}
+
+func attestBinariesMatrixSubjects(text string, t *testing.T) []string {
+	t.Helper()
+
+	// Limit to the attest-binaries job so image/chart matrices cannot pollute.
+	job := jobBody(text, "attest-binaries", t)
+	var out []string
```
mchmarny: `prealloc`: `out` can be preallocated with the known capacity.

mchmarny (commented):
Reviewed against `main` at 5ac169f. **All five of #274's acceptance criteria are genuinely enforced by this file** — I mutation-tested each one rather than reading it, since #274 itself says a test that does not fail on the reintroduced defect is decoration.

| Mutation | Result |
|---|---|
| Rogue `cosign sign` step added to `publish.yml` | caught — `job "tag" step "Rogue signer" invokes cosign sign` |
| Rogue `actions/attest-build-provenance` added to `release.yml` | caught — `provenance must be emitted by attest.yml via cosign` |
| `workflow_dispatch:` added alongside `workflow_call:` on `attest.yml` | caught |
| `id-token: write` granted to a job that does not sign | caught |
| `actions/checkout` SHA replaced with `@v7` | caught |
| Fifth platform added to `NVCRECTL_PLATFORMS` without updating the asset lists | caught |

The sole-signer criterion is covered on both of its paths (a `cosign` invocation in a `run:` body, and the provenance action in a `uses:`), which is the one I'd have expected to be weakest.

Also confirmed: no network calls, no cluster, passes cleanly alongside the existing `test/releasepolicy` suite (42s full package), so #274's last criterion — *"Runs in the existing `ci.yml`"* — is satisfied by construction.

**The only thing blocking this is lint** — 9 findings, all mechanical, no logic changes needed. `make lint` currently fails, which is presumably why it is still a draft. Specifics inline.

Context worth having: #305/#306/#307 landed the weekly image scan since this was opened, and the C4 pinning check is not hypothetical for it — dependabot moves `anchore/scan-action` weekly under `patterns: ["*"]`, and #306 had to add a runtime assertion precisely because nothing guaranteed that pin stays a pin. This test would cover that class directly.

test/releasepolicy/workflow_policy_test.go
asivanadi0: Wrapped all five long format/regexp lines with + concatenation (and a method-call break on the subject regexp) so they stay under the 120-char lll limit.

test/releasepolicy/workflow_policy_test.go
asivanadi0: Switched the on-key fallback to the package boolTrue constant for the YAML 1.1 on:→boolean case.

test/releasepolicy/workflow_policy_test.go
asivanadi0: Updated both loops to strings.SplitSeq and strings.FieldsSeq.

test/releasepolicy/workflow_policy_test.go
asivanadi0: Preallocated out from len(matches) before the append loop.

mchmarny:
Re-verified at `b3178f5` — the lint findings are cleared and the invariants still hold. **This is ready to merge from my side.**

`make lint`: 0 issues. `make test`: green.

The lint fixes touched real iteration (`SplitSeq`, `FieldsSeq`, and the `prealloc` restructure in `attestBinariesMatrixSubjects`), so I re-ran every mutation afterwards rather than assuming the behaviour survived. All six still fail as they should:

| Mutation | |
|---|---|
| Rogue `cosign sign` in `publish.yml` | caught |
| Rogue `actions/attest-build-provenance` in `release.yml` | caught |
| Second trigger on `attest.yml` | caught |
| `id-token: write` on a non-signing job | caught |
| `actions/checkout` unpinned to `@v7` | caught |
| Fifth platform in `NVCRECTL_PLATFORMS` | caught |

That is all five of #274's acceptance criteria, plus the sixth (runs in `ci.yml`, no network, no cluster) satisfied by construction.

I had prepared the same lint fixes locally before seeing yours land — they were effectively identical, so I have discarded mine rather than push over your branch. Nothing needed from me.

Only thing left is taking it out of draft. Worth landing soon: nothing on `main` enforces any of this today, and the pinning check in particular is live — dependabot moves `anchore/scan-action` weekly under `patterns: ["*"]`, and #306 had to add a runtime assertion precisely because nothing guaranteed that pin stays a pin.

mchmarny (dismissed):

lalitadithya (dismissed):

test/releasepolicy/workflow_policy_test.go
```diff
@@ -0,0 +1,487 @@
...
+//	.../attest.yml@refs/tags/<TAG>
+//
+// so any second workflow that invokes cosign sign/attest/attest-blob (or
+// actions/attest-build-provenance) silently widens what that pin accepts.
+// attest.yml itself must stay workflow_call-only: any other trigger makes the
+// signing identity reachable from a branch push or a dispatch.
+func TestAttestIsSoleSigner(t *testing.T) {
+	assertAttestIsWorkflowCallOnly(t)
+
+	for _, path := range workflowFiles(t) {
```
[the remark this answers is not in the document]
asivanadi0: Fixed in 9a1360e. TestAttestIsSoleSigner now also globs `.github/actions/*/action.yml` (same pattern as TestActionsAreSHAPinned) and parses composite `runs.steps`. Cosign/attest-build-provenance in a composite fails the same way as in a workflow (attest.yml still allowed for cosign). Mutation-verified: added a temporary `run: cosign sign ...` step to install-uat-tools/action.yml, confirmed the test failed, then reverted.

test/releasepolicy/workflow_policy_test.go
```diff
@@ -0,0 +1,487 @@
...
+				} `json:"steps"`
+			} `json:"jobs"`
+		}
+		if err := yaml.Unmarshal(raw, &doc); err != nil {
+			t.Fatalf("parse %s: %v", path, err)
+		}
+
+		for jobName, job := range doc.Jobs {
+			signs := jobSigns(job.Uses, job.Steps)
+			token, hasToken := job.Permissions["id-token"]
```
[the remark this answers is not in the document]
asivanadi0: Fixed in 9a1360e. TestIDTokenWriteOnlyOnSigningJobs now resolves effective permissions as workflow-level overridden by job-level when present (including `permissions: {}`), and `ghPermissions` accepts scalar `read-all`/`write-all` as well as the map form. A non-signing job with effective id-token write fails; a signing job without it fails. Mutation-verified: set workflow-level `id-token: write` on verify.yml with no job override (failed), and briefly confirmed scalar `permissions: write-all` is caught too; both reverted.

asivanadi0:
Ready for re-review at `9a1360e`. The push after your approvals only addressed CodeRabbit’s two Majors (sole-signer now covers composite actions; id-token check resolves workflow-level + scalar permissions), both mutation-verified. Prior lint/acceptance work unchanged. Approvals were dismissed by the new commit under stale-review protection.

lalitadithya (approved):

Issue #273, opened 2026-09-02
Daily release re-verification


mchmarny:
Part of #262. **Phase 4 — Assure.** Depends on: #270. Size: L.

## Problem

The release-time gate (#270) proves the release was correct at publication. It says nothing about ten minutes later. Release assets can be deleted or replaced by anyone with write access or a compromised token; a tag can be repointed; a registry tag can be moved to a different digest. None of that touches this repository, so nothing in CI would notice. The realistic detection window today is "a user reports it."

## Scope

Scheduled workflow, daily, on a cron offset from the release and UAT crons, plus `workflow_dispatch` — GitHub disables schedules after 60 days of repository inactivity, and a dispatch re-arms them as well as allowing an on-demand run after a release.

Resolve the latest release, re-run the full #270 verification suite against it, and on failure open or update a GitHub issue.

The design problem worth getting right is **not paging on a Sigstore outage**:

- Probe Sigstore liveness (TUF CDN *and* Fulcio — probing only one stays green through the other's outage) before classifying anything as a finding.
- Classify each failure as `security` (tamper) or `operational` (transport). Classification must be **demote-only**: an ambiguous case becomes a real finding reported as operational, never the reverse.
- Any failure occurring *before* the classifying step runs — checkout, release resolution, tool install — is operational by construction.
- `security` → open an issue immediately, `@`-mention maintainers, title carries the tag.
- `operational` → open a "degraded" issue only after 3 consecutive days.
- A clean run closes only the alert for the tag it actually verified. Once a newer release ships, the older tag stops being checked, so its issue must be closed by hand after remediation — say so in the issue body.
- A floor-tag constant: releases at or before it predate this epic's artifacts and must not be reported as missing them. Raise it only to correct history.

## Constraint from #265: more than one attestation per digest is normal

`attest.yml` retries every cosign call, and those retries are **not idempotent**: keyless cosign mints a fresh Fulcio certificate per invocation, so an attempt killed by its timeout *after* the registry push and Rekor entry landed will publish a second valid signature for the same digest on the next attempt. Re-running a job after a partial failure does the same.

This job must therefore treat "N > 1 attestations for a digest" as normal and verify that *at least one* satisfies the identity contract. Reporting a duplicate as tampering would page on a Sigstore flake — the precise false positive the demote-only classification rule exists to prevent.

- [ ] A digest carrying two valid attestations from the same identity verifies clean and files nothing.

## Non-release Rekor entries exist (from #265)

[...]

Issue #270, opened 2026-09-02
Release-time verification gate


mchmarny:
Part of #262. **Phase 2 — Verify.** Depends on: #266, #267, #268, #269. Size: M.

## Problem

`release.yml:274-304` is the only post-publish check, and it verifies SHA-256 digests and a version string. Every signing step this epic adds could silently no-op — wrong permissions, an expired OIDC token, a cosign flag rename — and the release would still publish green. Producing attestations without verifying them means the *first* party to discover they are broken is a user.

## Scope

Add a gate job that runs after everything is published and before the release is marked non-draft. It must verify over the same channel a user takes: unauthenticated download for public assets, authenticated fallback while the repo is private. The existing pattern at `release.yml:280-294` already handles this and should be reused, not reinvented.

Verify, using the **exact** identity from the contract — never a regexp:

- image index: provenance verifies **from the registry copy** (`cosign`, or `gh attestation verify --bundle-from-oci` — never the bare `gh attestation verify`, which reads GitHub's API copy and would pass on an image whose GHCR referrer is missing), signature verifies, and the **predicate body** is validated — `externalParameters.repository` is this repository, `externalParameters.ref` is `refs/tags/<TAG>`, and `resolvedDependencies[].digest.gitCommit` is the commit the released tag points at. A valid signature says who signed and on which ref; only the predicate says what was built. A tag moved after the build still passes an identity-only check.
- image per-platform: CycloneDX attestation verifies on each platform digest, and the two SBOMs are not identical
- chart: signature and provenance verify on the chart digest
- each binary + installer: blob attestation verifies against the downloaded bytes
- each `.cyclonedx.json`: sibling bundle verifies
- every expected asset is present — derived from a **static expected-asset list**, not from the release's own inventory. Deriving expectations from the artifact under test means deleting one asset makes the check verify the survivors and report clean.

## Acceptance criteria

- [x] Every artifact contract row in #262 is verified by the gate.
- [x] The expected-asset list is a constant in the workflow, reviewed on change.
- [x] Deleting one asset before the gate fails the release. — `TestGateRejectsMissingAssets` (#302) removes each of the 25 published assets in turn, and zero-lengths each `-s` site, asserting the gate refuses and names the missing asset.
[...]

Issue #262, opened 2026-09-02
Epic: Software supply chain security — verifiable release artifacts


mchmarny:
## Problem

Every artifact this project publishes is consumed by operators who run it with cluster-admin-adjacent RBAC on GPU fleets. Today a consumer cannot answer "did NVIDIA build this, from what source, containing what?" for most of what we ship.

The container image is signed and carries an SBOM. Nothing else is. And the image's own coverage has defects:

- **Wrong SBOM subject.** `publish.yml:96` generates one SBOM against the multi-architecture *index* digest. Syft resolves an index to a single platform, so the `linux/arm64` image is described by a `linux/amd64` SBOM. An SBOM describes exactly one root filesystem; its subject must be a per-platform manifest digest.
- **No build provenance anywhere.** No SLSA statement links any artifact to the source commit, the workflow, or the run that produced it. A signature proves "someone in this repo signed it," not "this came from tag `vX.Y.Z`."
- **The published verification command does not verify what it claims.** `SECURITY.md:68` instructs users to pass `--certificate-identity-regexp='https://github.com/NVIDIA/cluster-readiness-engine'`. That regexp is unanchored and workflow-agnostic. `publish.yml` also runs on push to `main` (`publish.yml:7-9`), signing `main-<sha>` images under `.../publish.yml@refs/heads/main` — which that regexp accepts. A user following our own documentation will accept an unreleased main-branch build as a release.
- **The signing toolchain is not pinned.** `anchore/sbom-action@v0.24.2` (`publish.yml:94`) is a floating tag while every other action in the repo is SHA-pinned. `cosign-installer` (`publish.yml:86`) has no `cosign-release`, so the cosign version — and with it the attestation bundle format and Rekor backend — is an inherited default that moves without a commit here.
- **Nothing ever verifies.** The post-publish gate at `release.yml:274-304` checks SHA-256 digests and the version string. It never runs `cosign verify`. A release where signing silently no-ops ships green.
- **The release notes make a claim the release never checked.** `release.yml:226` tells users the image is `.../manager:<tag>`. That image was published by a different workflow run (`publish.yml`) with no digest handoff. `release.yml` never confirms it exists, never learns its digest, and never verifies its signature.

Unsigned entirely: the Helm chart, all four `nvcrectl` binaries, and the `installer` script that our own README tells users to `curl | bash`.

## Status — 2026-09-04

[...]

PR #302, opened 2026-09-03
test: prove the release gate rejects absence, not only tampering

mchmarny:
Part of #262. Advances #270.

## Why

`v0.2.0-rc.2` proved the release gate rejects a **tampered** artifact. Nothing had ever shown it rejects one that is **absent**, and absence takes different branches than tampering does — the branches that run when a signing step silently no-ops, which is the failure the gate exists for.

A skipped job does not fail its caller, so disabling a signing job publishes an unsigned artifact on a green release. A missing asset makes a verification loop verify the survivors and report clean. Both read correctly. So did every other check in this epic that turned out not to work.

## What

Two tests that execute the gate's own shell against stubs, extracted from `release.yml` rather than copied so they cannot drift from what they cover. No production behaviour changes.

- **`TestGateRejectsMissingAssets`** — removes each of the 25 assets a release publishes, one at a time; zero-length files once per `-s` site; and asserts what the gate actually asked cosign to do: every bundle verified, against the pinned identity and issuer, with the right predicate type. Passing is not evidence of verifying.
- **`TestGateRejectsSkippedSigningJobs`** — drives the three job-result guards through `skipped`, `failure` and `cancelled` for every dependency each reads. `cancelled` is there because a guard written as `== skipped` would let a timed-out run through.

Mutation-verified against **21 weakenings of `release.yml`, enumerated in the file** so the claim stays checkable rather than resting on a number in a commit message.

## The first draft of this was decoration, and that is worth reading

Six mutations escaped it. The stub keyed on cosign's trailing argument, and **both** verification loops pass the same binary name there — so either loop's `cosign` call could be replaced with `if true`, the gate no longer verifying signatures at all, with every case green. Dropping `--certificate-identity` was invisible for the same reason, which is the exact hole this epic was filed to close. And the assertion that the gate names the failing asset was satisfied by the defect it excluded, since that name is a prefix of the next asset in the list.

The stub now records cosign's arguments instead of only its exit status. All six are caught.

## Also here

ADR-074 claimed *"the release gate and the daily job reject a legacy-format bundle rather than falling back to reading it."* Neither half is true — no such logic exists anywhere in the release path, and the daily job is unbuilt (#273). Corrected in place rather than left for whoever relied on it.

## What this does not close

`#270`'s **"skipping any signing step"** criterion says *step*. These guards read `needs.<job>.result`, so a signing step disabled inside a job that still succeeds passes all three. That case is caught later — `verify-release` re-checks signatures against the registry — but that path is untested here, and the criterion's *"test by dispatch with the step disabled"* prescribes it. The box stays unticked.

Out of scope but found while reviewing: `create-github-release`'s *"Require a bundle for every asset"* pre-flight check has no test at all. Same defect class, one job earlier.

## Verification

`make lint`, `make build`, `make test` (incl. the 180s integration suite) green. 63 subtests, ~4s, no network, no cluster, hermetic. Passes under bash 3.2 and 5.x.


ndipebot (approved):
/lgtm
/approve

PR #308, opened 2026-09-04
docs+test: correct ADR-074 D2 and gate reusable-workflow boundary

### LINKED ISSUE #301: ADR-074 D2: correct the Build Level reasoning, and gate the reusable-workflow boundary
Part of #262. **Phase 4 — Assure.** Depends on: #265, #266. Size: M. **Not blocking the v0.2.0 release.**

## Problem

ADR-074 decision D2 records that this project targets **SLSA Build L2**, with this reasoning:

> It does **not** reach L3: the builds stay in the callers, and L3 requires the build itself to run in the protected reusable workflow. Reaching L3 is deferred to its own decision.

That framing is imprecise, and the imprecision matters because it makes L3 look further away than it is.

SLSA v1.0's Build L3 requirement is that **provenance is unforgeable by the build process**. Reading D2 as "the build must run in the reusable workflow" conflates that with builder isolation, which is a related but distinct property. On the first requirement, `attest.yml` already does most of the work:

- It is a genuine reusable workflow (`on: workflow_call`), invoked as `uses: ./.github/workflows/attest.yml` from `release.yml`. It is not an inlined job.
- Because the attestation is minted inside the called workflow, the Fulcio certificate names **`attest.yml`**, not the caller. Verified against a real `v0.2.0-rc.1` bundle: the SAN is `https://github.com/NVIDIA/cluster-readiness-engine/.github/workflows/attest.yml@refs/tags/v0.2.0-rc.1`.
- Every value in the predicate comes from trusted context **inside** that workflow — `GITHUB_REPOSITORY`, `GITHUB_REF`, `GITHUB_SHA`, `GITHUB_SERVER_URL`, `GITHUB_RUN_ID` — and not from a single caller-supplied input. A caller cannot dictate what the predicate says.
- There is already a guard refusing to proceed if `builder_id` resolves to `attest.yml`, on the grounds that "the attestor cannot be the builder".

So the build steps in `release.yml` cannot forge or tamper with the provenance today. What is genuinely absent is **builder isolation**: the build runs in the caller, `runDetails.builder.id` truthfully names the caller, and a compromised `release.yml` would produce a bad artifact that `attest.yml` would then honestly attest.

That honesty is deliberate — `attest.yml` states that naming the attestor as the builder "would make the predicate false on its face" — and any change here must not trade a true predicate for a higher number.

## Scope

**1. Decide, and amend ADR-074 D2.**

Either correct D2's stated reasoning while keeping the L2 claim, or move to L3. If moving, the concrete path is that `build-image.yml` is *already* a reusable workflow that builds the image. If provenance were minted there, the build and the attestation would share one protected boundary and `builder.id` would be both honest and L3-qualifying. That is an image-only change; the `nvcrectl` binaries are built and attested from the release job and would stay at L2, so any claim we publish has to be **per artifact**, not a single project-wide level.

**2. Gate the property, regardless of which level we claim.**

This is the cheaper half and worth doing even if D2 keeps L2. Nothing currently asserts that provenance generation stays inside a reusable workflow. A refactor that inlined the attestation step into `release.yml` would demote the property with every test still green — the only external symptom is that identity-pinned verification quietly starts failing for consumers. `test/releasepolicy` exists for exactly this class of silent invariant, and the assertions are mechanical:

- `attest.yml` declares `on: workflow_call` and is never invoked as an inlined job
- no `cosign sign` / `cosign attest` / `cosign attest-blob` invocation exists outside `attest.yml`
- the predicate is built only from `GITHUB_*` context, never from `inputs.*`
- the `builder_id != attest.yml` guard is present

**3. Record that the level is only observable if the verifier pins the workflow.**

The isolation property is invisible unless verification names the signing workflow. Verification that omits it still succeeds, but proves a weaker claim. This project already publishes exact identities naming `attest.yml`, and `TestVerificationUsesExactIdentity` and `TestPublishedVerifyCommandsAreExact` already forbid the loose form — so the mechanism is in place. What is missing is stating plainly, in `SECURITY.md` and `docs/operations/verifying-artifacts.md`, that pinning the workflow identity *is* the check that distinguishes the levels. Without that, a reader who drops the identity gets a green result and a weaker guarantee than they think.

## Caveats to carry into the decision

- **A same-repo reusable workflow isolates from the build steps, not from write access.** `uses: ./…` means caller and attestor live in one repository, so the practical strength of the boundary rests on branch protection over `.github/workflows/attest.yml`. Any L3 claim should say so rather than imply stronger isolation than exists.
- **Per-artifact, not project-wide.** Whatever is decided, `SECURITY.md` must not state a single level for everything we publish.

## Acceptance criteria

- [ ] ADR-074 D2 amended: either its reasoning corrected with L2 retained, or a recorded decision to move images to L3 with the binaries' level stated separately.
- [ ] Tests in `test/releasepolicy` fail when provenance generation moves out of a reusable workflow, when a signing call appears outside `attest.yml`, or when a predicate field is sourced from a caller input. Each mutation-verified.
- [ ] `SECURITY.md` and the verification page state which artifact carries which level, and that pinning the workflow identity is what makes the level checkable.
- [ ] No claim of a level that is not enforced by a test.

asivanadi0:
Closes #301

## Summary

Corrects ADR-074 decision D2's Build Level framing and gates the reusable-workflow boundary so a silent demotion fails `make test`.

### D2 decision taken

**Keep SLSA Build L2** for every artifact published today (binaries **and** images). Do **not** move images to L3.

Corrected reasoning (replacing the imprecise "build must run in the reusable workflow" framing):

- Provenance is already **unforgeable by the build process**: `attest.yml` is `workflow_call`-only, Fulcio names `attest.yml`, predicate origin fields come from `GITHUB_*` / `github.workflow_ref` (not `inputs.*`), and a `builder_id != attest.yml` guard keeps the predicate honest.
- What is absent is **builder isolation** (build still runs in the caller). That is why we stay at L2 rather than inflate to L3.
- Claims are **per-artifact**, not project-wide.
- Same-repo caveat stated: `uses: ./…` isolates from build steps, not from write access; strength rests on branch protection over `attest.yml`.

## Changes

1. **ADR-074** — amend D2 / decision-4 Build Level section with the corrected L2 reasoning, per-artifact framing, and same-repo caveat.
2. **`test/releasepolicy`** — new `attest_boundary_test.go`:
   - `TestAttestIsSoleSigner` — `workflow_call`-only, invoked via `uses: ./…`, no `cosign sign|attest|attest-blob` / `actions/attest-build-provenance` outside `attest.yml` (workflows + composites)
   - `TestAttestPredicateUsesOnlyTrustedContext` — provenance origin fields from trusted context only
   - `TestAttestBuilderIdGuardRejectsAttestorAsBuilder` — attestor-as-builder guard present
3. **`SECURITY.md`** + **`docs/operations/verifying-artifacts.md`** — state which artifact is Build L2, and that pinning the workflow identity is what makes the level checkable. Level claims point at the tests that enforce them (`TestVerificationUsesExactIdentity` / `TestPublishedVerifyCommandsAreExact` already cover the pin).

## Test plan

- [x] `make lint` — 0 issues
- [x] `go test ./test/releasepolicy/...` — pass
- [x] `make test` — pass

## Mutation notes

| Invariant | How mutated | Caught by |
|---|---|---|
| `attest.yml` workflow_call-only | added `push:` trigger | `TestAttestIsSoleSigner` |
| sole signer | inserted `cosign sign` into `release.yml` | `TestAttestIsSoleSigner` |
| invoked as reusable workflow | renamed all `uses: ./…/attest.yml` | `TestAttestIsSoleSigner` |
| predicate from `GITHUB_*` only | `--arg repo` / env from `inputs.*` | `TestAttestPredicateUsesOnlyTrustedContext` |
| `CALLER_WORKFLOW_REF` from `github.workflow_ref` | wired to `inputs.subject_name` | `TestAttestPredicateUsesOnlyTrustedContext` |
| `builder_id != attest.yml` guard | deleted guard block | `TestAttestBuilderIdGuardRejectsAttestorAsBuilder` |

**Already covered (not re-invented here):** exact identity pin — `TestVerificationUsesExactIdentity` and `TestPublishedVerifyCommandsAreExact` (cited from SECURITY / verifying-artifacts). Note: open PR #295 (`TestAttestIsSoleSigner` for #274) overlaps the sole-signer / workflow_call half; this PR adds the #301-specific predicate + builder_id gates and the D2 docs. Maintainers may want to reconcile the shared sole-signer helper when both land.

## Risk

Docs + policy tests only. No workflow behaviour change. Low release-path risk; worst case is a false-positive policy failure on an intentional attestation refactor.

Draft — do not mark ready / do not merge from this PR without review.


test/releasepolicy/attest_boundary_test.go
```diff
@@ -0,0 +1,201 @@
...
+			if !strings.HasPrefix(job.Uses, "./") {
+				t.Errorf("%s: job %q calls attest.yml as %q; same-repo reusable "+
+					"calls must use the ./ form so the call is a workflow_call boundary",
+					base, jobName, job.Uses)
+			}
+		}
+	}
+	if found == 0 {
+		t.Fatalf("no release-path workflow calls ./.github/workflows/attest.yml; " +
+			"attestation must stay behind a reusable-workflow boundary")
```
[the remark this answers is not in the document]
asivanadi0: Fixed in f267ac4. The assertion now tracks foundInWorkflow per caller and fails for each of release.yml / publish.yml / attest-selftest.yml that lacks a local ./.github/workflows/attest.yml call; the ./ reusable-workflow form check is unchanged. Mutation: removing the attest calls from release.yml alone fails naming release.yml.

asivanadi0:
Friendly ping — this is rebased onto current `main` (incl. the grpc bump), DCO/lint/`make test` green locally, and ready for a look when you have a moment.

Happy to address any review notes. Also still needs a trustee `/ok to test` so the NVIDIA runner checks can run (same gate as #295).

Thanks —
cc @lalitadithya @ndipebot @mchmarny

asivanadi0:
Rebased onto `main` (includes #316 failureLog). Branch still needs human review and trustee `/ok to test` before CI runs.

asivanadi0:
Rebased onto latest main.

asivanadi0:
Hi @lalitadithya @ndipebot @mchmarny — gentle bump when you have a moment. ADR-074 D2 / attest boundary (issue #301) is rebased onto latest main and ready for review. It still needs a trustee `/ok to test` so the NVIDIA runner checks can run. Happy to address any notes. Thanks!

asivanadi0:
Rebased onto `main` (includes the #325–#328 wave). Head moved `cd31edb` → `97cec77`; another trustee `/ok to test` is needed for NVIDIA runners.

asivanadi0:
Rebased onto `main` (post-#329/#331/#332). Fresh trustee `/ok to test` needed if the SHA moved.

docs/designs/074-supply-chain-attestation.md
```diff
@@ -66,11 +66,19 @@ This contract is published to users and enforced by us. Both the release-time ga
...
+
+Two consequences follow. First, no artifact or document may claim L3 — not the ADR, not the release notes, not `SECURITY.md` — until a recorded decision moves a specific artifact across that line. Second, the same-repo reusable-workflow caveat must be stated whenever this boundary is discussed: `uses: ./…` isolates attestation from the caller's *build steps*, not from *write access*. Caller and attestor live in one repository, so the practical strength of the boundary rests on branch protection over `.github/workflows/attest.yml`. Claiming L3 without saying so would imply stronger isolation than exists.
```
ndipebot: I don't think branch protection is the control this boundary actually rests on, at least not for the release identity. `release.yml` runs on `push: tags: 'v*'`, and a local `uses: ./.github/workflows/attest.yml` resolves the called workflow from the same commit as the caller, so the SAN everyone pins (`attest.yml@refs/tags/${TAG}`) is minted from whatever commit the pushed tag points at. Someone with tag-push rights but no ability to merge to protected `main` can push `refs/tags/vX.Y.Z` at a commit carrying a doctored `attest.yml`, and every published verify command passes; branch protection over `attest.yml` never engages.

The control that gates this is tag protection / repository rulesets over `v*` tags, which none of the docs mention. Branch protection is the right answer for the `@refs/heads/main` dev-image identity (`publish.yml` only runs on `main` pushes), but this sentence and its echoes at `SECURITY.md` line 84 and `docs/operations/verifying-artifacts.md` lines 102-104 are written about the boundary in general. I'd name tag rulesets here, and ideally add one for `v*`.

test/releasepolicy/attest_boundary_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+// TestAttestBuilderIdGuardRejectsAttestorAsBuilder pins the guard that keeps
+// runDetails.builder.id honest. Naming attest.yml as the builder would make the
+// predicate false on its face and is exactly the trade ADR-074 D2 refuses when
+// it keeps L2 rather than inflating to L3.
+func TestAttestBuilderIdGuardRejectsAttestorAsBuilder(t *testing.T) {
+	step := provenanceStep(t)
+
+	const needle = `"/.github/workflows/attest.yml"`
+	if !strings.Contains(step.Run, needle) && !strings.Contains(step.Run, "'/.github/workflows/attest.yml'") {
```
ndipebot: `TestAttestBuilderIdGuardRejectsAttestorAsBuilder` greps for the guard's text but never executes it, so a guard that stops rejecting still passes. I checked: deleting the `exit 1` at `attest.yml:480`, flipping the `==` at line 478 to `!=`, and commenting out the whole block each leave this test, and every other test in the package, green, because the needles all survive (they even survive inside comments).

That's exactly the "a weakened guard would not break a build, it would just stop rejecting things" failure this package's extract-and-execute convention exists to prevent, and the step is directly runnable the way `attest_guards_test.go` runs the validate step: stub `CALLER_WORKFLOW_REF` and the `GITHUB_*` env, assert exit codes. Doing that would also cover the fail-closed `case` guard at `attest.yml:473-476`, which currently has no coverage at all.

Minor: the quoted-needle alternatives at lines 142-148 are redundant, both quoted forms contain the bare substring, so this reduces to one `strings.Contains`.

test/releasepolicy/attest_boundary_test.go
```diff
@@ -0,0 +1,201 @@
...
+//
+// A refactor that wired --arg repo ${{ inputs.repository }} (or passed the same
+// through env) would still produce a green release whose predicate said whatever
+// the caller asked. This test fails that change before it ships.
+func TestAttestPredicateUsesOnlyTrustedContext(t *testing.T) {
+	step := provenanceStep(t)
+
+	// The step must not pull workflow_call inputs into the provenance surface.
+	for envName, envVal := range step.Env {
+		if strings.Contains(envVal, "inputs.") {
```
ndipebot: Scanning step-level `env` and the run block for the literal `inputs.` misses caller data laundered through `needs.validate.outputs.*`, which is the exact route this step already uses for `SUBJECT_KIND`. I verified three variants against this test: adding `UNTRUSTED_KIND: ${{ needs.validate.outputs.subject_kind }}` to the step env and folding it into the `--arg repo` value keeps every gate green; so does routing it into `CALLER_WORKFLOW_REF`; and a job-level `env` containing a literal `inputs.` reference also passes, because only `step.Env` is read here even though `runSteps` in `shell_scope_test.go` already merges workflow, job, and step env.

Worth scanning all three env levels and treating `needs.*.outputs.*` values that trace back to `inputs.*` as untrusted too.

test/releasepolicy/attest_boundary_test.go
```diff
@@ -0,0 +1,201 @@
...
+			Jobs map[string]struct {
+				Uses string `json:"uses"`
+			} `json:"jobs"`
+		}
+		if err := yaml.Unmarshal(raw, &doc); err != nil {
+			t.Fatalf("parse %s: %v", base, err)
+		}
+		foundInWorkflow := 0
+		for jobName, job := range doc.Jobs {
+			if !localAttestUses.MatchString(strings.TrimSpace(job.Uses)) {
```
ndipebot: A `uses:` that doesn't match the `^\./`-anchored `localAttestUses` regex is skipped by this `continue`, so the `!strings.HasPrefix(job.Uses, "./")` branch below can never fire and the test only asserts that at least one `./` call exists per file.

I tried switching `publish.yml`'s `attest-index` job to `NVIDIA/cluster-readiness-engine/.github/workflows/attest.yml@<40-hex-sha>` while the sibling jobs kept the `./` form: this test stays green (the remaining `./` calls keep the count above zero), `TestActionsAreSHAPinned` accepts the sha pin, and nothing else in either policy package objects. So an artifact could be attested by a stale pre-hardening `attest.yml` with the whole suite green. Filtering with the existing `isAttestWorkflowCall` and then asserting the `./` form on every match closes the hole and makes the dead branch live.

docs/designs/074-supply-chain-attestation.md
```diff
@@ -66,11 +66,19 @@ This contract is published to users and enforced by us. Both the release-time ga
...
 
-Reaching L3 later means moving image, chart, and binary generation into the protected reusable workflow. That is a larger restructure than this record covers — it rewrites the build path rather than adding to it — and it should be its own decision once this contract is in place and stable. Recorded as deferred, not rejected.
+- It is a genuine reusable workflow (`on: workflow_call`), invoked as `uses: ./.github/workflows/attest.yml` from the callers. It is not an inlined job, so the Fulcio certificate names **`attest.yml`**, not the caller.
+- Every value that shapes the provenance predicate comes from trusted context **inside** that workflow — `GITHUB_REPOSITORY`, `GITHUB_REF`, `GITHUB_SHA`, `GITHUB_SERVER_URL`, `GITHUB_RUN_ID`, and the caller workflow ref GitHub itself sets — and not from a caller-supplied `inputs.*` field. A caller cannot dictate what the predicate says about origin.
```
ndipebot: "Every value that shapes the provenance predicate" overstates it: `externalParameters.subjectKind` is populated via `--arg kind "${SUBJECT_KIND}"`, and `SUBJECT_KIND` is `${{ needs.validate.outputs.subject_kind }}`, a pass-through of the caller-supplied `inputs.subject_kind`. It's enum-validated, but it is a caller-chosen `inputs.*` value shaping the predicate, so the absolute wording is falsified by `attest.yml` as it stands.

The Rationale bullet at line 158 ("makes provenance unforgeable by the build process", unqualified) inherits the overclaim; `SECURITY.md` survives only because it scopes its version to origin fields. I'd scope this sentence the same way.

docs/designs/074-supply-chain-attestation.md
```diff
@@ -66,11 +66,19 @@ This contract is published to users and enforced by us. Both the release-time ga
...
+
+A concrete path to per-artifact L3 for images exists and is recorded as deferred, not rejected: `build-image.yml` is already a reusable workflow that builds the image; if provenance were minted there, build and attestation would share one protected boundary and `builder.id` would be both honest and L3-qualifying. That would be an image-only change — `nvcrectl` binaries are still built and attested from the release job and would stay at L2 — which is why every published claim must stay per-artifact. Until that decision is taken, both binaries and images remain Build L2 under the corrected reasoning above.
```
ndipebot: Two costs of this deferred path are worth recording here so the follow-on doesn't discover them after building it.

1. The identity mechanics: if provenance were minted in `build-image.yml` with the derivation `attest.yml` uses today (`github.workflow_ref`, see `attest.yml:477`), `builder.id` would name the top-level `release.yml`/`publish.yml`, not `build-image.yml`. The honest, L3-qualifying identity comes from `job_workflow_ref`, the pattern in GitHub's SLSA Build L3 guide this ADR already cites in References.
2. Wherever the cert is issued, signing inside `build-image.yml` moves the Fulcio SAN off `attest.yml`, splitting the single pinned identity that decision 3 and Alternative 2 exist to protect: image attestations would verify only under a `build-image.yml@...` identity while the chart and binaries stay under `attest.yml@...`, every published image verify command breaks, and `TestAttestIsSoleSigner` fails.

Deferred-not-rejected is fine, but the paragraph should acknowledge both.

docs/designs/074-supply-chain-attestation.md
```diff
@@ -66,11 +66,19 @@ This contract is published to users and enforced by us. Both the release-time ga
...
 
-**This design targets SLSA Build L2, not L3.** The distinction matters and is easy to overclaim. L3 requires the *build* to be isolated from user-defined steps, and GitHub's mechanism for that is moving the build itself into the reusable workflow. Here the builds stay in the callers — `docker buildx` in `publish.yml`, the Go cross-compile and `helm package` in `release.yml` — and `attest.yml` receives a digest and signs it. A caller that produced the wrong artifact would get a faithful signature over the wrong digest. Isolating the signer is worth having, but it is not the isolation L3 asks for.
+**This design targets SLSA Build L2 for every artifact we publish today — binaries and images alike.** Claims are per-artifact, not project-wide: nothing here may state a single Build Level for "the release" as a whole.
```
ndipebot: "binaries and images alike" (repeated at line 81 and in Consequences) drops the Helm chart and `installer` that the sentence this replaces covered, and that this same PR's `SECURITY.md` line 84 and `verifying-artifacts.md` lines 94-95 explicitly enumerate at Build L2. Since the ADR is the record the other two docs derive from, it should state their level too; otherwise, when the image-only L3 follow-on lands, the chart's disposition is undefined in the one document that decides it.

SECURITY.md
```diff
@@ -81,6 +81,10 @@ We credit reporters of confirmed vulnerabilities in the release notes of the fix
...
+
+  Pinning the workflow identity above is what makes that level checkable. Verification that omits `--certificate-identity` (or loosens it to a regexp) can still succeed while proving a weaker claim: it no longer distinguishes an attestation minted inside `attest.yml` from one produced elsewhere. The exact pin is enforced by `TestVerificationUsesExactIdentity` and `TestPublishedVerifyCommandsAreExact` in `test/releasepolicy`; the reusable-workflow boundary itself is gated by `TestAttestIsSoleSigner`, `TestAttestIsInvokedAsReusableWorkflow`, `TestAttestPredicateUsesOnlyTrustedContext`, and `TestAttestBuilderIdGuardRejectsAttestorAsBuilder`. No level is claimed here without a test behind it.
```
ndipebot: "The exact pin is enforced by `TestVerificationUsesExactIdentity` and `TestPublishedVerifyCommandsAreExact`" promises more than those two tests deliver: they only reject the `--certificate-identity-regexp` form (plus a bare `gh attestation verify` in workflows). I checked what actually catches drift: dropping `--certificate-identity` or pinning another org's identity in `docs/operations/verifying-artifacts.md` is caught, but by docspolicy's `TestVerificationPage*` tests, not these two; the same edits inside this file's own fenced commands are caught by nothing at all, wrong-org pin included.

Since the paragraph closes with "No level is claimed here without a test behind it", I'd either extend a docspolicy check to `SECURITY.md`'s fenced commands or soften the enforcement sentence to what the tests actually reject.

test/releasepolicy/attest_boundary_test.go
```diff
@@ -0,0 +1,201 @@
...
+		"repo":    "GITHUB_REPOSITORY",
+		"ref":     "GITHUB_REF",
+		"sha":     "GITHUB_SHA",
+		"server":  "GITHUB_SERVER_URL",
+		"run_id":  "GITHUB_RUN_ID",
+		"builder": "builder_id",
+	}
+	for arg, want := range wantByArg {
+		flag := "--arg " + arg + " "
+		idx := strings.Index(step.Run, flag)
```
ndipebot: `wantByArg` looks at only the first occurrence of each `--arg <name> ` (via `strings.Index`) and only that line, which misses later writers and false-fails on formatting. I verified both directions: appending a second `jq --arg repo "${SUBJECT_KIND}" '.buildDefinition.externalParameters.repository = $repo' provenance.json > p && mv p provenance.json` inside the same run block keeps everything green while making the signed predicate's `repository` caller-controlled; and reformatting the original trusted `jq` so `"${GITHUB_REPOSITORY}"` sits on a backslash-continuation line makes this test fail on a semantically identical command (`bash -n` clean).

Extracting and executing the step, then asserting on the resulting `provenance.json` fields, would fix both at once.

docs/designs/074-supply-chain-attestation.md
```diff
@@ -66,11 +66,19 @@ This contract is published to users and enforced by us. Both the release-time ga
...
+- A guard refuses to proceed if `builder_id` resolves to `attest.yml`, on the grounds that the attestor cannot be the builder. Naming the attestor as the builder would make the predicate false on its face, which is worse than claiming the wrong level.
+
+So the build steps in `release.yml` / `publish.yml` cannot forge or tamper with the provenance today. What is genuinely absent is **builder isolation**: the builds stay in the callers — `docker buildx` in `publish.yml` / `build-image.yml`, the Go cross-compile and `helm package` in `release.yml` — and `attest.yml` receives a digest and signs it. `runDetails.builder.id` truthfully names the caller. A compromised caller would produce a bad artifact that `attest.yml` would then honestly attest. That honesty is deliberate, and any change here must not trade a true predicate for a higher number.
```
ndipebot: Small factual slip: `publish.yml` contains no `docker buildx`; both `publish.yml` and `release.yml` call the reusable `build-image.yml`, which is where every buildx invocation lives. Someone grepping `publish.yml` for the image build to move it behind the L3 boundary finds nothing.

Related, "`runDetails.builder.id` truthfully names the caller" glosses that for images the id names the top-level orchestrator, which ran no build steps, rather than `build-image.yml` where the build actually runs (same imprecision at `verifying-artifacts.md` line 97, "the build runs in the caller").

test/releasepolicy/attest_boundary_test.go
```diff
@@ -0,0 +1,201 @@
...
+// Fulcio identity while TestAttestIsSoleSigner still saw cosign only inside
+// attest.yml — until the copy started signing too.
+func TestAttestIsInvokedAsReusableWorkflow(t *testing.T) {
+	assertAttestIsInvokedAsReusableWorkflow(t)
+}
+
+func assertAttestIsInvokedAsReusableWorkflow(t *testing.T) {
+	t.Helper()
+
+	callers := []string{wfRelease, wfPublish, wfAttestSmoke}
```
ndipebot: This caller list is hardcoded, so a workflow outside it that starts calling `attest.yml` is never form-checked, despite the doc-comment's "every release-path caller". I tried it: a new `publish-chart.yml` whose job uses the org-qualified, sha-pinned form passes this test, `TestActionsAreSHAPinned`, `TestAttestIsSoleSigner`, and every other test in the package (with `id-token: write` set, `TestIDTokenWriteOnlyOnSigningJobs` is satisfied too, and any real caller needs that permission anyway).

The package already has `workflowFiles` and `isAttestWorkflowCall`; sweeping all workflows for attest calls and requiring the `./` form on each, keeping this explicit list only as the must-call assertion, closes it.

SECURITY.md
```diff
@@ -81,6 +81,10 @@ We credit reporters of confirmed vulnerabilities in the release notes of the fix
 
   Use `--certificate-identity`, not `--certificate-identity-regexp`. An identity that names no workflow and no ref also accepts images built from branches, which are not releases and are labelled non-production when they are signed.
 
+  **Build Level (per artifact).** Every artifact this repository publishes today — the `manager` image (index provenance), the Helm chart, the `nvcrectl` binaries, and `installer` — carries SLSA Build Provenance at **Build L2**. The claim is per-artifact, not project-wide: nothing here states a single level for "the release". Provenance is unforgeable by the build process because it is minted inside the reusable `attest.yml` workflow (Fulcio names that workflow; predicate origin fields come from `GITHUB_*` context; a guard refuses to name `attest.yml` as the builder). What is absent is builder isolation — the build still runs in the caller — so we do not claim L3. A same-repo `uses: ./…` boundary additionally rests on branch protection over `.github/workflows/attest.yml`.
```
ndipebot: "Every artifact this repository publishes today" reads as exhaustive, but the `attest-binaries` matrix in `release.yml` also attests `THIRD_PARTY_NOTICES.md` and the four standalone `nvcrectl-*.cyclonedx.json` SBOM files, with `emit_provenance` defaulting to true. So five of the ten blob subjects carry Build L2 provenance without being named here (the list at `verifying-artifacts.md` lines 94-95 has the same gap), and someone enumerating provenance-carrying artifacts from this paragraph would treat a valid `THIRD_PARTY_NOTICES.md` bundle as unexpected.

Aside: `release.yml`'s own "Nine published assets" / "Nine legs" comments are stale too; the matrix has ten legs.

test/releasepolicy/attest_boundary_test.go
```diff
@@ -0,0 +1,201 @@
...
+// Complements isAttestWorkflowCall in workflow_policy_test.go: that helper
+// accepts any path whose base is attest.yml; this one insists on the ./ form
+// that keeps the call a same-repo workflow_call boundary.
+var localAttestUses = regexp.MustCompile(`^\./\.github/workflows/attest\.yml(@.+)?$`)
+
+// TestAttestIsInvokedAsReusableWorkflow pins the half of ADR-074 D2 that
+// TestAttestIsSoleSigner (workflow_policy_test.go) does not cover: every
+// release-path caller must reach attest.yml through `uses: ./…`, not by
+// inlining its jobs. Without this, a refactor could copy the attest steps into
+// release.yml, keep workflow_call on an unused attest.yml, and demote the
```
ndipebot: The scenario this comment gives can't happen as described: copying the attest steps into `release.yml` copies the `retry cosign sign` / `retry cosign attest` invocations, and `cosignSignCmds` (`workflow_policy_test.go:31`) flags those in any non-attest workflow immediately. I verified: the copy makes `TestAttestIsSoleSigner` fail the moment it lands, so there is no window "until the copy started signing too".

The gap this test genuinely closes, and I confirmed nothing else covers it, is a caller silently dropping its `attest.yml` call so nothing gets signed at all (converting `release.yml`'s five attest jobs to no-op run jobs fails only this test). I'd reword the comment to say that, so a future test consolidation doesn't delete the one check this file uniquely adds.

docs/operations/verifying-artifacts.md
```diff
@@ -89,6 +89,31 @@ built on it reports success for an artifact that was never released. The exact f
...
+
+The same-repo reusable-workflow form (`uses: ./.github/workflows/attest.yml`) isolates
+attestation from the caller's build steps, not from write access to the repository. The
+practical strength of that boundary rests on branch protection over `attest.yml`.
+
+**Pinning the identity above is the check that makes the level observable.** Drop the
+`--certificate-identity` flag (or replace it with a loose regexp) and verification can
+still go green while proving less than Build L2: you no longer know the attestation was
+minted inside the reusable workflow. The published commands keep the exact pin, and
+`TestVerificationUsesExactIdentity` / `TestPublishedVerifyCommandsAreExact` fail the
```
ndipebot: This gate-test enumeration is now duplicated verbatim between here and `SECURITY.md` line 86, and the L2 rationale prose is carried in three places (those two plus ADR-074), while nothing gates prose-cited Go test names against the test source: `test/docspolicy` checks fenced commands and nav, not test-name references. Renaming or consolidating any of the six cited tests leaves both pages asserting enforcement by tests that no longer exist.

I'd keep the enumeration in one place and link to it, or add a docspolicy check that each cited `Test*` identifier exists under `test/releasepolicy/`.

ndipebot (commented):
I went through this in depth, including running the new gates against mutated copies of the workflows to see what they actually catch. The direction is right: the boundary is real, the tests pass as written, and the honesty-first framing is the correct one. The inline comments are where a gate is weaker than the prose says it is, or where a claim doesn't match the workflows as they are.

One item with no diff line to hang it on: `.github/workflows/attest.yml`'s own header still states the framing this PR retires. Lines 25-31 keep the old "L3 requires the build to run inside the protected reusable workflow" framing, including `docker buildx` in `publish.yml` where no buildx lives anymore, and lines 7-9 still say attest.yml is not yet the sole signer because `publish.yml` signs the image inline, which contradicts the current workflows and `TestAttestIsSoleSigner`. Worth fixing here or in a fast follow so we don't carry two contradictory normative descriptions of the same boundary.

docs/designs/074-supply-chain-attestation.md
asivanadi0: Fixed in 0b15384. ADR-074 decision 4, SECURITY.md, and verifying-artifacts.md now distinguish branch protection (main / dev-image identity) from tag protection / repository rulesets over `v*` (release identity the SAN pins). Adding the actual `v*` ruleset on NVIDIA/cluster-readiness-engine needs maintainer admin rights, so that part is deferred to a follow-on with someone who can create it.

test/releasepolicy/attest_boundary_test.go
asivanadi0: Fixed in 2fe17c5. `TestAttestBuilderIdGuardRejectsAttestorAsBuilder` now extracts and executes the provenance step (same convention as `attest_guards_test.go`): accepts a legitimate caller, rejects `attest.yml` as builder, and covers the fail-closed malformed `workflow_ref` case. The redundant quoted-needle grep is gone.

test/releasepolicy/attest_boundary_test.go
asivanadi0: Fixed in 2fe17c5. Provenance env scanning now merges workflow / job / step env (same levels `runSteps` uses) and treats `needs.*.outputs.*` as untrusted for origin-shaped env vars. `SUBJECT_KIND` may still expand `needs.validate.outputs.subject_kind` (enum-validated, not an origin field); routing a needs output into `CALLER_WORKFLOW_REF` or another origin env fails the gate.

test/releasepolicy/attest_boundary_test.go
asivanadi0: Fixed in 2fe17c5. The loop now selects with `isAttestWorkflowCall` and asserts the `./` form on every match, so an org-qualified sha-pinned call can no longer hide behind sibling `./` calls. The previously dead `HasPrefix` branch is live.

docs/designs/074-supply-chain-attestation.md
asivanadi0: Fixed in 0b15384. Decision 4 and the Rationale bullet now claim only that provenance **origin** fields are minted from trusted context, and explicitly carve out `externalParameters.subjectKind` as the enum-validated `inputs.subject_kind` pass-through.

docs/designs/074-supply-chain-attestation.md
asivanadi0: Fixed in 0b15384. The deferred image L3 paragraph now records both costs: (1) honest L3 identity needs `job_workflow_ref`, not today's `github.workflow_ref` derivation; (2) signing inside `build-image.yml` splits the sole Fulcio SAN off `attest.yml` and breaks the published verify pin / `TestAttestIsSoleSigner`.

docs/designs/074-supply-chain-attestation.md
asivanadi0: Fixed in 0b15384. Decision 4, Consequences, SECURITY.md, and verifying-artifacts.md now name the Helm chart and `installer` (and the other provenance-carrying blob subjects) alongside binaries and images.

SECURITY.md
asivanadi0: Fixed in 2fe17c5 / 0b15384. Softened the SECURITY.md sentence to what those two tests actually reject (regexp form + bare `gh attestation verify` in workflows), pointed at `TestVerificationPagePinsAnExactIdentity` for the operations page, and added `TestSecurityMdPinsAnExactIdentity` so SECURITY.md bash fences also require the NVIDIA `attest.yml` identity (wrong-org / missing pin now fail).

test/releasepolicy/attest_boundary_test.go
asivanadi0: Fixed in 2fe17c5. `TestAttestPredicateUsesOnlyTrustedContext` extracts and executes the provenance step, then asserts `provenance.json` origin fields (repository, ref, commit, builder.id, invocationId). A second jq writer or a backslash-continued `--arg` line can no longer slip past a first-occurrence scan.

docs/designs/074-supply-chain-attestation.md
asivanadi0: Fixed in 0b15384. ADR-074 and the `attest.yml` header now say `docker buildx` lives in reusable `build-image.yml` (invoked by publish/release), and that `builder.id` names the top-level orchestrator for images rather than the workflow that ran the build steps. verifying-artifacts.md matches.

test/releasepolicy/attest_boundary_test.go
asivanadi0: Fixed in 2fe17c5. Form-checking now sweeps every workflow via `workflowFiles` + `isAttestWorkflowCall` and requires the `./` form on each; the hardcoded list remains only as the must-call assertion for release / publish / selftest.

SECURITY.md
asivanadi0: Fixed in 0b15384. SECURITY.md and verifying-artifacts.md now include `THIRD_PARTY_NOTICES.md` and the standalone `nvcrectl-*.cyclonedx.json` SBOM assets in the Build L2 list. `release.yml` comments updated Nine → Ten to match the matrix.

test/releasepolicy/attest_boundary_test.go
asivanadi0: Fixed in 2fe17c5. Reworded the `TestAttestIsInvokedAsReusableWorkflow` doc comment: the unique gap this test closes is a caller silently dropping its `attest.yml` call (nothing signed at all). Copying cosign into a non-attest workflow is already caught by `TestAttestIsSoleSigner`.

docs/operations/verifying-artifacts.md
asivanadi0: Fixed in 2fe17c5 / 0b15384. verifying-artifacts.md now links to SECURITY.md for the gate-test enumeration instead of duplicating it, and `TestCitedGateTestsExist` in docspolicy fails if either page cites a `Test*` that is not defined under `test/releasepolicy` or `test/docspolicy`.

asivanadi0:
## ndipebot review follow-up (adr-074-d2-build-level-gate)

Pushed `2fe17c5` (tests) and `0b15384` (docs/workflows). `make lint` = 0; `go test ./test/releasepolicy/...` and `make test` green locally. Fresh `/ok to test` needed for CI on these commits.

### Addressed (all 14 ndipebot threads + attest.yml header)
| Thread | Topic | Fix |
|---|---|---|
| 3982022817 | Tag rulesets vs branch protection | Named tag rulesets for release identity in ADR / SECURITY / verifying-artifacts |
| 3982022830 | Execute builder-id guard | Extract-and-run + fail-closed coverage |
| 3982022842 | inputs. laundering via needs/env levels | Scan workflow+job+step; treat needs.*.outputs as untrusted for origin env |
| 3982022852 | Assert `./` on every attest call | `isAttestWorkflowCall` then require `./` |
| 3982022860 | SUBJECT_KIND overclaim | Scoped claim to origin fields |
| 3982022867 | Deferred L3 path costs | Documented `job_workflow_ref` + sole-signer split |
| 3982022875 | Chart/installer in L2 wording | Included in ADR / SECURITY / verifying-artifacts |
| 3982022883 | Exact-pin enforcement overclaim | Softened prose; added `TestSecurityMdPinsAnExactIdentity` |
| 3982022890 | wantByArg first-occurrence hole | Execute step; assert provenance.json |
| 3982022899 | buildx / builder.id factual slip | build-image.yml + top-level orchestrator wording; attest.yml header fixed |
| 3982022911 | Hardcoded caller list | Sweep all workflows for form; list = must-call only |
| 3982022923 | Notices/SBOMs + Nine comments | Listed; Nine→Ten in release.yml |
| 3982022932 | Misleading test comment | Reworded to drop-call gap |
| 3982022939 | Duplicated gate-test list | Link to SECURITY.md + `TestCitedGateTestsExist` |
| (summary) | attest.yml header contradiction | Header now matches sole-signer + L2 framing |

CodeRabbit 3961385868 was already resolved (per-workflow attest call in 60f9b93).

### Deferred
- **Create an actual repository ruleset over `v*` tags** (3982022817): docs now name the control; creating it on `NVIDIA/cluster-readiness-engine` needs maintainer admin rights.

test/releasepolicy/attest_boundary_test.go
ndipebot: The rework catches everything I threw at it through the `needs.*.outputs.*` and job-level env routes. One lookalike still gets through: adding `SNEAKY: ${{ inputs['subject_name'] }}` (bracket index form, valid expression syntax) to the provenance step env and folding it into the jq call as `--arg repo "${GITHUB_REPOSITORY}${SNEAKY:-}"` passes this test and both policy packages in full. The env scan misses it because the value contains neither the literal `inputs.` nor a `needs.*.outputs.*` reference, and the executed-step assertion misses it because `SNEAKY` is unset in the harness, so `${SNEAKY:-}` expands empty and the exact-value checks pass, while a real run would fold caller data into `externalParameters.repository`.

Two ways to close it: also reject `inputs[` (and `needs[`) in the env scan, or have the harness set every non-exempt env var the script references to a sentinel and assert the sentinel never reaches an origin field.

test/releasepolicy/attest_boundary_test.go
ndipebot: Executing the step and asserting the origin fields kills both directions I demonstrated; I re-verified the second-writer and the backslash-reformat cases at the new head. One field is still unasserted: `resolvedDependencies[0].uri`, the SLSA source-origin URI built as `git+` + server + `/` + repo + `@` + ref. Appending

```
jq --arg u "git+https://evil.example/attacker/repo@refs/heads/main" '.buildDefinition.resolvedDependencies[0].uri = $u' provenance.json > p && mv p provenance.json
```

to the step passes this test and both packages, because the parse struct only reads `digest.gitCommit` out of `resolvedDependencies`. With `uri` asserted, `buildType` is the only unasserted field left, and that one is a hardcoded constant.

docs/designs/074-supply-chain-attestation.md
ndipebot: The split between the two identities is right, and deferring the actual `v*` ruleset to someone with admin makes sense. One nit in the new sentence: the parenthetical pairing `release.yml` / `publish.yml` with `v*` tags overstates `publish.yml`, which cannot run on a `v*` tag. Its `on:` is `push: branches: [main]` plus `workflow_dispatch`, its header says it no longer runs on tags, and the load-bearing ref guard on the tag job (`github.ref == 'refs/heads/main'`) exists precisely so a dispatch on a tag ref skips every job and cannot mint `attest.yml@refs/tags/<TAG>`. I'd drop `publish.yml` from that parenthetical; its guard is itself evidence for the sentence's claim.

docs/designs/074-supply-chain-attestation.md
ndipebot: The origin-fields scoping and the `subjectKind` carve-out check out here and in the Rationale bullet. One mirror picked the old shape back up though: the rewritten `attest.yml` header now says "Provenance is unforgeable by the build process today (minted here from trusted context)" with no origin-fields qualifier, in the one file where `inputs.subject_kind` demonstrably reaches the predicate. Worth adding the same qualifier there so the header matches the ADR it summarizes.

SECURITY.md
ndipebot: Both lists are complete now, and ten matches the matrix. One number got left behind in `release.yml`: the comment now says "Ten legs, and the four binary legs sign twice" but still concludes "~13 near-simultaneous calls". Ten legs plus the four second signatures is 14; the 13 was coherent with the old nine. Should read ~14.

docs/operations/verifying-artifacts.md
ndipebot: `TestCitedGateTestsExist` closes the staleness gap for the two pages it scans, in both directions. The fix added one new ungated citation though: ADR-074's sole-signer-split cost now cites `TestAttestIsSoleSigner`, and `gateTestCitePages` covers only `SECURITY.md` and this page, so renaming the func while updating just the scanned pages leaves the ADR citing a test that no longer exists. It also makes `SECURITY.md` and the ADR two pages citing the same name, which the parenthetical here says the single home prevents. Either add the ADR to `gateTestCitePages` or unbacktick the name in the ADR prose.

docs/designs/074-supply-chain-attestation.md
asivanadi0: Fixed in cf4b87c. Dropped `publish.yml` from the release-identity parenthetical; it is now `release.yml` on `v*` tags only. `publish.yml`'s main-only `on:` + tag-job ref guard remain the evidence that it cannot mint `attest.yml@refs/tags/<TAG>`.

test/releasepolicy/attest_boundary_test.go
asivanadi0: Fixed in cf4b87c. Provenance env scanning (and the run-block check) now also reject `inputs[` / `needs[` bracket-index forms, so `${{ inputs['subject_name'] }}`-style laundering fails the gate the same way as `inputs.` / `needs.*.outputs.*`.

docs/designs/074-supply-chain-attestation.md
asivanadi0: Fixed in cf4b87c. The `attest.yml` header now matches the ADR: provenance **origin** fields are minted from trusted context, with an explicit `externalParameters.subjectKind` carve-out for the enum-validated `inputs.subject_kind` pass-through.

test/releasepolicy/attest_boundary_test.go
asivanadi0: Fixed in cf4b87c. `TestAttestPredicateUsesOnlyTrustedContext` now asserts `resolvedDependencies[0].uri` equals `git+` + server + `/` + repo + `@` + ref, so a second jq writer that only swaps the URI fails.

SECURITY.md
asivanadi0: Fixed in cf4b87c. `release.yml` comment now says ~14 near-simultaneous calls (ten legs + four second signatures).

docs/operations/verifying-artifacts.md
asivanadi0: Fixed in cf4b87c. Added `docs/designs/074-supply-chain-attestation.md` to `gateTestCitePages`, so ADR-074's backtick citation of `TestAttestIsSoleSigner` is gated the same way as SECURITY.md / verifying-artifacts.md.

test/docspolicy/verification_page_test.go
```diff
@@ -189,3 +189,55 @@ func navHasPath(node any, want string) bool {
...
+	if len(cmds) == 0 {
+		t.Fatal("SECURITY.md publishes no cosign verify commands in bash fences")
+	}
+	for _, c := range cmds {
+		flat := strings.Join(strings.Fields(c), " ")
+		if !strings.Contains(flat, "--certificate-identity") {
+			t.Errorf("SECURITY.md command pins no --certificate-identity: %s", flat)
+			continue
+		}
+		if !strings.Contains(flat, wantIdentity) && !strings.Contains(flat, "${ID}") {
```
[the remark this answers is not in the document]
asivanadi0: Fixed in cf4b87c. `TestSecurityMdPinsAnExactIdentity` no longer accepts `${ID}` as a stand-in: every SECURITY.md `cosign verify*` fence must contain the literal NVIDIA `attest.yml@refs/tags/` identity. (SECURITY.md already inlines it; verifying-artifacts.md keeps its own `${ID}` + page-wide check.)

asivanadi0:
Addressed ndipebot's 2nd-pass (and the open CodeRabbit `${ID}` note) in **cf4b87c**.

### Still-open items from the 2nd pass
1. **Bracket-index env laundering** — provenance env/run scan now rejects `inputs[` / `needs[` as well as `inputs.` / `needs.*.outputs.*`.
2. **`resolvedDependencies[0].uri`** — executed provenance step asserts the SLSA source URI, not only `digest.gitCommit`.

### Nits
- ADR-074 release-identity parenthetical: `release.yml` on `v*` only (dropped `publish.yml`).
- `attest.yml` header matches the ADR origin-fields + `subjectKind` carve-out.
- `release.yml` comment: ~14 near-simultaneous calls.
- ADR-074 added to `gateTestCitePages`.
- `TestSecurityMdPinsAnExactIdentity` requires the literal NVIDIA identity (no `${ID}` escape hatch).

`make lint` 0 issues; `go test ./test/releasepolicy/... ./test/docspolicy/...` and `make test` green.

Fresh `/ok to test` would help re-run CI on the new head (`cf4b87c`).

All 7 unresolved threads (6 ndipebot + 1 CodeRabbit) replied + resolved; nothing deferred.

asivanadi0:
Pushed two cheap nits from the 3rd-pass hold note (docs-only; #340 is up as https://github.com/NVIDIA/cluster-readiness-engine/pull/341):

1. ADR-074 topology: `publish.yml (tag)` → `publish.yml (main)` so it matches the prose / `on: push: branches: [main]`.
2. `verifying-artifacts.md`: SECURITY.md link now uses the same absolute GitHub URL style as the See also section (keeps `#supply-chain`).

asivanadi0:
Head moved to `de122e6` (docs nits). Fresh trustee `/ok to test` needed on the current SHA for NVIDIA runners.

asivanadi0:
Rebased onto latest `main` (includes #341 / `84e994c`). Tip is now `161c07f` (was `de122e6`). Conflicts resolved in `SECURITY.md` (kept #341 old-tag path + ADR-074 Build Level) and `localAttestUses` rename vs #341's string const. Fresh `/ok to test` may be needed on the new SHA.

asivanadi0:
Rebased onto latest `main` (includes #334/#348/#355/#356). Tip is now `160594c` (was `161c07f`). Clean rebase (no conflicts). `make lint` 0 issues; `go test ./test/releasepolicy/...` and `make test` green locally. Fresh trustee `/ok to test` needed on the new SHA for NVIDIA runners.

docs/designs/074-supply-chain-attestation.md
```diff
@@ -103,7 +116,7 @@ A **daily re-verification job** re-runs the same suite against the latest releas
 ### Workflow topology
 
 ```
-publish.yml (tag)  ──┐
+publish.yml (main)  ──┐
```
ndipebot, line 119: `3e303fc4` has the fact right, `publish.yml` does run on `main`, but the extra character pushed the corner one column right, so the top branch no longer meets the junction: `┐` lands at column 22 while `├` and `┘` are both at column 21. Dropping one space before `──┐` lines it back up.

Cosmetic, not blocking.

ndipebot (approved):
/lgtm
/approve

asivanadi0:
Rebased onto `main` after #363 (also includes #360–#362 and #364). New tip: `6ab6ee9` (`6ab6ee9c21277124e9664e28960eeab77d371a3d`).

asivanadi0:
Rebased onto main after tonight’s merges. New tip: `ce5f777a32a52cd9ada8e520a2fc82acd680f428`.

Issue #213, opened 2026-08-24
[Bug]: Suspended TrainJob is treated as   Running, causing spurious node failures on Kueue-managed clusters   Labels:  bug ,  needs-triage  · Component: Controller / Reconcilers


martin-piercy:
### Prerequisites

- [x] I searched existing issues
- [x] I can reproduce this issue

### Code of Conduct

- [x] I agree to follow the project's Code of Conduct

### Bug Description

Note, I cant reproduce as I dont have a cluster, this came out of analysis with Claude but I was asked to add these requests in. 

On a cluster where Kueue manages TrainJobs, Kueue's webhook sets spec.suspend: true  and holds the workload until quota is admitted. During that window the TrainJob has no pods and no status conditions.

   TrainJobAdapter.GetStatus()  only recognises two conditions and falls  through to  Running  for everything else:

    // pkg/workload/trainjob.go:146-172
    for _, cond := range trainJob.Status.Conditions {
        if cond.Type == trainerv1alpha1.TrainJobFailed  && cond.Status ==
  metav1.ConditionTrue { ... }
    }
    for _, cond := range trainJob.Status.Conditions {
        if cond.Type == trainerv1alpha1.TrainJobComplete && cond.Status ==
  metav1.ConditionTrue { ... }
    }
    return &WorkloadStatus{Phase: WorkloadRunning}, nil     // ← line 172: a  suspended TrainJob lands here

  So CRE believes a queued workload is running, starts its stall clock, sees no progress, and after  startupStallTimeoutSeconds  (1200s in the training catalog entries) or  DefaultTimeoutPerJob  ( 1h ,  pkg/catalog/loader.go )   marks the Job failed — attributing  WorkloadFailed  to very node in the group.

  This is the only issue in this cluster that produces a wrong certification  result rather than an operational annoyance. It fails hardware that is  perfectly healthy, which is a direct attack on the product's core claim.

  Note  WorkloadPhase  ( pkg/workload/adapter.go:19-23 ) has only  Running , Succeeded ,  Failed  — there is no  Pending , so this cannot be expressed today. A grep for  suspend  across  pkg/  and  api/  returns nothing, despite the vendored TrainJob CRD carrying  spec.suspend  ( hack/crds/trainer.kubeflow.org_trainjobs.yaml:5562 ) and  spec.managedBy  (:447, whose CEL rule explicitly permits  kueue.x-k8s.io/multikueue ).

  ### Proposed Fix

[...]

PR #309, opened 2026-09-06
feat(controller): surface scheduling stalls as WorkloadSchedulingBlocked

### LINKED ISSUE #310: [Bug]: Job stays InProgress(WorkloadRunning) when workload pods are unschedulable
## Prerequisites

## Code of Conduct

## Bug Description

A Job's workload pods sat Pending for **7 hours** (`FailedScheduling: Insufficient nvidia.com/gpu` ×114) while the Job, Workflow, and Certification all reported a healthy-looking `InProgress(WorkloadRunning)`. `timeoutPerJob` burned against hardware that never started a single process. The operator's only discovery path was `kubectl describe pod` on a guessed pod name — nothing in `nvcrectl certification report`, `--wait` streaming, or the status conditions said "this is not running, it cannot even be placed".

### What happened

A workload pod was pinned (via node affinity) to a node whose GPUs were held by other tenants. The pod carried `PodScheduled=False/Unschedulable`, but `TrainJobAdapter.GetStatus` maps non-suspended TrainJobs with no terminal conditions to `WorkloadRunning` — a pod the scheduler explicitly rejected is indistinguishable from one that's actually running. The Job controller's default branch then stamps `WorkloadStartTime`, which starts the stall and timeout clocks against hardware that is doing nothing.

### What was expected

The Job should surface a distinct, non-terminal condition (e.g. `WorkloadSchedulingBlocked`) relaying the scheduler's own `FailedScheduling` diagnosis, with blocked time excluded from `timeoutPerJob` and stall detection — the same clock-neutral treatment as #213's pending-time contract.

## Component

- Controller / Reconcilers (Job controller + `pkg/workload` adapter)

## Steps to Reproduce

1. Create a Job whose workload TrainJob references a TrainingRuntime with a `nodeSelector` matching no node
2. Observe the pod go `Pending` with `PodScheduled=False/Unschedulable`
3. `kubectl get job.nvcre.nvidia.com <name> -o jsonpath='{.status.conditions[0]}'` — reports `InProgress=True(WorkloadRunning)` instead of a scheduling-blocked signal

## Environment

- NVCRE version: v0.1.0
- Kubernetes version: v1.36.1
- GPU architecture: H200 (3-node on-prem, no cloud providerID → onprem)
- Cloud provider / platform: onprem
- Deployment method: `nvcrectl setup init` (Helm chart)

## Logs/Output

```
# Pod condition
PodScheduled: False (Unschedulable)
Message: 0/3 nodes are available: 1 Insufficient nvidia.com/gpu, 2 node(s) didn't match Pod's node affinity/selector

# Job status during the 7h stall
InProgress=True(WorkloadRunning) — no scheduling-blocked signal
workloadStartTime: 2026-09-06T01:47:23Z (clock running against a pod that never started)
```

A proposed fix (ADR-075 with clock-pause semantics) is in PR #309.

ntheanh201:
## Summary

Adds a first-class `WorkloadSchedulingBlocked` condition on the Job when a workload's pods cannot be placed — replacing the current behavior where unschedulable pods silently read as `WorkloadRunning` while `timeoutPerJob` burns against hardware that never ran.

Motivated by a real incident on an on-prem 3-node HGX H200 cluster: a per-node loopback job was hard-pinned to one node whose GPUs were held by tenants; the pod sat `Pending` for 7 hours (`FailedScheduling` ×114) with no operator-visible signal.

## What this adds

- **New Job-tier condition reason**: `WorkloadSchedulingBlocked` — set when the workload's pods carry `PodScheduled=False/Unschedulable` past a grace window (default 5 min, tunable via `spec.schedulingStallGraceSeconds`)
- **Clock-pause semantics**: `WorkloadStartTime` is cleared when blocked persists (pausing `timeoutPerJob` and the stall clock) and restored on the next genuinely-running observation — extends the issue #213 pending-time contract to the unschedulable case
- **Message relays the scheduler's own diagnosis** (newest `FailedScheduling` event text) via a new Event `involvedObject.name` field index
- **`schedulingBlockedSince`** status field persists the blocked-episode start so controller restarts don't reset the grace clock
- **Detection runs on every reconcile** of a non-terminal workload, including after the clock started (amended after live testing exposed a guard race — see ADR-075)

Closes #310

## Design

Full design in [ADR-075](docs/designs/075-scheduling-stall-visibility.md) (Status: Proposed).

## Testing

- 10 unit tests covering grace window, mixed pods, recovery, event relay, no-condition pods
- Integration golden case `job-scheduling-blocked` pinning the full contract
- Live-validated on a 3-node on-prem HGX H200 cluster: pinned job produced `WorkloadSchedulingBlocked` with the relayed scheduler diagnosis within the grace window, `workloadStartTime` correctly paused

## Type of Change
- [x] ✨ New feature

## Component(s) Affected
- [x] API / CRDs

## Testing
- [x] Tests added (unit + integration golden)
- [x] `make manifests generate` run (CRDs regenerated for new fields)
- [x] Documentation updated (concept + api-reference + troubleshooting + ADR)

## Checklist
- [x] Ready for review

### DIFF
--- a/docs/designs/075-scheduling-stall-visibility.md
+++ b/docs/designs/075-scheduling-stall-visibility.md
@@ -216,3 +216,36 @@ behavior.
 - [ADR-068: group-nodes compressed ConfigMap](068-group-nodes-compressed-configmap.md) — ConfigMap capture precedent
 - Issue #213 — pending time exclusion from timeout/stall clocks
 - `pkg/workload/trainjob.go:187` — the fall-through this ADR augments
+
+
+## Amendment: clock-pause semantics (post-live-testing)
+
+Live testing on a 3-node on-prem cluster exposed a race in the original
+guard: the `WorkloadStartTime != nil` early-return prevented detection in
+the common case where the clock started on reconcile 1 (workload created,
+clock started) and the scheduler rejected the pod on reconcile 2. The
+detector's guard skipped detection precisely when the clock was already
+ticking against a pod that had never run.
+
+**Amended decision (point 4 revised):** detection runs on every reconcile
+of a non-terminal workload, including after the clock started. When blocked
+persists past the grace window, the caller **clears `WorkloadStartTime`**
+(pauses the clock) in the same status write that sets the
+`WorkloadSchedulingBlocked` condition; the first-observe logic restores the
+timestamp on the next genuinely-running observation. This keeps timeout
+accounting honest in both directions: a workload that never ran is never
+charged, and a workload that ran and then became unschedulable is not
+charged for the blocked period either.
+
+The `schedulingBlockedSince` marker semantics are unchanged: set on first
+blocked observation, cleared when any pod schedules; the caller never
+writes it.
+
+**Unit test updated:** the "WorkloadStartTime set — never blocked" case
+becomes "WorkloadStartTime set — blocked still fires (clock-pause
+amendment)".
+
+**Deployment note:** the 3-node on-prem cluster validated the amendment
+live: the pinned-job test produced the `WorkloadSchedulingBlocked`
+condition with the scheduler's relayed message within the grace window
+after this amendment.
--- a/docs/designs/075-scheduling-stall-visibility.md
+++ b/docs/designs/075-scheduling-stall-visibility.md
@@ -1,0 +1,218 @@
+# ADR-075: Surface Runtime Scheduling Stalls as a First-Class Job Condition
+
+> **Status:** Proposed
+
+## Context
+
+When a workload's pods cannot schedule, NVCRE gives the operator no signal that
+distinguishes "waiting for GPUs" from "running". The condition hierarchy
+reports `InProgress=True(WorkloadRunning)` regardless.
+
+A real incident on an on-prem 3-node HGX H200 cluster: the `nccl-loopback`
+category's per-node job was hard-pinned to one node (nodeAffinity) and
+requested all 8 GPUs. Two tenant pods held 3 of those GPUs. The workload pod
+sat `Pending` for **7 hours** — `FailedScheduling: 1 Insufficient nvidia.com/gpu,
+2 node(s) didn't match Pod's node affinity/selector` fired 114 times — while:
+
+- the Job reported `InProgress=True(WorkloadRunning)` the whole time,
+- the Workflow and Certification showed a healthy-looking run,
+- `timeoutPerJob` and the startup-stall detector counted against a workload
+  that never started a single process.
+
+The operator's only discovery path was `kubectl describe pod` on a pod whose
+name had to be guessed from the Job name. Nothing in `nvcrectl certification
+report`, `--wait` streaming, or the status conditions said "this is not running,
+it cannot even be placed".
+
+### Why the current phases hide this
+
+`TrainJobAdapter.GetStatus` ([trainjob.go:142-188](../../pkg/workload/trainjob.go#L142))
+maps the TrainJob to exactly four phases:
+
+- `Complete` condition → `WorkloadSucceeded`
+- `Failed` condition → `WorkloadFailed`
+- `spec.suspend=true` → `WorkloadPending` (deliberate: Kueue-queued time must
+  not burn `timeoutPerJob` or the stall budget — issue #213)
+- **everything else falls through to `WorkloadRunning`** (line 187)
+
+A non-suspended TrainJob whose pods are unschedulable lands in the last bucket.
+The Job controller's `default` branch
+([job_controller.go:505-540](../../pkg/controller/job_controller.go#L505)) then
+stamps `WorkloadRunning` and `WorkloadStartTime`, which starts the stall and
+timeout clocks against hardware that is doing nothing.
+
+The genuinely-pending path (suspended TrainJob) is handled correctly — but
+there is no phase for "unsuspended, admitted, and still not schedulable".
+
+Pre-flight blockers (cordoned nodes, GPU-architecture mismatch, insufficient
+GPU capacity) *are* surfaced before any Job is created, via exclusion records
+and events like `InsufficientGPUCapacity`
+([workflow_controller.go:442-505](../../pkg/controller/workflow_controller.go#L442)).
+The gap is exclusively the runtime case: pods exist, cannot be placed, and the
+reason lives in pod events that nothing reads.
+
+## Decision
+
+Introduce a **scheduling-stall detector on the Job tier** that inspects the
+workload's pods directly and reports the finding as a first-class, non-terminal
+condition — without changing phase semantics, timeout accounting, or retry
+behavior.
+
+1. **New Job-tier reason: `WorkloadSchedulingBlocked`.** A new condition/reason
+   in the existing `InProgress` family
+   ([helpers.go:103-114](../../pkg/controller/helpers.go#L103)). It follows the
+   tier-prefixed reason convention (`ReasonWorkload*`).
+
+2. **Detection lives in the Job controller, not the adapter.** The adapter
+   interface (`Adapter.GetStatus`) stays phase-only — adapters are typed to
+   their workload resource and have no pod-listing capability, and adding one
+   would push Kubernetes-pod knowledge into every adapter (torch, mpi, exec,
+   custom) for information the Job controller already has. The Job controller
+   lists pods by the existing `nvcre.nvidia.com/job: <name>` label (same
+   mechanism as `NodeDiscoverer.DiscoverNodesForJob`
+   [nodes.go:36-51](../../pkg/nodemonitor/nodes.go#L36) and
+   `captureTimeoutLog` [workflow_controller.go:2510](../../pkg/controller/workflow_controller.go#L2510)).
+
+3. **Classification — `WorkloadSchedulingBlocked` fires when ALL of:**
+   - the workload is unsuspended and in the running path (phase would be
+     `WorkloadRunning`), i.e. pods exist but the workload has never started;
+   - at least one workload pod has `PodScheduled=False`
+     (`status.conditions[type=PodScheduled]`, status `False`, reason
+     `Unschedulable`);
+   - the blocked state has persisted for a grace window (default **5 minutes**,
+     overridable via `Job.spec.schedulingStallGraceSeconds`, following the
+     `StartupStallTimeoutSeconds` pattern) — so a slow-but-healthy bin-packing
+     event does not flap the condition.
+
+4. **Non-terminal, clock-neutral.** Like `WorkloadPending`, a scheduling-blocked
+   workload is not running: `WorkloadStartTime` stays unset, so queued/stalled
+   time never counts against `timeoutPerJob` or the stall budget. The Job stays
+   `InProgress`. No retry, no restart, no node-failure attribution — NVCRE does
+   not modify the cluster (ADR-061); it reports.
+
+5. **Message carries the scheduler's reason.** The condition message embeds the
+   most recent `FailedScheduling` event message for the blocked pod (e.g.
+   `0/3 nodes are available: 1 Insufficient nvidia.com/gpu, 2 node(s) didn't
+   match Pod's node affinity/selector`). The scheduler already wrote the
+   diagnosis; NVCRE's job is to relay it, not re-derive it.
+
+6. **Recovery is automatic and silent.** If the pod schedules on a later
+   reconcile, the normal `WorkloadRunning` path proceeds and the condition
+   simply stops being re-asserted (`setExclusiveCondition` semantics mean the
+   InProgress condition updates to `WorkloadRunning` on the next observation).
+   No terminal state is ever set by this detector.
+
+7. **Visibility paths.** The blocked reason lives on the Job CR and propagates
+   upward through the existing aggregation: the Workflow's InProgress condition
+   reflects the Job's reason, and `nvcrectl certification report` renders
+   Job/Workflow condition chains. The `--wait` stream today prints only
+   category-status transitions
+   ([certification.go:1303-1312](../../pkg/certification/certification.go#L1303)),
+   so a stalled category appears as an enduring `InProgress` line; surfacing
+   the reason in the watch stream itself is a small waiter enhancement,
+   included in scope below.
+
+## Implementation
+
+- **`pkg/controller/helpers.go`** — add `ReasonWorkloadSchedulingBlocked =
+  "WorkloadSchedulingBlocked"` to the Job-tier reason block.
+- **`pkg/controller/job_controller.go`**
+  - New `checkSchedulingBlocked(ctx, job) (blocked bool, message string)`:
+    lists pods by `labelJobKey`, filters to pods with `PodScheduled=False`
+    and no `nodeName`, extracts the newest `FailedScheduling` event message
+    per pod, returns the aggregate message.
+  - Called from the `default` (running) branch **before** the
+    `WorkloadStartTime`/stall logic: if blocked past the grace window, set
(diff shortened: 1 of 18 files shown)


pkg/controller/job_controller.go
```diff
@@ -673,6 +705,166 @@ func (r *JobReconciler) checkStallTimeout(ctx context.Context, job *nvcrev1alpha
...
+		}
+		// Only pods the scheduler has explicitly rejected count as blocked:
+		// PodScheduled=False. A pod with no PodScheduled condition at all has
+		// not been processed yet — counting it would fire on pods that are
+		// merely new.
+		schedulingRejected := false
+		for _, cond := range pod.Status.Conditions {
+			if cond.Type == corev1.PodScheduled &&
+				cond.Status == corev1.ConditionFalse {
+				schedulingRejected = true
```
asivanadi0, line 769: Draft feedback: Decision §3 and the API/docs all require PodScheduled=False with reason Unschedulable, but this only checks Type/Status. Pods waiting on a scheduling gate (Reason=SchedulingGated — common with ResourceClaims / DRA on GPU nodes) also carry PodScheduled=False and no nodeName, so after schedulingStallGraceSeconds they would surface as WorkloadSchedulingBlocked even though the scheduler has not rejected them.

On an HGX cert fleet that uses claim templates, that looks like a real false positive: the Job would flip to WorkloadSchedulingBlocked and pause the timeout clock while the gate is still doing intentional work.

Suggest gating on cond.Reason == corev1.PodReasonUnschedulable (and adding a SchedulingGated fixture that must stay non-blocked past grace). SchedulerError is a separate question — worth an explicit call either way.

docs/designs/075-scheduling-stall-visibility.md
```diff
@@ -0,0 +1,251 @@
...
+    Job status (`SchedulingBlockedSince *metav1.Time`, cleared on recovery) —
+    persisted like `WorkloadStartTime` so controller restarts do not reset it.
+- **`api/v1alpha1/job_types.go`** — add
+  `SchedulingStallGraceSeconds *int32` (optional, default 300) and
+  `SchedulingBlockedSince *metav1.Time` (status).
+- **`pkg/certification/certification.go`** — the `--wait` watcher additionally
+  prints Job/Workflow condition-reason transitions for InProgress workloads
+  (e.g. `[watch] communication/nccl-all-reduce: InProgress — WorkloadSchedulingBlocked (2m)`)
+  so a stalled category is distinguishable from a progressing one without
+  leaving the stream.
```
asivanadi0, line 138: Draft feedback: Implementation still lists a pkg/certification/certification.go --wait enhancement as in-scope (reason transitions like InProgress — WorkloadSchedulingBlocked (2m)), and Consequences claim operators see the reason in the --wait stream. This PR does not touch certification.go.

Either land that waiter change here, or demote it to an explicit follow-up and soften the Consequences / Decision §7 wording so the ADR matches what actually ships. As written, a reader of the ADR will expect the watch stream to distinguish a scheduling stall from a progressing category, and the current code will not.

pkg/controller/job_controller.go
```diff
@@ -504,6 +509,33 @@ func (r *JobReconciler) updateStatusFromWorkload(ctx context.Context, job *nvcre
...
+					j.Status.WorkloadStartTime = nil
+					return true
+				}); err != nil {
+				return ctrl.Result{}, fmt.Errorf("failed to update Job status: %w", err)
+			}
+			return ctrl.Result{RequeueAfter: r.getWorkloadRequeueInterval()}, nil
+		}
+		if job.Status.SchedulingBlockedSince != nil {
+			// Within the grace window: requeue, keep the clock as-is.
+			return ctrl.Result{RequeueAfter: r.getWorkloadRequeueInterval()}, nil
```
asivanadi0, line 536: Draft feedback / clarifying question on the clock-pause contract during grace.

When WorkloadStartTime is already set and pods then become unschedulable, this branch requeues for the whole grace window with the clock as-is. timeoutPerJob therefore keeps burning until grace elapses and the blocked path clears WorkloadStartTime. The API comment and troubleshooting docs say blocked time does not count against timeoutPerJob — that is only true after the condition surfaces, not for the grace interval itself.

Is burning up to schedulingStallGraceSeconds of timeout budget on a newly-unschedulable-but-previously-running workload intentional (noise filter), or should the pause start when SchedulingBlockedSince is first recorded? If intentional, a one-line note on the field docs / ADR would save the next reader from treating the stronger claim as literal.

asivanadi0 (commented):
Draft feedback — treating this as reviewable, not an approve/request-changes gate.

The clock-pause amendment (detect after WorkloadStartTime is set, clear it once blocked past grace) matches the incident you describe, and the unit matrix around grace / mixed pods / event relay is in good shape.

Same ADR numbering collision as on #288: both PRs land an ADR-075. Please renumber one before either merges.

Two concrete gaps below (Reason filter vs SchedulingGated, and the --wait work the ADR still claims is in scope).

kaynetu (commented):
This addresses an important part of #310: the Job now has a distinct, non-terminal scheduling-blocked reason, the condition relays the scheduler's diagnosis, and the persisted blocked-episode timestamp prevents controller restarts from resetting the grace window. Checking after WorkloadStartTime has already been set also covers the creation/scheduler race described in the amendment. Keeping detection in the Job controller avoids expanding the adapter contract.

I reviewed head 08817b5 and ran the supplied scheduling unit tests and job-scheduling-blocked integration case; both passed. Additional isolated controller probes reproduced the four behavioral issues below. I did not run the full suite.

There are still five gaps to address before considering #310 fully resolved, each anchored once inline:

1. An existing timeout clock continues during grace, so the Workflow can terminate an unschedulable Job before the blocked condition appears.
2. Clearing and recreating WorkloadStartTime resets the timeout budget instead of preserving previously consumed runtime.
3. Training-stall detection still includes the blocked interval after recovery and can immediately fail/restart the Job.
4. PodScheduled=False also matches SchedulingGated; the documented Unschedulable reason is not checked.
5. Workflow/Certification aggregation, certification report, and --wait still do not expose the blocked reason as the ADR claims.

The integration fixture starts without WorkloadStartTime, has no Workflow enforcing timeoutPerJob, and does not exercise recovery, so its passing result does not establish the complete clock-pause contract. Please add coverage for an already-started Job entering and leaving a block, timeout enforcement during grace, training-stall recovery, scheduling gates, and the operator-facing visibility paths.

pkg/controller/job_controller.go
kaynetu, line 536: [P1] Exclude blocked time from timeout accounting during grace

This returns with an existing WorkloadStartTime unchanged. WorkflowReconciler.isJobTimedOut independently compares that timestamp with timeoutPerJob, so an unschedulable Job with less than the grace period remaining can be marked JobTimedOut and have its workload deleted before WorkloadSchedulingBlocked appears. I reproduced this with a start time two minutes ago, a blocked observation one minute ago, the default five-minute grace, and a 90-second timeout. Please exclude blocked time from the first observation while retaining grace for condition visibility, and cover this with a Workflow timeout test.

pkg/controller/job_controller.go
```diff
@@ -504,6 +509,33 @@ func (r *JobReconciler) updateStatusFromWorkload(ctx context.Context, job *nvcre
...
+		// cleared when a pod schedules); the caller never writes it.
+		blocked, blockMsg := r.checkSchedulingBlocked(ctx, job)
+		if blocked {
+			log.V(1).Info("Workload pods are unschedulable", "kind", ref.Kind, "name", ref.Name)
+			if err := r.setJobInProgress(ctx, job, ReasonWorkloadSchedulingBlocked, blockMsg,
+				func(j *nvcrev1alpha1.Job) bool {
+					if j.Status.WorkloadStartTime == nil {
+						return false
+					}
+					j.Status.WorkloadStartTime = nil
```
kaynetu, line 527: [P2] Preserve consumed runtime when pausing the timeout

Clearing WorkloadStartTime here causes the recovery path to stamp the current time as a new start. That discards runtime consumed before the block: a workload that ran for 49 minutes and then spent 10 minutes blocked recovers with a completely fresh timeoutPerJob budget. Repeated blocks can keep extending execution. I reproduced the reset through updateStatusFromWorkload. Please preserve elapsed active runtime (or equivalent accumulated pause accounting) and test that recovery receives only the previously remaining budget.

pkg/controller/job_controller.go
```diff
@@ -673,6 +705,166 @@ func (r *JobReconciler) checkStallTimeout(ctx context.Context, job *nvcrev1alpha
...
+					if j.Status.SchedulingBlockedSince == nil {
+						return false
+					}
+					j.Status.SchedulingBlockedSince = nil
+					return true
+				}); err != nil {
+				logf.FromContext(ctx).Error(err, "Failed to clear schedulingBlockedSince")
+			}
+		}
+		return false, ""
```
kaynetu, line 799: [P2] Exclude the blocked interval from training-stall detection on recovery

Returning to the normal running path here immediately calls checkStallTimeout. For a workload that already emitted training steps, that function measures time since GoodputMeasurement.LastStepTimestamp; the cleared WorkloadStartTime only helps the startup-stall path. A long scheduling block therefore causes an immediate WorkloadStalled failure or checkpoint restart on recovery, before a new step can arrive. I reproduced the failure with an existing measurement and a step timestamp preceding the block. Please account for the paused interval in the training-stall budget too, with a recovery regression test.

pkg/controller/job_controller.go
kaynetu, line 769: [P2] Require the documented Unschedulable reason

This checks only the condition type and status. A pod with PodScheduled=False and Reason=SchedulingGated also has no nodeName, so intentional gating is classified as scheduler rejection after grace and clears the timeout clock. I reproduced this with a SchedulingGated pod. Please require cond.Reason == corev1.PodReasonUnschedulable for the documented contract, or explicitly define any additional reasons, and add a gate case that remains non-blocked past grace. This confirms the earlier review observation on this filter.

docs/designs/075-scheduling-stall-visibility.md
```diff
@@ -0,0 +1,251 @@
...
+
+6. **Recovery is automatic and silent.** If the pod schedules on a later
+   reconcile, the normal `WorkloadRunning` path proceeds and the condition
+   simply stops being re-asserted (`setExclusiveCondition` semantics mean the
+   InProgress condition updates to `WorkloadRunning` on the next observation).
+   No terminal state is ever set by this detector.
+
+7. **Visibility paths.** The blocked reason lives on the Job CR and propagates
+   upward through the existing aggregation: the Workflow's InProgress condition
+   reflects the Job's reason, and `nvcrectl certification report` renders
```
kaynetu, line 107: [P2] Implement the claimed visibility beyond the Job condition

The existing aggregation does not relay this reason: Workflow reconciliation still sets ReasonJobRunning with a running-group count, certification report converts category InProgress to Running and only populates the category failure reason for Failed, and --wait prints category-status transitions only. None of those paths changes in this PR. The diagnosis is consequently visible on the Job CR but remains absent from the operator surfaces identified in #310. Please propagate/render the blocked reason and cover the report/watch behavior, or explicitly narrow the shipped scope and track these unresolved parts separately. This gap extends beyond the earlier comment about the missing waiter enhancement.

asivanadi0 (commented):
## Judging-lane re-review (tip `ca1ab84`)

**State:** DRAFT · `mergeable=CONFLICTING` / `mergeStateStatus=DIRTY` vs `main` · author `ntheanh201`.
**Delta since prior reviews (on `08817b5`):** tip is effectively the prior head plus `fix(rbac): add get/list/watch on events for nvcre-manager`. No substantive change to the scheduling-stall detector, clock-pause contract, stall recovery, or operator visibility paths. Force-push / history rewrite also leaves the branch conflicting with `main` (large unrelated churn in the PR file list).

Holding **COMMENT** only (not APPROVE): still draft + conflicting, and the P1 clock-grace timeout hole is still open.

### Checklist vs prior threads

#### asivanadi0 (2026-09-09)

| Item | Status | Evidence at `ca1ab84` |
|---|---|---|
| Require `PodScheduled=False` **reason `Unschedulable`** (exclude `SchedulingGated`) | **OPEN** | `checkSchedulingBlocked` still accepts any `PodScheduled=False` (`job_controller.go` ~767–769). No `cond.Reason == corev1.PodReasonUnschedulable` gate; no SchedulingGated negative fixture in `scheduling_stall_test.go`. |
| `--wait` / certification scope vs ADR claims | **OPEN** | ADR Decision §7 + Implementation still list `pkg/certification/certification.go` waiter enhancement; Consequences still claim operators see the reason in `--wait` / report. PR file list still has **no** `certification.go` (or report/workflow aggregation) change. |
| Clock-pause during grace (clarifying Q) | **OPEN** (answered as bug by kaynetu P1) | Grace branch still: `// Within the grace window: requeue, keep the clock as-is.` (`job_controller.go` ~534–536). Docs/ADR still imply blocked time does not burn `timeoutPerJob` without distinguishing grace. |

#### kaynetu (2026-09-18)

| Item | Status | Evidence at `ca1ab84` |
|---|---|---|
| **[P1]** Exclude blocked time from timeout during grace | **OPEN** | Same grace early-return leaves `WorkloadStartTime` set; Workflow `isJobTimedOut` can still `JobTimedOut` / delete before `WorkloadSchedulingBlocked` surfaces. No Workflow timeout coverage added. |
| **[P2]** Preserve consumed runtime when pausing | **OPEN** | Past-grace path still nils `WorkloadStartTime` (~527); recovery first-observe stamps `metav1.Now()` (~541–544) → full timeout budget reset. No accumulated-pause / remaining-budget accounting. |
| **[P2]** Exclude blocked interval from training-stall on recovery | **OPEN** | Recovery clears `schedulingBlockedSince` and returns into the normal path that calls `checkStallTimeout`; training-stall still keys off `GoodputMeasurement.LastStepTimestamp` with no pause credit. No recovery regression test. |
| **[P2]** Require documented Unschedulable reason | **OPEN** | Same filter gap as asivanadi0 row above; still no SchedulingGated stay-non-blocked case. |
| **[P2]** Implement claimed visibility beyond Job condition | **OPEN** | Job condition + event relay only. No Workflow reason propagation, certification report rendering, or `--wait` reason transitions in this tip. ADR scope not narrowed / follow-ups not carved out. |

### What *did* land (adjacent, not a thread closer)

- **PARTIAL / adjacent:** manager RBAC now includes `get`/`list`/`watch` on `events` / `events.k8s.io` (needed for FailedScheduling relay). Good; does not close any of the items above.
- Core Job-tier condition + `schedulingBlockedSince` persistence + unit/integration golden for the happy path remain as previously reviewed — useful foundation, still incomplete vs #310 / ADR-075.

### Top residuals (block merge readiness)

1. **P1 — pause timeout from first blocked observation** (not only after grace / condition surface); cover with a Workflow `timeoutPerJob` test that cannot terminate during grace.
2. **Unschedulable reason filter + SchedulingGated fixture** (shared asivanadi0 / kaynetu P2).
3. **Preserve pre-block consumed runtime** on pause/resume (kaynetu P2).
4. **Training-stall pause credit on recovery** (kaynetu P2).
5. **Either ship Workflow/report/`--wait` visibility or demote ADR Decision §7 / Implementation / Consequences** and track the rest for #310 follow-ups (kaynetu P2 + prior ADR-scope note).
6. **Resolve conflicts with `main`** and clean the force-push churn before asking for another pass.
7. **ADR-075 number collision** with #288 (`075-mcp-server.md`) still outstanding — renumber one before either merges.

Happy to re-review once the P1 grace/timeout contract and Unschedulable filter land (even while draft); approve gate stays with Anpoo after draft/conflicts/open P1s clear.

PR #316, opened 2026-09-08
fix(report): include captured failure logs

### LINKED ISSUE #140: [Bug]: FailureLog captured by setJobFailed never surfaced in report output
### Prerequisites

### Code of Conduct

### Bug Description
`r.captureFailureLog(ctx, job)` in `pkg/controller/job_controller.go:784` writes log content to `job.Status.FailureLog`. However, the report generation code in `pkg/report/` (grep: `FailureLog` returns no hits) does not read or include this field. Operators reviewing the Certification report after a failure have no access to the captured log excerpt, making it harder to diagnose what went wrong on the failed nodes.

**Location:** `pkg/controller/job_controller.go:784` (write path), `pkg/report/` (missing read path)

**Suggested fix:** Include `job.Status.FailureLog` in the per-category failure section of the generated report, or surface it via `kubectl get job -o yaml` documentation.

#

### Component

Controller / Reconcilers

### Steps to Reproduce
1. Run a Certification where a Job fails.
2. Retrieve the Certification report via `ncrectl certification report`.
3. Observe that the failure section contains no log excerpt despite `kubectl get job -o yaml` showing a non-empty `status.failureLog` field.

#### Environment
- CRE version: all
- Kubernetes version: all
- GPU architecture: all
- Deployment method: all

kaynetu:
## Summary

- Include the existing `Job.status.failureLog` in each failed group's machine-readable report so JSON consumers receive the captured pod, node, exit code, termination reason, and bounded log tail.
- Render the same diagnostics in the human report, using a shared terminal-cell width calculation for wrapping and padding (including wide characters and combining marks), and escaping C0, DEL, C1, and Unicode bidirectional controls from workload output.
- Omit unknown exit codes from both human and JSON output, preserving explanatory reasons for timeout, `PodNotFound`, and `NoTerminatedContainer` captures.
- Identify the single captured pod and limit human tails to the last 20 rendered lines from at most 4 KiB of input. Truncation is explicit and points to the full captured excerpt in JSON or `Job.status.failureLog`; JSON retains the existing bounded tail.
- Document the report fields and their current Job-scoped lifetime. Durable excerpts after Job teardown remain outside this fix.

Closes #140

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] API / CRDs
- [ ] Catalog / Workloads
- [x] CLI (nvcrectl)
- [ ] Helm / Deployment
- [x] Documentation / CI
- [ ] Other: ____________

## Testing
- [x] Tests pass locally
- [ ] Manual testing completed

Added report coverage for human and JSON output, control-character sanitization, line wrapping, and timeout captures without a false exit code. Handwritten output fixtures pin wide-character and combining-mark wrapping and padding independently of the width helper. Tests cover all 32 C1 control code points and the rendered CSI/OSC escape strings.

Additional fixtures pin JSON exit-code omission and human output for timeout, `PodNotFound`, and `NoTerminatedContainer`. Budget tests cover exact limits, overflow, long-line wrapping, UTF-8 byte boundaries, truncation notices, and preservation of the full JSON tail. The existing deleted-Job report golden continues to verify omission of `failureLog`.

Bidi coverage checks all 12 Unicode `Bidi_Control` characters, wrapping of visible escapes, preservation of ordinary RTL letters and joiners, and unchanged JSON tail data.

The existing `golang.org/x/text v0.41.0` dependency is now imported directly for Unicode width properties; no dependency version changed relative to main.

Verified with:

- `make manifests generate`
- `make lint-fix`
- `make build`
- `make test`

## Risk

Low and additive. No controller or CRD behavior changes. Failure-log tails remain capped at 32 KB by the existing capture path and are emitted only while the referenced Job exists.

## Checklist
- [x] Self-review completed
- [x] `make manifests generate` run (if `*_types.go` was modified) — no generated changes
- [x] Golden files updated (if integration test output changed) — report golden updated intentionally; integration goldens unchanged
- [x] Ready for review


asivanadi0 (commented):

Surfacing Job.status.failureLog into FailedGroupReport (and the human card) is the right layer for #140, and the Exit-vs-Reason split plus C0/C1 sanitization / East-Asian width handling look deliberate. A few concrete gaps before this is ready to land:

### 1. JSON still advertises exitCode: 0 for non-terminated captures

printFailureLog correctly suppresses Exit: 0 when only Reason is set (good for the timeout / no-terminated-container path). FailureLogReport.ExitCode is a non-pointer int32 with no omitempty, so the machine-readable report still emits "exitCode": 0 for the same objects — including the controller's real explanatory captures (NoTerminatedContainer / PodNotFound), which never set an exit code.

That is a footgun for JSON consumers that treat exitCode as authoritative. Even if the CRD keeps the zero-value field, the report schema could use *int32 / omitempty (or omit the key when there was no terminated container) so human and JSON agree on "no exit code". Worth confirming the intended contract and documenting it either way under failedGroups[].failureLog.

### 2. Human report dumps the full 32 KiB tail

Capture is capped at failureLogMaxBytes (32 KiB). This PR prints the entire Tail into the boxed human report with wrapping and no further elision. For a chatty failing workload that is hundreds of wrapped lines inside a certification card. JSON keeping the full bounded tail makes sense; human output probably wants a tighter line/byte budget plus an elision marker, with a pointer that the full excerpt remains in JSON / Job.status.failureLog.

### 3. One representative pod vs multi-pod group failure

captureFailureLog already picks a single pod (OOMKilled-preferring). A group reason like "BackoffLimitExceeded — 2 of 2 pods failed" can therefore sit next to a Failure Log for only one of those pods, with no cue that sibling failures were not aggregated. interpret-results.md says "the failed workload pod's" (singular), which helps, but the card itself does not. A short clarifying line in the Failure Log header (or in the how-to) that this is the representative captured pod would avoid operators chasing missing sibling logs.

### 4. Test coverage vs real capture shapes

TestPrintFailureLog invents Reason: "Timeout" with pod/node filled in. The controller paths that actually produce ExitCode == 0 today write NoTerminatedContainer / PodNotFound (often with empty podName/nodeName and an explanatory tail). The Job-deleted omission claimed in the docs also has no buildFailedGroups golden. Pinning those two shapes (explanatory zero-exit capture + Job gone → no failureLog) would lock the schema/docs better than the fictional Timeout fixture alone.

Happy to re-review once the exitCode JSON contract and human-tail budget are settled.

kaynetu:
@asivanadi0 Addressed in [3d657c9](https://github.com/NVIDIA/cluster-readiness-engine/commit/3d657c9fc0fdf85c5e87f058e5034eb3c0003595).

1. JSON now omits unknown exit codes via `omitempty`. Nonzero exit codes remain. [Schema](https://github.com/NVIDIA/cluster-readiness-engine/blob/3d657c9fc0fdf85c5e87f058e5034eb3c0003595/pkg/report/report.go#L178).
2. Human tails keep the last 20 rendered lines from at most 4 KiB of input. Truncation points to the full JSON/Job excerpt. JSON remains unchanged. [Implementation](https://github.com/NVIDIA/cluster-readiness-engine/blob/3d657c9fc0fdf85c5e87f058e5034eb3c0003595/pkg/report/report.go#L1155), [limit tests](https://github.com/NVIDIA/cluster-readiness-engine/blob/3d657c9fc0fdf85c5e87f058e5034eb3c0003595/pkg/report/report_test.go#L490).
3. The header now says “Failure Log (one captured pod).” [Code](https://github.com/NVIDIA/cluster-readiness-engine/blob/3d657c9fc0fdf85c5e87f058e5034eb3c0003595/pkg/report/report.go#L1135).
4. Added human/JSON fixtures for `PodNotFound`, `NoTerminatedContainer`, and timeout. [Fixtures](https://github.com/NVIDIA/cluster-readiness-engine/tree/3d657c9fc0fdf85c5e87f058e5034eb3c0003595/pkg/report/testdata/failure-log-capture-shapes).

Two clarifications: [`captureTimeoutLog`](https://github.com/NVIDIA/cluster-readiness-engine/blob/3d657c9fc0fdf85c5e87f058e5034eb3c0003595/pkg/controller/workflow_controller.go#L2555) does produce the timeout shape. The existing [deleted-Job golden](https://github.com/NVIDIA/cluster-readiness-engine/blob/3d657c9fc0fdf85c5e87f058e5034eb3c0003595/pkg/report/testdata/build-failed-groups/job-deleted-configmap-fallback/expected.txt) already covers omission of `failureLog`.

Lint, build, and the full unit/integration suite passed.

asivanadi0 (commented):
Follow-up COMMENT on `00685b1` (after [3d657c9](https://github.com/NVIDIA/cluster-readiness-engine/commit/3d657c9fc0fdf85c5e87f058e5034eb3c0003595) + bidi follow-up) — still not an approve / request-changes gate.

Thanks @kaynetu — re-checked the four items from https://github.com/NVIDIA/cluster-readiness-engine/pull/316#pullrequestreview-5156562890 against current HEAD.

### Prior findings

1. **JSON `exitCode: 0` for non-terminated captures — FIXED**
   `FailureLogReport.ExitCode` is now `json:"exitCode,omitempty"` (`pkg/report/report.go`). Human/JSON fixtures for `NoTerminatedContainer`, `PodNotFound`, and timeout omit the key while keeping nonzero exits (e.g. execution-failed-precedence `exitCode: 137`). Docs under `failedGroups[].failureLog` match.

2. **Human 32 KiB dump — FIXED**
   `failureLogExcerpt` caps human output at 4 KiB input / 20 rendered lines, marks truncation, and points at JSON / `Job.status.failureLog`. JSON tail unchanged. Covered by `TestFailureLogExcerptLimits`.

3. **Multi-pod opacity — FIXED**
   Card header is `Failure Log (one captured pod):` when a pod was selected; how-to states the log is one selected (OOMKilled-preferring) pod, not an aggregate. execution-failed-precedence golden pins the `BackoffLimitExceeded — 2 of 2 pods failed` + single-pod log pairing.

4. **Capture-shape / omission goldens — FIXED**
   `failure-log-capture-shapes` pins `NoTerminatedContainer`, `PodNotFound`, and timeout (human + remarshaled JSON). Agreed the existing `job-deleted-configmap-fallback` golden already asserts `failureLog` absence when the Job is gone; timeout is a real controller shape via `captureTimeoutLog`.

### CodeRabbit since that review

The post-`3d657c9` bidi/`Cf` note is addressed in `00685b1` via `unicode.Bidi_Control` escaping (with ZWJ/ZWNJ preserved on purpose). Earlier width / C1 notes were already marked addressed. Nothing from that thread still looks open on HEAD.

### Residual (non-blocking)

Only optional schema polish: `podName` / `nodeName` still serialize as `""` on explanatory captures while `exitCode` uses `omitempty`. Harmless; empty strings are honest. Not a land blocker.

From this pass, the four prior gaps look closed. Leaving as COMMENT (not a formal approve). Happy to look again if anything else lands on top of `00685b1`.

kaynetu:
@asivanadi0  Thanks for re-checking. I’ll keep the empty pod/node fields as-is. No further changes planned from this review.

pkg/report/report.go
```diff
@@ -1100,9 +1121,131 @@ func printFailedGroups(w io.Writer, groups []FailedGroupReport) {
...
+	tail = strings.TrimSuffix(strings.ReplaceAll(tail, "\r\n", "\n"), "\n")
+	truncated := len(tail) > failureLogHumanMaxBytes
+	if truncated {
+		start := len(tail) - failureLogHumanMaxBytes
+		for start < len(tail) && !utf8.RuneStart(tail[start]) {
+			start++
+		}
+		tail = tail[start:]
+	}
+	lines := wrappedBoxTextLines("           ", tail)
```
ndipebot: These 11 spaces have to stay equal to the ones used when printing the tail above, or the wrap width and the real indent drift apart and the box misaligns. Worth a shared const.

pkg/report/report_test.go
```diff
@@ -404,6 +405,151 @@ func TestPrintReportFailed(t *testing.T) {
...
+		if err != nil {
+			return err
+		}
+		tc.Actual = buf.String() + string(encoded) + "\n"
+		return nil
+	})
+}
+
+// TestFailureLogExcerptLimits checks boundary budgets and preservation of JSON.
+func TestFailureLogExcerptLimits(t *testing.T) {
```
ndipebot, line 496: `pkg/report` doesn't use table tests anywhere. The other three tests you added here already go through `testutil.TestCaseParser`, and these five budget cases map cleanly onto testdata directories.

ndipebot (changes requested):
The width and sanitization work here is careful, but the wrapping has a bug that shows up on every real capture.

`printWrappedBoxText` prepends the full prefix to each wrapped line, not just the first, so a long pod name renders as two pods under a header saying there is one:

```
| Failure Log (one captured pod):                        |
|   Pod: aws-gb300-nccl-communicat-e795f-job-workload-lau|
|   Pod: ncher-0-0-x7k2m                                 |
```

The budget for `Pod: ` is 48 cells, and real NCCL workload pods are 63 chars: `test/uat/testdata/aws/gb300/nccl/expected_pods.yaml` carries a 58-char `generateName` and Kubernetes appends 5 more. So this is not an edge case, it hits every NCCL failure capture. `Node:` goes through the same path on a 47-cell budget, which AWS internal DNS names only just fit. The new fixtures miss it because they use short pod names, so a 60+ char one would be worth adding.

Fix suggested inline, plus two small things.

One I would leave alone: the unsanitized JSON tail is the right call, since `WriteJSON` writes to `CRE_RESULTS_FILE` rather than stdout, so those bytes never reach a terminal through nvcrectl.

pkg/report/report.go
```diff
@@ -1100,9 +1121,131 @@ func printFailedGroups(w io.Writer, groups []FailedGroupReport) {
...
+	}
+	return lines, truncated
+}
+
+// printWrappedBoxText prints text inside the report card, preserving explicit
+// newlines and wrapping every resulting line to the available terminal width.
+func printWrappedBoxText(w io.Writer, prefix, value string) {
+	for _, line := range wrappedBoxTextLines(prefix, value) {
+		printBoxLine(w, prefix+line)
+	}
```
ndipebot: `prefix` is prepended to every wrapped line, so a pod name longer than the 48-cell budget renders as two `Pod:` entries under a header that says "one captured pod". Real NCCL workload pods are 63 chars (see `test/uat/testdata/aws/gb300/nccl/expected_pods.yaml`), so this fires on every capture.

```suggestion
	cont := pad(displayWidth(prefix))
	for i, line := range wrappedBoxTextLines(prefix, value) {
		if i == 0 {
			printBoxLine(w, prefix+line)
			continue
		}
		printBoxLine(w, cont+line)
	}
```

Checked this keeps the box at 66 columns, with the package suite green and no golden changes.

asivanadi0 (commented):
Follow-up COMMENT on `00685b1` — still not an approve.

Checked in after @kaynetu's note on empty `podName`/`nodeName` (keeping as-is is fine; that polish was optional) and @ndipebot's CHANGES_REQUESTED on wrapping.

Agree with the wrap finding: `printWrappedBoxText` still does `prefix+line` on every continuation (`pkg/report/report.go`), so a 63-char NCCL pod name (48-cell `Pod: ` budget at `boxWidth=66`) renders as a second `Pod:` line under the "one captured pod" header. Same path for `Node:`. Worth fixing before merge; the inline continuation-indent suggestion looks right. The shared indent const and moving `TestFailureLogExcerptLimits` onto `TestCaseParser`/testdata are fair follow-ons once the wrap is corrected.

No new commits since our last pass, so nothing else from our side.

pkg/report/report.go
```diff
@@ -1100,9 +1121,131 @@ func printFailedGroups(w io.Writer, groups []FailedGroupReport) {
...
+// Newlines are handled by the caller before sanitizing each individual line.
+func sanitizeTerminalText(s string) string {
+	var b strings.Builder
+	for _, r := range s {
+		switch {
+		case r == '\t':
+			b.WriteString("    ")
+		case r < ' ' || (r >= 0x7f && r <= 0x9f):
+			_, _ = fmt.Fprintf(&b, "\\x%02x", r)
+		case unicode.Is(unicode.Bidi_Control, r):
```
ndipebot, line 1248: The escape set here is narrower than the zero-width set in `runeDisplayWidth`, and the gap disables wrapping entirely.

`runeDisplayWidth` returns 0 for all of `unicode.Cf` (report.go:1883), but only the `Bidi_Control` subset is escaped here. For anything in the gap, `cells + 0 > availableCells` is never true, so the wrap loop at report.go:1219 can never split the line:

```
U+00AD SOFT HYPHEN       width=0  lines=1   maxLineBytes=4096  escaped=false
U+200B ZERO WIDTH SPACE  width=0  lines=1   maxLineBytes=4095  escaped=false
U+FEFF ZERO WIDTH NBSP   width=0  lines=1   maxLineBytes=4095  escaped=false
U+2060 WORD JOINER       width=0  lines=1   maxLineBytes=4095  escaped=false
U+202E RLO               width=0  lines=20  maxLineBytes=51    escaped=true
```

With `Tail = "ERR " + strings.Repeat("\u00ad", 3000)` the card emits one physical line of 4166 bytes whose computed `displayWidth` is 66, so `printBoxLine` pads it as though it were exactly full width. The 20-line cap gives no protection because nothing ever wraps. U+00AD is the worst of the set in practice: it is `Cf` by category, but glibc wcwidth returns 1, so xterm and the Linux console paint roughly 3000 visible hyphens inside a 66-column card.

Two ways out, and the call is yours. Escaping all of `unicode.Cf` closes it, but collides with the joiner assertion in `TestFailureLogBidiControls`, which deliberately preserves ZWJ and ZWNJ in Arabic text. Otherwise leave the escape policy alone and give the wrap loop a byte-based backstop so it always makes forward progress regardless of cell width.

pkg/report/report.go
```diff
@@ -1100,9 +1121,131 @@ func printFailedGroups(w io.Writer, groups []FailedGroupReport) {
...
 
+const (
+	failureLogHumanMaxBytes = 4 * 1024
+	failureLogHumanMaxLines = 20
+)
+
+// failureLogExcerpt keeps the end of the captured tail for human output.
+// The byte cap applies before sanitizing; the line cap applies after wrapping.
+func failureLogExcerpt(tail string) ([]string, bool) {
+	tail = strings.TrimSuffix(strings.ReplaceAll(tail, "\r\n", "\n"), "\n")
```
ndipebot: Only CRLF is normalized, so a lone CR falls through to the `r < ' '` case in `sanitizeTerminalText` and renders as `\x0d`:

```
|             Epoch 0:  10%|#         | 100/1000\x0dEpoch 0:  50%|
|             |#####     | 500/1000\x0dEpoch 0: 100%|##########| |
|             1000/1000                                          |
|             RuntimeError: CUDA out of memory                   |
```

tqdm and Lightning both write progress with bare CR, and the `training/nemotron5-8b` entries run them. A terminal would show three overwritten lines; here each CR costs 4 characters of escape noise against a budget of 20 rendered lines.

The traceback does survive, since the excerpt keeps the end, so this is about readability and how much of the budget progress output eats rather than losing the error.

```suggestion
	tail = strings.TrimSuffix(strings.NewReplacer("\r\n", "\n", "\r", "\n").Replace(tail), "\n")
```

which gives:

```
|             Epoch 0:  10%|#         | 100/1000                 |
|             Epoch 0:  50%|#####     | 500/1000                 |
|             Epoch 0: 100%|##########| 1000/1000                |
|             RuntimeError: CUDA out of memory                   |
```

Package suite stays green with no golden changes.

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/spf13/cobra v1.10.2
 	github.com/spf13/pflag v1.0.10
 	github.com/stretchr/testify v1.12.1
+	golang.org/x/text v0.41.0
```
ndipebot, line 13: Promoting `golang.org/x/text` to a direct dependency leaves `THIRD_PARTY_NOTICES.md` incomplete. That file opens with:

> This file lists every third-party Go module that is a direct dependency of the `nvcrectl` binary (including the internal packages it imports), along with the verbatim text of each module's license.

`go list -deps ./cmd/nvcrectl` now reports 8 `golang.org/x/text` packages linked into the binary, and `grep x/text THIRD_PARTY_NOTICES.md` returns nothing. The file is more than documentation: `release.yml` copies it to release assets (:391), includes it in the required digest map (:428), attests it as a Sigstore blob subject (:481) and uploads it with its `.sigstore.json` (:871). As it stands the next release publishes a signed license disclosure that omits a linked BSD-3-Clause dependency.

Fair caveat: nothing regenerates or validates this file and it has already drifted on its own, with 6 of the 15 listed modules carrying stale versions. So this is one more omission in an artifact that needs its own fix rather than something this PR broke. Adding the entry keeps it from getting worse.

kaynetu:
@ndipebot Addressed all six in [9564a10](https://github.com/NVIDIA/cluster-readiness-engine/commit/9564a1070840325a673c7ddb0442d9f777a41b80).

- Shared the tail indentation constant. [Code](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/pkg/report/report.go#L1172-L1191).
- Moved budget cases to TestCaseParser fixtures. [Tests](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/pkg/report/report_test.go#L495).
- Print Pod/Node labels only on the first line. [Code](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/pkg/report/report.go#L1201-L1209), [63-character pod fixture](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/pkg/report/testdata/print-wrapped-box-text/long-pod/expected.txt).
- Added a 256-byte per-line backstop while preserving joiners. [Code](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/pkg/report/report.go#L1224-L1233), [soft-hyphen fixture](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/pkg/report/testdata/failure-log-excerpt-limits/soft-hyphen/expected.json).
- Normalize bare CR and CRLF in human excerpts. JSON stays unchanged. [Code](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/pkg/report/report.go#L1182), [fixture](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/pkg/report/testdata/failure-log-excerpt-limits/bare-cr/input.yaml).
- Added the x/text BSD-3-Clause notice and verbatim license. [Notice](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/THIRD_PARTY_NOTICES.md#L34).

Local lint, build, and the full unit/integration suite passed.

pkg/report/report.go
kaynetu: Fixed in 9564a10. Printing and wrapping now share `failureLogTailIndent`. [Code](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/pkg/report/report.go#L1163-L1191).

pkg/report/report_test.go
kaynetu: Fixed in 9564a10. The budget cases now use `TestCaseParser` with testdata fixtures. [Tests](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/pkg/report/report_test.go#L495-L543).

pkg/report/report.go
kaynetu: Fixed in 9564a10. Continuation lines use spaces instead of repeating Pod/Node labels. [Code](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/pkg/report/report.go#L1201-L1209), [63-character pod fixture](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/pkg/report/testdata/print-wrapped-box-text/long-pod/expected.txt). Report tests pass.

pkg/report/report.go
kaynetu: Fixed in 9564a10 with a 256-byte per-line backstop, preserving ZWJ/ZWNJ. [Code](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/pkg/report/report.go#L1224-L1233). The [soft-hyphen](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/pkg/report/testdata/failure-log-excerpt-limits/soft-hyphen/expected.json) and [zero-width](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/pkg/report/testdata/failure-log-excerpt-limits/zero-width/expected.json) fixtures pass. JSON remains unchanged.

pkg/report/report.go
kaynetu: Fixed in 9564a10. Bare CR and CRLF become newlines in human excerpts; JSON preserves the original text. [Code](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/pkg/report/report.go#L1182), [regression fixture](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/pkg/report/testdata/failure-log-excerpt-limits/bare-cr/input.yaml). Report tests pass.

go.mod
kaynetu: Fixed in 9564a10. Added the x/text v0.41.0 index entry and verbatim BSD-3-Clause license from the module. [Notice](https://github.com/NVIDIA/cluster-readiness-engine/blob/9564a1070840325a673c7ddb0442d9f777a41b80/THIRD_PARTY_NOTICES.md#L34-L69). Unrelated version drift is unchanged.

ndipebot (approved):

PR #330, opened 2026-09-09
feat(helm): add controller disruption protection

### LINKED ISSUE #141: [Feature]: Helm chart does not include a PodDisruptionBudget for the controller — node maintenance can evict it mid-run
### Prerequisites

### Code of Conduct

### Bug Description
The CRE Helm chart (`helm/cluster-readiness-engine/templates/`) has no `PodDisruptionBudget` for the controller Deployment. During rolling node maintenance (e.g. `kubectl drain`), the controller pod can be evicted while active Certifications are in progress. Controller restarts mid-run cause temporary reconcile gaps, requeue storms, and — in the case of the `GoodputMeasurementReconciler` — in-process state loss (see issue about state persistence on leader election).

**Location:** `helm/cluster-readiness-engine/templates/`

**Suggested fix:** Add a `PodDisruptionBudget` with `minAvailable: 1` (for single-replica deployments) and expose it as a Helm value (`pdb.enabled`, `pdb.minAvailable`).

#

### Component

Helm / Deployment

### Steps to Reproduce
1. Deploy CRE and start a long-running Certification.
2. Drain the node running the CRE controller pod: `kubectl drain <node> --ignore-daemonsets --delete-emptydir-data`.
3. Observe the controller is evicted and Certifications experience a reconcile gap.

#### Environment
- CRE version: all
- Kubernetes version: all
- Deployment method: Helm

kaynetu:
## Summary

The chart cannot currently retain a ready controller during voluntary disruptions such as node drains. This change adds an opt-in `policy/v1` PodDisruptionBudget with configurable `minAvailable`, plus preferred hostname anti-affinity for controller replicas by default.

The PDB remains disabled by default because `minAvailable: 1` with the default singleton controller would block voluntary drains. The deployment guide documents the supported two-replica configuration, required anti-affinity and infrastructure-node placement, the leader handover gap, singleton maintenance behavior, affinity replacement semantics, and the absence of PDB flags from `nvcrectl setup init`.

## Related Issue

Closes #141

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] API / CRDs
- [ ] Catalog / Workloads
- [ ] CLI (nvcrectl)
- [x] Helm / Deployment
- [x] Documentation / CI
- [ ] Other: ____________

## Testing
- [x] Tests pass locally (`make ci` and `helm lint helm/cluster-readiness-engine`)
- [ ] Manual testing completed

The Helm tests cover PDB enablement, defaults, integer and percentage values, explicit zero, resource naming and selectors, the default preferred anti-affinity, and full replacement by a custom `manager.affinity` value.

## Risk Assessment

- Existing installations retain current drain behavior because the PDB is opt-in.
- Preferred anti-affinity improves replica spreading without blocking installation on a single-node cluster.
- Any non-empty `manager.affinity` value replaces the chart default; this contract is documented and tested.
- Enabling `minAvailable: 1` with one controller replica blocks voluntary eviction, as documented.

## Checklist
- [x] Self-review completed
- [x] Golden files updated (if integration test output changed; no golden changes were required)
- [x] Ready for review


asivanadi0 (commented):

Opt-in controller PDB + preferred hostname anti-affinity is the right shape for #141, and the risk posture matches the chart defaults (`manager.replicas: 1`). Walked the templates, values, deployment guide, and `test/helm/pdb_test.go` against HEAD `37c411577db8d7305146de9d25f55f3db7ced2fc`.

### What looks solid

- **`pdb.enabled: false` by default** — correct. `minAvailable: 1` with the shipped singleton would block voluntary drains; values comment and deployment guide both say so explicitly.
- **Selector / identity alignment** — PDB `matchLabels` (`nvcre.selectorLabels` + `control-plane: manager`) matches the Deployment selector; PDB name/namespace reuse `nvcre.resourceName` / release namespace. Tests assert identity + selector equality and that `maxUnavailable` stays unset (no min/max conflict).
- **Anti-affinity** — default is `preferredDuringSchedulingIgnoredDuringExecution` / `kubernetes.io/hostname` / weight 100, so single-node installs still schedule. Empty `manager.affinity` → chart default; non-empty → full replace (not merge). Covered by `TestHelmTemplateDefaultControllerAntiAffinity` and `TestHelmTemplateControllerAffinityOverrideReplacesDefault`.
- **Docs** — leader handover gap, co-location drain behavior, “PDB preserves a ready replica not the leader/lease”, singleton maintenance window (`minAvailable: 0` / disable), and “no `nvcrectl setup init` PDB flags” are all called out clearly. Checklist row correctly warns that `manager.affinity` replaces the whole default.

CI (including Test) is green on this SHA.

### Concrete gaps before this is ready to land

### 1. Hard-HA docs example selector is missing `app.kubernetes.io/name`

The chart’s Deployment selector, PDB selector, and default preferred anti-affinity all emit both `nvcre.selectorLabels` keys:

```text
app.kubernetes.io/name: <nvcre.name>
app.kubernetes.io/instance: <Release.Name>
control-plane: manager
```

The required anti-affinity snippet under **Node maintenance and disruption protection** only lists `app.kubernetes.io/instance` + `control-plane: manager`. With the default `nameOverride: nvcre` that is usually enough in practice, but it diverges from what the chart actually renders and from what an operator would copy out of `kubectl get deploy … -o jsonpath='{.spec.selector.matchLabels}'`. Please add `app.kubernetes.io/name: nvcre` (and the same “rename if overridden” note you already have for instance) so the hard-HA example matches the chart contract.

### 2. Upgrade behavior: preferred anti-affinity is a silent default change

Prior chart versions left affinity empty when `manager.affinity: {}`. This PR injects preferred pod anti-affinity on every upgrade that still uses the empty default. Soft affinity should not break single-node clusters, but it *does* change scheduling / descheduler / topology pressure versus today’s installs. Worth an explicit upgrade/release-note bullet (PR body is fine if there is no CHANGELOG) so platform owners are not surprised when two replicas start preferring different hosts after a routine chart bump.

### 3. Generalize the “budget ≥ replica count blocks drains” warning

The singleton case (`replicas: 1` + `minAvailable: 1`) is documented well. The same permanent voluntary-eviction deadlock also happens for `replicas: 2` + `minAvailable: 2` (or `100%`). A one-liner next to the singleton warning — “any `minAvailable` ≥ ready replica count blocks drains” — would cover the misconfig without new validation code. Optional: a helm test that renders `pdb.minAvailable=2` with `manager.replicas=1` and only asserts the values round-trip (not that Helm rejects them), so docs and tests agree this is operator-owned.

### Residual / non-blocking

- **`unhealthyPodEvictionPolicy`**: not exposed. For an HA controller PDB, `AlwaysAllow` (policy/v1) is often desirable so a NotReady standby does not complicate drains when the healthy count is already at `minAvailable`. Fine as a follow-up; depends on the cluster’s min kube version story.
- **Affinity override test** only exercises a dotted `--set` nodeAffinity path. The default-anti-affinity test already covers the empty-values path; no ask unless you want an explicit `affinity: {}` values-file fixture for belt-and-suspenders.

Happy to re-review once the docs selector and upgrade note are sorted. Leaving as COMMENT (not a formal approve).

kaynetu:
@asivanadi0 Addressed the three documentation points in [c282d4c](https://github.com/NVIDIA/cluster-readiness-engine/commit/c282d4cae84607fa5b2ff2a76381f5ca39a6faec).

- Added the name label and guidance for both selector values. [Example](https://github.com/NVIDIA/cluster-readiness-engine/blob/c282d4cae84607fa5b2ff2a76381f5ca39a6faec/docs/operations/deployment.md#L137-L156).
- Added an upgrade note. It explicitly calls out the controller rollout even when PDB is disabled. [Note](https://github.com/NVIDIA/cluster-readiness-engine/blob/c282d4cae84607fa5b2ff2a76381f5ca39a6faec/docs/operations/deployment.md#L135).
- Generalized the healthy-pod eviction warning. Included two replicas with `minAvailable: 2` or `"100%"`, plus percentage rounding. [Warning](https://github.com/NVIDIA/cluster-readiness-engine/blob/c282d4cae84607fa5b2ff2a76381f5ca39a6faec/docs/operations/deployment.md#L165).

The existing [integer override test](https://github.com/NVIDIA/cluster-readiness-engine/blob/c282d4cae84607fa5b2ff2a76381f5ca39a6faec/test/helm/pdb_test.go#L43-L45) already covers `minAvailable: 2` with the default single replica. No extra test was added. The optional eviction-policy change remains outside this update.

Validation: `make verify-doc-links` passed (255 links). `git diff --check` passed. This commit changes documentation only.

asivanadi0 (approved):
Re-reviewed at HEAD `965e0bf965976fb01e6265d2c4ca0406af660ed3` against the three docs asks from review https://github.com/NVIDIA/cluster-readiness-engine/pull/330#pullrequestreview-5159960565 (prior HEAD `37c4115`). Docs-only follow-up in `c282d4c` plus merge to main.

### Prior asks — all FIXED

1. **Hard-HA selector `app.kubernetes.io/name`** — FIXED. Required anti-affinity example now includes `app.kubernetes.io/name: nvcre` alongside instance + `control-plane: manager`, with explicit guidance to match the Deployment selector and adjust both labels if `nameOverride` / release name differ (`docs/operations/deployment.md` L137–L156). Matches `nvcre.selectorLabels` + chart default affinity/PDB selectors.

2. **Upgrade note for new default preferred anti-affinity** — FIXED. Explicit upgrade note at L135: empty `manager.affinity` previously rendered none; upgrade injects preferred hostname anti-affinity and triggers a controller Deployment rollout even when `pdb.enabled` is false; soft affinity still allows co-location / single-node.

3. **Generalize minAvailable ≥ replicas drain-block warning** — FIXED. L165 generalizes beyond the singleton case: voluntary eviction of a healthy controller is blocked while ready count is at or below the required minimum; calls out `replicas: 2` + `minAvailable: 2` / `"100%"`, plus percentage rounding-up. Optional extra helm test not required — existing integer override path already covers `minAvailable: 2` with default singleton replicas.

### Residual (still non-blocking)

- `unhealthyPodEvictionPolicy` still not exposed — fine as follow-up, as before.

### CI / trustee / mergeability note (no merge)

- Author self-posted `/ok to test 965e0bf` (comment 5609915066). Trustee/copy-pr gate **did clear in practice**: `copy-pr-bot` dispatched full NVIDIA-runner workflows on HEAD ~5s later; all completed **success** — beyond DCO + CodeRabbit, including **CI (Lint/Build/Test)**, **UAT (Kind + KWOK + Tilt)**, CodeQL Analyze Go, Container Build, Verify, govulncheck, Fern Docs CI / Preview.
- `mergeable: MERGEABLE`, `mergeStateStatus: BLOCKED` / `reviewDecision: REVIEW_REQUIRED` prior to this approve (approval was the remaining gate). No merge from me.

Approving — ready from the docs/contract side.

ndipebot (commented):
Nothing blocking from me. Two things I'd fix while the branch is open, both inline: the hard-HA example needs a line about rolling updates at two nodes, and the anti-affinity test panics rather than fails when the default goes away.

docs/operations/deployment.md
```diff
@@ -151,7 +116,53 @@ manager:
...
+            matchLabels:
+              app.kubernetes.io/name: nvcre
+              app.kubernetes.io/instance: nvcre
+              control-plane: manager
+          topologyKey: kubernetes.io/hostname
+```
+
+The default preferred rule does not guarantee separation, so both replicas may still share a node. During a drain, one pod can be evicted, but the PDB then waits for the Deployment to schedule and ready a replacement elsewhere before allowing the other eviction. If no replacement can become Ready, the drain remains blocked. Co-location also leaves both replicas exposed to an involuntary failure of that node, which a PDB cannot prevent.
+
+This placement requires at least two eligible infrastructure nodes. The PDB can still allow eviction of the leader; it preserves a ready replica, not uninterrupted reconciliation or process memory. It does not protect against node failure, direct pod deletion, or Deployment rolling updates.
```
ndipebot: Two is right for placing the replicas, but it's also the count where every later controller upgrade wedges, and nothing on the page says so.

There's no `spec.strategy` on the controller Deployment, so at `replicas: 2` you get the Kubernetes defaults: `maxSurge` 25% rounded up is 1, `maxUnavailable` 25% rounded down is 0. An upgrade that touches the pod template has to surge a pod before it can retire one, and with both infra nodes holding old-ReplicaSet pods the required rule above has nowhere to put it. `pod-template-hash` isn't in that `matchLabels`, so it repels new-RS pods from old-RS pods just as happily. The rollout sits Pending until `progressDeadlineSeconds`, and step 4 of `## Upgrades` further down this page (`kubectl rollout status -n nvcre deploy/nvcre-manager`) fails.

Drains are fine at two once you `uncordon`, so it's only the upgrade path that bites.

```suggestion
This placement requires at least two eligible infrastructure nodes, and three if you want upgrades to roll without intervention. The chart sets no `spec.strategy`, so at two replicas the rolling update defaults to `maxSurge: 1` and `maxUnavailable: 0`, and on exactly two nodes the surge pod has nowhere to schedule. If you only have two, set `spec.strategy.rollingUpdate.maxUnavailable: 1` or keep the preferred anti-affinity rule instead of the required one. The PDB can still allow eviction of the leader; it preserves a ready replica, not uninterrupted reconciliation or process memory. It does not protect against node failure, direct pod deletion, or Deployment rolling updates.
```

test/helm/pdb_test.go
```diff
@@ -0,0 +1,192 @@
...
+	dir := chartDir(t)
+	requireChartInputs(t, dir)
+
+	rendered, err := helmTemplate(dir, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	deployment := decodeManagerDeployment(t, rendered)
+	preferred := deployment.Spec.Template.Spec.Affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution
```
ndipebot: This chain assumes `Affinity` and `PodAntiAffinity` are non-nil, but the regression this test exists to catch is exactly the one that leaves them nil. Putting the old `{{- with .Values.manager.affinity }}` block back in `deployment.yaml` makes the run die with `SIGSEGV ... addr=0x10` here instead of failing an assertion, and it takes the rest of the package with it: `TestHelmTemplateControllerAffinityOverrideReplacesDefault`, `TestHelmTemplateRendersConcurrencyArgs` and `TestHelmTemplatePinsImageByDigest` never run.

`TestHelmTemplateControllerAffinityOverrideReplacesDefault` already guards this way at line 161.

```suggestion
	affinity := deployment.Spec.Template.Spec.Affinity
	if affinity == nil || affinity.PodAntiAffinity == nil {
		t.Fatal("default pod anti-affinity was not rendered")
	}
	preferred := affinity.PodAntiAffinity.PreferredDuringSchedulingIgnoredDuringExecution
```

docs/operations/deployment.md
kaynetu: Documented the two-node rollout stall and alternatives, including the chart’s strategy limitation ([e74a267](https://github.com/NVIDIA/cluster-readiness-engine/blob/e74a26759c35482dd8987181fee508c994ec6ac3/docs/operations/deployment.md#L198)). `make verify-doc-links` passed (265 links).

test/helm/pdb_test.go
kaynetu: Added the nil guard so missing anti-affinity fails with a clear assertion ([e74a267](https://github.com/NVIDIA/cluster-readiness-engine/blob/e74a26759c35482dd8987181fee508c994ec6ac3/test/helm/pdb_test.go#L128-L132)). The Helm test suite passed.

asivanadi0 (approved):
Re-reviewed at HEAD `fd6da330b34a37374b764ac06dcf667d40471073` (prior APPROVE was on `c0ba141`). Diff vs that approve for this PR's files is exactly `e74a267` (`docs/operations/deployment.md` + `test/helm/pdb_test.go`); the rest of `c0ba141...fd6da33` is merge-from-main (#331/#332).

### ndipebot asks (https://github.com/NVIDIA/cluster-readiness-engine/pull/330#pullrequestreview-5172216114) — FIXED

1. **Two-node hard-HA rollout stall** — FIXED. Hard-HA section now states: rolling upgrades at `replicas: 2` need a third eligible node for the surge pod; chart leaves Deployment strategy unset so defaults are `maxSurge=1` / `maxUnavailable=0`; required anti-affinity also matches old-RS pods, so on exactly two nodes the surge stays Pending. Documents the two-node alternatives (keep preferred anti-affinity, or set `maxUnavailable: 1` out-of-band) and that the chart does not expose strategy as a Helm value (`docs/operations/deployment.md` ~L198–L200).

2. **Default anti-affinity test nil-deref** — FIXED. `TestHelmTemplateDefaultControllerAntiAffinity` now nil-guards `Affinity` / `PodAntiAffinity` and `t.Fatal`s with a clear message instead of SIGSEGV (`test/helm/pdb_test.go` L128–L132), matching the override test's pattern.

### Prior asivanadi0 docs asks (still hold on HEAD)

Hard-HA selector includes `app.kubernetes.io/name: nvcre`; upgrade note for injecting preferred anti-affinity on empty `manager.affinity`; generalized minAvailable ≥ ready-count drain-block warning — all still present and unchanged by `e74a267`.

### CI / merge

Checks on HEAD are green (CI Lint/Build/Test, UAT, CodeQL, Container Build, Verify, govulncheck, Fern, DCO, CodeRabbit). No merge from me.

Approving — ndipebot + prior docs asks are addressed; ready from the docs/contract/test side.

ndipebot (approved):
/lgtm
/approve

PR #329, opened 2026-09-09
docs: catalog entries are compile-time embedded, document runtime alternatives

ndipebot:
## Summary

The custom catalog guide says entries are discovered "by scanning this directory tree at startup", which reads as runtime drop-in. They are `go:embed`'d (`pkg/catalog/loader.go:66`); adding an entry means rebuilding `nvcrectl` and shipping a new controller image. A user reading the guide as written would drop a YAML into a deployed system and wait for a category that never appears.

- correct the wording in `custom-catalog-entries.md` and `concepts/catalog.md`, keeping the accurate half: no registration step, the loader discovers entries in the embedded tree automatically
- add the missing rebuild steps to the guide's verify section (`nvcrectl` for local render, controller image to run the category on a cluster)
- add a "Runtime alternatives" section covering the two paths that do work without a rebuild: `WorkloadRun` (thresholds, goodput/bandwidth measurement, per-node results in status, `spec.overrides`) and a hand-authored `Workflow` CR (full control over jobTemplate/dependencies/orchestration/overrides, platform and GPU override matching still applies at reconcile), so readers land on an answer instead of a dead end
- cross-reference #244 for delivering curated catalog updates independently of operator upgrades

`make verify-doc-links` passes (257 links).

## Related Issue

Closes #323

## Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] API / CRDs
- [ ] Controller / Reconcilers
- [ ] Catalog / Workloads
- [ ] CLI (nvcrectl)
- [ ] Helm / Deployment
- [x] Documentation / CI
- [ ] Other: ____________

## Testing
- [x] Tests pass locally
- [ ] Manual testing completed
- [x] No breaking changes (or documented)

## Checklist
- [x] Self-review completed
- [x] Commits are signed off for the DCO (`git commit -s`)
- [ ] `make manifests generate` run (if `*_types.go` was modified)
- [ ] Golden files updated (if integration test output changed)
- [x] Documentation updated (if needed)
- [x] Ready for review

PR #334, opened 2026-09-09
docs: admission-policy samples for Kyverno and Sigstore policy-controller (#272)

### LINKED ISSUE #272: Admission-policy samples (Kyverno, Sigstore policy-controller)
Part of #262. **Phase 3 — Document.** Depends on: #271. Size: S.

## Problem

Verification a human runs once at install time does not stop a compromised image from being pulled on the next pod restart. Operators running this controller on GPU fleets generally already run an admission layer; we should hand them a policy rather than make them derive our identity contract from prose.

## Scope

**Rescoped 2026-09-03 to one sample, then parked.** The first attempt is recorded in the comments and was not merged. The earlier claim here — that policy-controller fails only because v0.13.1 vendors `sigstore/cosign/v2`, and resolves when it moves to v3 — was **wrong**: v0.15.0 (2026-03-26) already vendors cosign v3, [#1958](https://github.com/sigstore/policy-controller/issues/1958) is filed against v0.15.1 and still open, and [#1899](https://github.com/sigstore/policy-controller/issues/1899) reports a `signatureFormat: bundle` setting that works even on v0.13.1. policy-controller was tested only at the chart default with default settings, so it is **not** confirmed blocked and should be re-tested before being ruled out.

- `config/samples/policy/kyverno-verify-images.yaml` — a Kyverno **`ImageValidatingPolicy`** (not `ClusterPolicy`/`verifyImages`, which is on the same legacy path and fails identically) for `ghcr.io/nvidia/cluster-readiness-engine/manager*`, keyless, exact issuer, identity pinned to `attest.yml` at a release tag.
- A comment explaining what a `refs/tags/.+` regexp would and would not buy, so an operator loosening it does so knowingly.
- An enforcement section in `docs/operations/verifying-artifacts.md` (#271), including a compatibility table for which verifiers can read the format we publish.

## Acceptance criteria

- [ ] The policy applied to a Kind cluster admits the signed release image pinned by digest.
- [ ] It rejects an unsigned image and an image signed by a different identity.
- [ ] Covered by a test so the sample cannot rot.

**Unchecked 2026-09-03.** These were ticked against a branch that was reviewed and then parked; none of it is on `main`. The Kind results below are real and reproducible, but they describe code that does not exist in the repository, so the boxes were asserting something untrue. They stay unchecked until the sample lands.

Verified on Kind (Kyverno v1.19.0) via server-side dry-run, so the admission webhook ran:

| Case | Result |
|---|---|
| signed release image, by digest | admitted |
| `manager:main-ce70f5f` — signed under `publish.yml@refs/heads/main` | denied |
| unsigned / nonexistent tag | denied |
| `busybox:1.36`, outside the glob | admitted, untouched |

`TestPolicySamplePinsTheReleaseIdentity` pins the kind, the fail-closed policy, the issuer, the identity prefix and the glob narrowness. The live admission behaviour is not reproduced in CI — it needs a cluster and a network — and that is stated in the test's own comment rather than implied.

asivanadi0:
Closes #272

## Summary
Revive and land admission-policy samples for verifying NVCRE release artifacts with Kyverno and Sigstore policy-controller, aligned to the current ADR-074 / verifying-artifacts identity contract (`signatureFormat: bundle`, exact subject pin).

## What’s included
- Kyverno ImageValidatingPolicy sample (namespace opt-in, Fail after opt-in, provenance required, exact subject / no manager*)
- Sigstore policy-controller ClusterImagePolicy sample with `signatureFormat: bundle`
- Docs updates in verifying-artifacts.md + RELEASE.md pin note
- Structural tests in `test/docspolicy/policy_samples_test.go`
- Kind admit/deny repro documented (full Kind not required to merge structural coverage)

## Test plan
- [x] `make test` / package tests for `test/docspolicy`
- [x] Samples match current published identity contract
- [ ] Maintainer `/ok to test` for NVIDIA runners if required
- [ ] Optional Kind admit/deny smoke per docs

## Risk
Docs + sample YAML + structural tests only. No runtime controller changes. Misconfigured Fail policies can block pulls if applied cluster-wide — samples stay namespace-opt-in with Fail after opt-in as documented.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -51,6 +51,12 @@ people do not tag at the same time.
    made visible only by the `Verify release` job, after it has verified every published
    artifact. If that job fails, the release stays a draft — see Troubleshooting.
 5. Check the published release, then announce it.
+6. If you maintain the admission-policy samples, bump the pinned `subject:` tag in
+   `config/samples/policy/kyverno-verify-images.yaml` and
+   `config/samples/policy/policy-controller-verify-images.yaml` to the new release
+   (or open a follow-up PR). Nothing in CI rewrites those strings; a stale pin is a
+   policy that will deny the next upgrade for anyone who applied the sample as-is.
+   See [Verifying release artifacts](docs/operations/verifying-artifacts.md#admission-policy-samples).
 
 Do not publish a draft release by hand. A draft left behind by a failed `Verify release`
 is a release the pipeline determined it could not verify; publishing it from the UI is
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -35,7 +35,15 @@ people do not tag at the same time.
 1. Make sure `main` is green. Every required check must pass: Lint, Build, Test, Verify,
    and UAT.
 2. Decide the version. See Versioning above.
-3. Tag the commit on `main` and push the tag.
+3. Bump the pinned `subject:` tag in
+   `config/samples/policy/kyverno-verify-images.yaml` and
+   `config/samples/policy/policy-controller-verify-images.yaml` to the version you are
+   about to cut, and land that change on `main` **before** tagging. Nothing in CI
+   rewrites those strings. Tagging first and bumping later leaves the samples at that
+   tag pinned to the previous release — exactly the stale-pin outage the bump exists to
+   prevent. See [Verifying release artifacts](docs/operations/verifying-artifacts.md#admission-policy-samples).
+4. Tag the commit on `main` (the one that already carries the bumped pins) and push the
+   tag.
 
    ```bash
    git checkout main && git pull --ff-only
@@ -47,16 +55,10 @@ people do not tag at the same time.
    `NVIDIA/cluster-readiness-engine`, which is usually `upstream`.
 
    Sign the tag (`-s`). Pushing the tag is the release trigger.
-4. Watch the `Release` workflow. The GitHub Release is created as a **draft** and is
+5. Watch the `Release` workflow. The GitHub Release is created as a **draft** and is
    made visible only by the `Verify release` job, after it has verified every published
    artifact. If that job fails, the release stays a draft — see Troubleshooting.
-5. Check the published release, then announce it.
-6. If you maintain the admission-policy samples, bump the pinned `subject:` tag in
-   `config/samples/policy/kyverno-verify-images.yaml` and
-   `config/samples/policy/policy-controller-verify-images.yaml` to the new release
-   (or open a follow-up PR). Nothing in CI rewrites those strings; a stale pin is a
-   policy that will deny the next upgrade for anyone who applied the sample as-is.
-   See [Verifying release artifacts](docs/operations/verifying-artifacts.md#admission-policy-samples).
+6. Check the published release, then announce it.
 
 Do not publish a draft release by hand. A draft left behind by a failed `Verify release`
 is a release the pipeline determined it could not verify; publishing it from the UI is
--- a/config/samples/policy/kyverno-verify-images.yaml
+++ b/config/samples/policy/kyverno-verify-images.yaml
@@ -1,0 +1,144 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Kyverno ImageValidatingPolicy for the NVCRE manager image.
+#
+# WHY ImageValidatingPolicy (not ClusterPolicy / verifyImages)
+# -----------------------------------------------------------
+# NVCRE publishes Sigstore *new-bundle* signatures as OCI referrers
+# (https://sigstore.dev/cosign/sign/v1 and https://slsa.dev/provenance/v1).
+# There is no legacy `sha256-<digest>.sig` tag for current releases.
+# Kyverno's deprecated ClusterPolicy `verifyImages` looks only for that
+# legacy tag and reports "no signatures found" against a correctly signed
+# image. ImageValidatingPolicy reads the referrer / bundle path.
+#
+# BEFORE YOU APPLY
+# ----------------
+# 1. Edit `subject:` below so the tag matches the release you actually run.
+#    Applying a pin that does not match the image already in the cluster
+#    will deny the next upgrade, rollout restart, node drain, or crash
+#    replacement until you edit the policy again.
+# 2. Label the namespace that runs the controller (default `nvcre`):
+#      kubectl label namespace nvcre \
+#        kubernetes.nvcre.nvidia.com/image-admission=enforce
+# 3. Prefer deploying the manager by *index digest* (see verifying-artifacts.md).
+#    Provenance is attested to the multi-platform index digest only; a
+#    per-platform child manifest carries a signature + CycloneDX SBOM and
+#    no SLSA provenance, so requiring provenance against a platform digest
+#    correctly denies.
+#
+# PARKED FAILURE MODES THIS SAMPLE ADDRESSES (issue #272)
+# -------------------------------------------------------
+# 1. Namespace scoping — matchConstraints.namespaceSelector confines the
+#    ValidatingAdmissionPolicy / webhook to labeled namespaces. Without it,
+#    failurePolicy: Fail is effectively cluster-wide for every Pod /
+#    Deployment / Job create in every namespace except Kyverno's excludes.
+# 2. failurePolicy — Fail is intentional (fail closed), but only safe once
+#    namespace-scoped. matchImageReferences is evaluated *inside* Kyverno
+#    after the API server has already called the webhook; it cannot narrow
+#    which requests reach a Fail webhook. Namespace selectors can.
+# 3. Provenance — validations require verifyAttestationSignatures for
+#    https://slsa.dev/provenance/v1 in addition to the image signature,
+#    matching the human-run `cosign verify` + `verify-attestation --type
+#    slsaprovenance1` contract in docs/operations/verifying-artifacts.md.
+# 4. subject / glob — `subject:` is a STRICT match. Putting `refs/tags/.+`
+#    under `subject:` matches no real SAN and denies everything; the regexp
+#    form needs `subjectRegExp:`. The image globs name the repository path
+#    exactly (no `manager*` prefix wildcard that crosses path segments).
+#
+# Identity contract (ADR-074 / verifying-artifacts.md):
+#   issuer:  https://token.actions.githubusercontent.com
+#   subject: https://github.com/NVIDIA/cluster-readiness-engine/.github/workflows/attest.yml@refs/tags/<TAG>
+#
+# Requires Kyverno >= v1.19 (ImageValidatingPolicy stable on policies.kyverno.io/v1).
+apiVersion: policies.kyverno.io/v1
+kind: ImageValidatingPolicy
+metadata:
+  name: nvcre-manager-verify-images
+  labels:
+    app.kubernetes.io/name: cluster-readiness-engine
+    app.kubernetes.io/component: admission-sample
+spec:
+  validationActions: [Deny]
+  failurePolicy: Fail
+  webhookConfiguration:
+    timeoutSeconds: 20
+  evaluation:
+    background:
+      # Do not background-scan existing objects: a stale pin must not page
+      # operators for workloads that were admitted under a previous release.
+      enabled: false
+  matchConstraints:
+    # Confines webhook evaluation to namespaces the operator opted in.
+    # failurePolicy: Fail still fails closed — but only for these namespaces.
+    namespaceSelector:
+      matchExpressions:
+        - key: kubernetes.nvcre.nvidia.com/image-admission
+          operator: In
+          values: ["enforce"]
+    resourceRules:
+      # Include controllers that own pod templates. Matching only Pods lets a
+      # Deployment create succeed and then fail when the Pod is created.
+      - apiGroups: [""]
+        apiVersions: ["v1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["pods"]
+      - apiGroups: ["apps"]
+        apiVersions: ["v1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["deployments", "replicasets", "daemonsets", "statefulsets"]
+      - apiGroups: ["batch"]
+        apiVersions: ["v1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["jobs", "cronjobs"]
+  matchImageReferences:
+    # Narrow globs on purpose. A trailing `manager*` can match manager-debug,
+    # managerial-tool, or manager/nested/extra depending on the glob library.
+    # Tag and digest forms are listed explicitly instead.
+    - glob: "ghcr.io/nvidia/cluster-readiness-engine/manager"
+    - glob: "ghcr.io/nvidia/cluster-readiness-engine/manager:*"
+    - glob: "ghcr.io/nvidia/cluster-readiness-engine/manager@*"
+  attestors:
+    - name: nvcreRelease
+      cosign:
+        keyless:
+          identities:
+            # EDIT THIS TAG to the release you run before applying.
+            # Exact identity — never a prefix regexp that also accepts
+            # publish.yml@refs/heads/main development images.
+            - issuer: "https://token.actions.githubusercontent.com"
+              subject: "https://github.com/NVIDIA/cluster-readiness-engine/.github/workflows/attest.yml@refs/tags/v0.2.0"
+              # Loosening (operator choice, not the published contract):
+              # subject: is a STRICT string match. A value like
+              #   .../attest.yml@refs/tags/.+
+              # under subject: matches no certificate SAN and denies every
+              # image. To accept any release tag while still rejecting
+              # refs/heads/main, replace subject: with subjectRegExp:
+              #   subjectRegExp: "^https://github\\.com/NVIDIA/cluster-readiness-engine/\\.github/workflows/attest\\.yml@refs/tags/.+$"
+              # Do not use subjectRegExp that omits the workflow or the
+              # refs/tags/ anchor — that re-opens main-<sha> development images.
+        ctlog:
+          url: "https://rekor.sigstore.dev"
+  attestations:
+    # CamelCase name so CEL can use attestations.slsaProvenance (hyphens are not
+    # valid field selectors).
+    - name: slsaProvenance
+      intoto:
+        # Same predicate the docs verify with --type slsaprovenance1.
+        type: "https://slsa.dev/provenance/v1"
+  validations:
+    - expression: >-
+        images.containers.map(image,
+          verifyImageSignatures(image, [attestors.nvcreRelease])).all(e, e > 0)
+      message: >-
+        NVCRE manager image signature verification failed. Expected a new-bundle
+        cosign signature under attest.yml@refs/tags/<pinned-tag>. Legacy
+        ClusterPolicy/verifyImages cannot see these signatures.
+    - expression: >-
+        images.containers.map(image,
+          verifyAttestationSignatures(image, attestations.slsaProvenance,
+            [attestors.nvcreRelease])).all(e, e > 0)
+      message: >-
+        NVCRE manager image is missing a verified SLSA Build Provenance v1
+        attestation on the index digest (ADR-074). Per-platform manifests carry
+        an SBOM, not provenance — pin the index digest.
--- a/config/samples/policy/kyverno-verify-images.yaml
+++ b/config/samples/policy/kyverno-verify-images.yaml
@@ -26,6 +26,12 @@
 #    per-platform child manifest carries a signature + CycloneDX SBOM and
 #    no SLSA provenance, so requiring provenance against a platform digest
 #    correctly denies.
+# 4. If you mirrored the controller image off ghcr.io, rewrite every
+#    matchImageReferences glob below to your mirror repository path.
+#    Kyverno pre-filters images.* to glob matches and initialises an empty
+#    list when nothing matches — both validations then succeed vacuously
+#    (fail-open). A silent no-op on the documented air-gap path is worse
+#    than a deny.
 #
 # PARKED FAILURE MODES THIS SAMPLE ADDRESSES (issue #272)
 # -------------------------------------------------------
@@ -61,6 +67,17 @@ metadata:
 spec:
   validationActions: [Deny]
   failurePolicy: Fail
+  # Explicit CRD defaults (v1.19): each of these defaults to true. Setting them
+  # here documents that applying this sample also registers a *mutating*
+  # webhook that rewrites matching tags to digests (mutateDigest), and that
+  # required/verifyDigest run alongside the CEL validations below. required
+  # is load-bearing for init/ephemeral coverage even before the concatenations
+  # in validations: (EnforceRequired denies matched images that no validation
+  # checked).
+  validationConfigurations:
+    mutateDigest: true
+    verifyDigest: true
+    required: true
   webhookConfiguration:
     timeoutSeconds: 20
   evaluation:
@@ -82,7 +99,10 @@ spec:
       - apiGroups: [""]
         apiVersions: ["v1"]
         operations: ["CREATE", "UPDATE"]
-        resources: ["pods"]
+        # pods/ephemeralcontainers is required for kubectl debug / ephemeral
+        # coverage; images.ephemeralContainers is populated only when this
+        # subresource is in matchConstraints.
+        resources: ["pods", "pods/ephemeralcontainers"]
       - apiGroups: ["apps"]
         apiVersions: ["v1"]
         operations: ["CREATE", "UPDATE"]
@@ -95,6 +115,8 @@ spec:
     # Narrow globs on purpose. A trailing `manager*` can match manager-debug,
     # managerial-tool, or manager/nested/extra depending on the glob library.
     # Tag and digest forms are listed explicitly instead.
+    # Hard-pinned to ghcr.io for the public default. Mirror / air-gap installs
+    # must rewrite these (see BEFORE YOU APPLY #4) — unmatched globs fail open.
     - glob: "ghcr.io/nvidia/cluster-readiness-engine/manager"
     - glob: "ghcr.io/nvidia/cluster-readiness-engine/manager:*"
     - glob: "ghcr.io/nvidia/cluster-readiness-engine/manager@*"
@@ -105,7 +127,9 @@ spec:
           identities:
             # EDIT THIS TAG to the release you run before applying.
             # Exact identity — never a prefix regexp that also accepts
-            # publish.yml@refs/heads/main development images.
+            # attest.yml@refs/heads/main development images (publish.yml now
+            # routes all signing through attest.yml; main builds carry
+            # attest.yml@refs/heads/main).
             - issuer: "https://token.actions.githubusercontent.com"
               subject: "https://github.com/NVIDIA/cluster-readiness-engine/.github/workflows/attest.yml@refs/tags/v0.2.0"
               # Loosening (operator choice, not the published contract):
@@ -115,8 +139,9 @@ spec:
               # image. To accept any release tag while still rejecting
               # refs/heads/main, replace subject: with subjectRegExp:
               #   subjectRegExp: "^https://github\\.com/NVIDIA/cluster-readiness-engine/\\.github/workflows/attest\\.yml@refs/tags/.+$"
-              # Do not use subjectRegExp that omits the workflow or the
-              # refs/tags/ anchor — that re-opens main-<sha> development images.
+              # Do not use subjectRegExp that omits the repository path, the
+              # workflow, or the refs/tags/ anchor — that re-opens main-<sha>
+              # development images and foreign forks.
         ctlog:
           url: "https://rekor.sigstore.dev"
   attestations:
@@ -127,15 +152,23 @@ spec:
         # Same predicate the docs verify with --type slsaprovenance1.
         type: "https://slsa.dev/provenance/v1"
   validations:
+    # Concatenate containers + initContainers + ephemeralContainers. The three
+    # are distinct keys on images (pkg/cel/compiler/images.go); containers alone
+    # leaves init/ephemeral to validationConfigurations.required (deny with a
+    # "no policy performed a check" message). Explicit coverage keeps the deny
+    # message accurate. ephemeralContainers only appears when
+    # pods/ephemeralcontainers is in matchConstraints above.
     - expression: >-
-        images.containers.map(image,
+        (images.?containers.orValue([]) + images.?initContainers.orValue([]) +
+          images.?ephemeralContainers.orValue([])).map(image,
           verifyImageSignatures(image, [attestors.nvcreRelease])).all(e, e > 0)
       message: >-
         NVCRE manager image signature verification failed. Expected a new-bundle
         cosign signature under attest.yml@refs/tags/<pinned-tag>. Legacy
         ClusterPolicy/verifyImages cannot see these signatures.
     - expression: >-
-        images.containers.map(image,
+        (images.?containers.orValue([]) + images.?initContainers.orValue([]) +
+          images.?ephemeralContainers.orValue([])).map(image,
           verifyAttestationSignatures(image, attestations.slsaProvenance,
             [attestors.nvcreRelease])).all(e, e > 0)
       message: >-
--- a/config/samples/policy/kyverno-verify-images.yaml
+++ b/config/samples/policy/kyverno-verify-images.yaml
@@ -99,9 +99,12 @@ spec:
       - apiGroups: [""]
         apiVersions: ["v1"]
         operations: ["CREATE", "UPDATE"]
-        # pods/ephemeralcontainers is required for kubectl debug / ephemeral
-        # coverage; images.ephemeralContainers is populated only when this
-        # subresource is in matchConstraints.
+        # pods/ephemeralcontainers is required so the API server routes
+        # kubectl debug / ephemeral-container patches to this policy. Without
+        # it, CREATE/UPDATE on pods alone never sees those requests.
+        # (images.ephemeralContainers is always present on built-in GVRs —
+        # initialised to [] — and can be non-empty on a plain pods UPDATE;
+        # the subresource rule is about interception, not key presence.)
         resources: ["pods", "pods/ephemeralcontainers"]
       - apiGroups: ["apps"]
         apiVersions: ["v1"]
@@ -156,8 +159,9 @@ spec:
     # are distinct keys on images (pkg/cel/compiler/images.go); containers alone
     # leaves init/ephemeral to validationConfigurations.required (deny with a
     # "no policy performed a check" message). Explicit coverage keeps the deny
-    # message accurate. ephemeralContainers only appears when
-    # pods/ephemeralcontainers is in matchConstraints above.
+    # message accurate. The pods/ephemeralcontainers matchConstraints rule above
+    # is what routes kubectl debug into these checks — the ephemeralContainers
+    # key itself is always populated (as [] when empty).
     - expression: >-
         (images.?containers.orValue([]) + images.?initContainers.orValue([]) +
           images.?ephemeralContainers.orValue([])).map(image,
--- a/config/samples/policy/policy-controller-verify-images.yaml
+++ b/config/samples/policy/policy-controller-verify-images.yaml
@@ -1,0 +1,82 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Sigstore policy-controller ClusterImagePolicy for the NVCRE manager image.
+#
+# WHY signatureFormat: bundle
+# --------------------------
+# NVCRE publishes cosign *new-bundle* artifacts as OCI referrers (ADR-074 pins
+# cosign v3.1.3, where --new-bundle-format defaults to true). policy-controller
+# defaults to the legacy signature format and will not see our referrers unless
+# `signatureFormat: bundle` is set on the authority.
+#
+# In bundle mode, policy-controller verifies *attestations*, not the legacy
+# plain-signature path. The sample therefore requires:
+#   - https://sigstore.dev/cosign/sign/v1  (the signed identity, bundle form)
+#   - https://slsa.dev/provenance/v1      (the provenance the docs also check)
+# See sigstore/policy-controller#1899 for the bundle + cosign/sign/v1 shape.
+#
+# NAMESPACE SCOPING
+# -----------------
+# policy-controller is opt-in by namespace label (default chart behaviour):
+#   kubectl label namespace nvcre policy.sigstore.dev/include=true
+# Unlabeled namespaces are not evaluated, so a Fail / enforce mode here does
+# not strand workload creation cluster-wide. Do not flip the webhook to
+# opt-out without understanding the blast radius.
+#
+# BEFORE YOU APPLY
+# ----------------
+# 1. Edit the `subject:` pin to the release tag you run.
+# 2. Label the namespace (above).
+# 3. Deploy the manager by index digest — provenance is on the index only.
+#
+# Tested versions to prefer when re-checking: policy-controller >= v0.15.0
+# (vendors cosign v3). Chart-default older versions may still work with
+# signatureFormat: bundle; do not treat a default-settings failure as a
+# format incompatibility without that field set.
+#
+# Identity contract (ADR-074 / verifying-artifacts.md):
+#   issuer:  https://token.actions.githubusercontent.com
+#   subject: https://github.com/NVIDIA/cluster-readiness-engine/.github/workflows/attest.yml@refs/tags/<TAG>
+apiVersion: policy.sigstore.dev/v1beta1
+kind: ClusterImagePolicy
+metadata:
+  name: nvcre-manager-verify-images
+  labels:
+    app.kubernetes.io/name: cluster-readiness-engine
+    app.kubernetes.io/component: admission-sample
+spec:
+  mode: enforce
+  images:
+    # golang filepath globs; `**` matches across `/`. Do not use a bare
+    # `manager*` suffix — that over-matches manager-debug and nested paths.
+    - glob: "ghcr.io/nvidia/cluster-readiness-engine/manager"
+    - glob: "ghcr.io/nvidia/cluster-readiness-engine/manager:*"
+    - glob: "ghcr.io/nvidia/cluster-readiness-engine/manager@sha256:*"
+  authorities:
+    - name: nvcre-release
+      keyless:
+        url: https://fulcio.sigstore.dev
+        identities:
+          # EDIT THIS TAG to the release you run before applying.
+          - issuer: "https://token.actions.githubusercontent.com"
+            subject: "https://github.com/NVIDIA/cluster-readiness-engine/.github/workflows/attest.yml@refs/tags/v0.2.0"
+            # Loosening: subject is a STRICT match. A pattern under subject
+            # such as refs/tags/.+ matches no SAN. Use subjectRegExp instead:
+            #   subjectRegExp: "^https://github\\.com/NVIDIA/cluster-readiness-engine/\\.github/workflows/attest\\.yml@refs/tags/.+$"
+            # Still reject refs/heads/main and any identity that omits attest.yml.
+      # Required for NVCRE's referrer / new-bundle layout. Without this field
+      # the authority looks for legacy .sig tags we do not publish on current
+      # release digests.
+      signatureFormat: bundle
+      ctlog:
+        url: https://rekor.sigstore.dev
+      attestations:
+        # Bundle format exposes the cosign signature as this predicate type
+        # (not as a separate legacy signature object).
+        - name: signature
+          predicateType: https://sigstore.dev/cosign/sign/v1
+        # Same predicate docs/operations/verifying-artifacts.md checks with
+        # cosign verify-attestation --type slsaprovenance1.
+        - name: provenance
+          predicateType: https://slsa.dev/provenance/v1
--- a/config/samples/policy/policy-controller-verify-images.yaml
+++ b/config/samples/policy/policy-controller-verify-images.yaml
@@ -20,20 +20,32 @@
 # -----------------
 # policy-controller is opt-in by namespace label (default chart behaviour):
 #   kubectl label namespace nvcre policy.sigstore.dev/include=true
-# Unlabeled namespaces are not evaluated, so a Fail / enforce mode here does
-# not strand workload creation cluster-wide. Do not flip the webhook to
-# opt-out without understanding the blast radius.
+# Unlabeled namespaces are not evaluated. Labelling a namespace does *not*
+# narrow enforcement to images that match this ClusterImagePolicy alone:
+# chart-default `no-match-policy` is `deny`, so every image in a labeled
+# namespace that matches no ClusterImagePolicy is rejected. A Fail / enforce
+# mode here therefore does not strand workload creation cluster-wide, but it
+# *does* deny non-NVCRE images inside the labeled namespace unless the
+# operator has separately set `no-match-policy` to `allow` or `warn`. Do not
+# flip the webhook to opt-out without understanding the blast radius.
 #
 # BEFORE YOU APPLY
 # ----------------
 # 1. Edit the `subject:` pin to the release tag you run.
-# 2. Label the namespace (above).
+# 2. Label the namespace (above) — and decide `no-match-policy` if that
+#    namespace runs any non-NVCRE images.
 # 3. Deploy the manager by index digest — provenance is on the index only.
+# 4. If you mirrored the controller image off ghcr.io, rewrite every
+#    `images[].glob` below to your mirror repository path. Globs that match
+#    nothing leave the image under no-match-policy (deny by default), which
+#    is a hard deny rather than Kyverno's empty-list fail-open — still wrong
+#    for a mirrored manager image you intended to admit under this policy.
 #
 # Tested versions to prefer when re-checking: policy-controller >= v0.15.0
 # (vendors cosign v3). Chart-default older versions may still work with
 # signatureFormat: bundle; do not treat a default-settings failure as a
-# format incompatibility without that field set.
+# format incompatibility without that field set. This sample shape has not
+# been re-tested end-to-end for issue #272; see verifying-artifacts.md.
 #
 # Identity contract (ADR-074 / verifying-artifacts.md):
 #   issuer:  https://token.actions.githubusercontent.com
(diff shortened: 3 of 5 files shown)


asivanadi0:
Rebased onto latest main.

asivanadi0:
Rebased onto latest main.

asivanadi0:
Rebased onto latest main.

asivanadi0:
Hi @lalitadithya @ndipebot @mchmarny — gentle bump when you have a moment. Admission-policy samples for Kyverno/Sigstore (closes #272) is rebased onto latest main and ready for review. It still needs a trustee `/ok to test` so the NVIDIA runner checks can run. Happy to address any notes. Thanks!

asivanadi0:
**Docs Verify failure on [run 34485314123](https://github.com/NVIDIA/cluster-readiness-engine/actions/runs/34485314123):** our admission-policy edits added kubectl examples as bash fences on `docs/operations/verifying-artifacts.md`. Docs Verify extracts and runs every bash block, so those commands dialed localhost:8080 (no cluster) and failed at block 13 with exit 1. Same check passed on #308 (does not touch that page).

**Fix in 98519ec:** retargeted the three cluster-only apply/Kind repro fences to shell (Docs Verify only runs bash), and noted that in the page. make lint + go test ./test/docspolicy/... green locally.

**/ok to test:** yes — after this new SHA, please re-post /ok to test so trusted checks (UAT etc.) run on the fix commit. I will not post it myself.

asivanadi0:
Rebased onto `main` (includes the #325–#328 wave). Head moved `98519ec` → `568b6c8`; another trustee `/ok to test` is needed for NVIDIA runners.

asivanadi0:
Rebased onto `main` (post-#329/#331/#332). Fresh trustee `/ok to test` needed if the SHA moved.

config/samples/policy/kyverno-verify-images.yaml
```diff
@@ -0,0 +1,144 @@
...
+# Requires Kyverno >= v1.19 (ImageValidatingPolicy stable on policies.kyverno.io/v1).
+apiVersion: policies.kyverno.io/v1
+kind: ImageValidatingPolicy
+metadata:
+  name: nvcre-manager-verify-images
+  labels:
+    app.kubernetes.io/name: cluster-readiness-engine
+    app.kubernetes.io/component: admission-sample
+spec:
+  validationActions: [Deny]
```
ndipebot, line 68: Applying this turns on three things the sample never mentions.

In the v1.19.0 CRD, `spec.validationConfigurations` carries `default: {}`, and `mutateDigest`, `required` and `verifyDigest` each carry `default: true`. So this policy also registers a **mutating** webhook that rewrites `.../manager:v0.2.0` into `.../manager@sha256:...` on matched workloads, and it enables `required` and `verifyDigest` alongside it.

I do not think any of those defaults are wrong for us. `required` is arguably doing useful work (see my note on the validations below). But the sample walks through every other field it sets, and the docs carry a whole "Why the samples look the way they do" table, so inheriting a mutating webhook silently reads as an omission rather than a decision.

Could we set `validationConfigurations` explicitly with a line on why, even if the values land on the defaults?

config/samples/policy/kyverno-verify-images.yaml
```diff
@@ -0,0 +1,144 @@
...
+        resources: ["deployments", "replicasets", "daemonsets", "statefulsets"]
+      - apiGroups: ["batch"]
+        apiVersions: ["v1"]
+        operations: ["CREATE", "UPDATE"]
+        resources: ["jobs", "cronjobs"]
+  matchImageReferences:
+    # Narrow globs on purpose. A trailing `manager*` can match manager-debug,
+    # managerial-tool, or manager/nested/extra depending on the glob library.
+    # Tag and digest forms are listed explicitly instead.
+    - glob: "ghcr.io/nvidia/cluster-readiness-engine/manager"
```
ndipebot, line 123: These globs are hard-pinned to `ghcr.io`, but `docs/operations/deployment.md:235-236` tells air-gapped operators to mirror the controller image and point `manager.image.repository` at their own registry. On that path none of the three globs match.

That is worse than it first sounds, because Kyverno fails open here rather than closed. `images.containers` is pre-filtered to images that matched `matchImageReferences` and is initialised to `[]` (`pkg/image/verification/evaluator/policy.go:162-174`), so both validations below evaluate `[].all(e, e > 0)`, which is `true`. The pod is admitted with no signature check, no provenance check, and no error anywhere for the operator to notice.

Neither the `BEFORE YOU APPLY` block above nor the runbook in `verifying-artifacts.md` says to re-point `matchImageReferences` when you mirror. One line in each would cover it. As written this is the same false assurance the page declines to publish a Flux manifest over.

Same applies to `policy-controller-verify-images.yaml:53-55`.

config/samples/policy/kyverno-verify-images.yaml
```diff
@@ -0,0 +1,144 @@
...
+    - glob: "ghcr.io/nvidia/cluster-readiness-engine/manager:*"
+    - glob: "ghcr.io/nvidia/cluster-readiness-engine/manager@*"
+  attestors:
+    - name: nvcreRelease
+      cosign:
+        keyless:
+          identities:
+            # EDIT THIS TAG to the release you run before applying.
+            # Exact identity — never a prefix regexp that also accepts
+            # publish.yml@refs/heads/main development images.
```
ndipebot: `publish.yml@refs/heads/main` is not the identity on our development images any more. `publish.yml` routes all signing through `attest.yml` now, so a `main` build carries `attest.yml@refs/heads/main` and differs from a release only in the ref.

The policy is still correct either way, since a `subjectRegExp` anchored on `attest.yml@refs/tags/` rejects a `refs/heads/main` SAN regardless. It is just that this comment, the one on line 118, and the matching line in `verifying-artifacts.md` all name a workflow that no longer signs anything, which makes the guidance harder to act on than it needs to be.

config/samples/policy/kyverno-verify-images.yaml
```diff
@@ -0,0 +1,144 @@
...
+  attestations:
+    # CamelCase name so CEL can use attestations.slsaProvenance (hyphens are not
+    # valid field selectors).
+    - name: slsaProvenance
+      intoto:
+        # Same predicate the docs verify with --type slsaprovenance1.
+        type: "https://slsa.dev/provenance/v1"
+  validations:
+    - expression: >-
+        images.containers.map(image,
```
ndipebot: On CodeRabbit's comment here. The coverage gap is real: `pkg/cel/compiler/images.go` defines `containers`, `initContainers` and `ephemeralContainers` as three separate keys, and all three are populated for every GVR we match. But I do not think "Authorization Bypass" is the right reading, and I would rather have that on the record before we act on it.

`validationConfigurations.required` defaults to `true`, and `EnforceRequired` runs over every matched image including the init containers, failing with "no policy performed a signature or attestation check on it". So an unsigned manager image sitting in `initContainers` is denied today, not admitted. What we actually have is a confusing denial message, not a hole an attacker can walk through.

Worth applying the concatenation anyway for clarity. If we do, note the suggested fix is incomplete on its own terms: `kubectl debug` adds ephemeral containers through the `pods/ephemeralcontainers` subresource, which is not in `matchConstraints.resourceRules` on line 85, so the ephemeral half stays unreachable until that is added too.

config/samples/policy/policy-controller-verify-images.yaml
```diff
@@ -0,0 +1,82 @@
...
+# plain-signature path. The sample therefore requires:
+#   - https://sigstore.dev/cosign/sign/v1  (the signed identity, bundle form)
+#   - https://slsa.dev/provenance/v1      (the provenance the docs also check)
+# See sigstore/policy-controller#1899 for the bundle + cosign/sign/v1 shape.
+#
+# NAMESPACE SCOPING
+# -----------------
+# policy-controller is opt-in by namespace label (default chart behaviour):
+#   kubectl label namespace nvcre policy.sigstore.dev/include=true
+# Unlabeled namespaces are not evaluated, so a Fail / enforce mode here does
```
ndipebot: This is the opposite of what the label does at chart defaults.

`pkg/config/store.go:74` reads `NoMatchPolicy: "deny"`, and the `switch` immediately below it maps the `default:` branch to `DenyAll` as well. So labelling `nvcre` with `policy.sigstore.dev/include=true` puts the whole namespace into deny-by-default for every image, not only ours. Anything that matches no `ClusterImagePolicy` is rejected.

"Unlabeled namespaces are not evaluated" is true. "A Fail / enforce mode here does not strand workload creation cluster-wide" is not, unless the operator has separately set `no-match-policy` to `allow` or `warn`. Could we say that outright? It is the difference between an opt-in narrowing and an opt-in that denies everything else in the namespace, and an operator who reads this comment and then labels their controller namespace will be surprised.

config/samples/policy/policy-controller-verify-images.yaml
```diff
@@ -0,0 +1,82 @@
...
+kind: ClusterImagePolicy
+metadata:
+  name: nvcre-manager-verify-images
+  labels:
+    app.kubernetes.io/name: cluster-readiness-engine
+    app.kubernetes.io/component: admission-sample
+spec:
+  mode: enforce
+  images:
+    # golang filepath globs; `**` matches across `/`. Do not use a bare
```
ndipebot: Two small things on this comment.

policy-controller does not use `filepath.Match`. It compiles the glob to a regexp itself in `pkg/apis/glob/glob.go`, rewriting `.` to `\.`, `**` to `.*` and `*` to `[^/]*`. So `*` cannot cross `/` on this engine and the nested-path half of the warning does not apply. Naming the mechanism it does use would be clearer than "golang filepath globs".

Also the bare `manager` glob on line 53 can never match: `Match` runs the compiled pattern against `ref.Name()`, which is always fully qualified with a tag or a digest. Harmless, but the docs table counts it as coverage and the structural test would accept it as the only glob. Worth knowing it differs from the Kyverno sample, which uses `gobwas/glob` against the raw pod-spec string, where the bare form does work for an untagged image.

ndipebot (commented):
Went through this against the Kyverno v1.19.0 and policy-controller v0.15.1 sources rather than against their docs, since most of what these samples assert is upstream behaviour we cannot exercise in CI.

The good news first, so nobody re-litigates it. The identity contract is right. `apiVersion: policies.kyverno.io/v1` really is served by v1.19.0 alongside `v1alpha1` and `v1beta1`. Every field the Kyverno sample sets exists in that schema, including top-level `failurePolicy`. `ctlog.url` really is the Rekor URL in both projects despite the field name. `signatureFormat` with values `legacy` and `bundle` goes back to v0.13.1. Kyverno's `ImageValidatingPolicy` cosign verifier does auto-detect our cosign v3 bundles (`buildCheckOptsWithBundleDetection`, `cosign.GetBundles`) and does only accept `intoto` referrers as attestations, which is exactly the shape we use. And `v0.2.0` is the current release, so the pin is not stale.

Four things I would like fixed before this lands. All of them are in docs or sample comments, none in the policy logic itself.

1. Both `Denied` cases in the Kind repro use tags we have never published, so neither one actually tests identity matching.
2. The globs are hard-pinned to `ghcr.io`, and Kyverno fails open on a glob that matches nothing, so our own documented mirror and air-gap install path turns both samples into a silent no-op.
3. For policy-controller, the namespace label is deny-by-default rather than the narrowing the sample describes, and the repro's `Untouched` line is false on that engine.
4. `validationConfigurations` defaults mean applying the Kyverno sample also registers a mutating webhook, which nothing mentions.

I have also left a note on CodeRabbit's open `images.containers` comment. The coverage gap is real, but I think the severity is overstated and the suggested fix is incomplete. Reasoning inline.

Separately on scope. #272 was rescoped to one Kyverno sample and asked for policy-controller to be re-tested before being ruled out. As far as I can tell that re-test has not happened for this shape, and the issue's three acceptance criteria are still unmet, since the current sample adds a mandatory provenance check and a namespace selector that the earlier Kind-tested shape did not have. Not necessarily a blocker if we hedge the table row, but I would rather it were a decision than a default.

Thanks for picking this back up. The parked-failure-mode comments in both samples are genuinely good and I would keep that style.

config/samples/policy/policy-controller-verify-images.yaml
```diff
@@ -0,0 +1,82 @@
...
+            subject: "https://github.com/NVIDIA/cluster-readiness-engine/.github/workflows/attest.yml@refs/tags/v0.2.0"
+            # Loosening: subject is a STRICT match. A pattern under subject
+            # such as refs/tags/.+ matches no SAN. Use subjectRegExp instead:
+            #   subjectRegExp: "^https://github\\.com/NVIDIA/cluster-readiness-engine/\\.github/workflows/attest\\.yml@refs/tags/.+$"
+            # Still reject refs/heads/main and any identity that omits attest.yml.
+      # Required for NVCRE's referrer / new-bundle layout. Without this field
+      # the authority looks for legacy .sig tags we do not publish on current
+      # release digests.
+      signatureFormat: bundle
+      ctlog:
```
ndipebot: `ctlog.url` is dead config once `signatureFormat: bundle` is set on the authority. In bundle mode policy-controller takes its verification material from the TUF trusted root and only compares `ctlog.trustRootRef` against `keyless.trustRootRef`; the URL is never read.

Since we are naming the public default anyway, dropping lines 72 and 73 changes no behaviour and removes something that reads like a transparency-log pin but is not one. If we would rather keep it for symmetry with the Kyverno sample, where `ctlog.url` genuinely is read, a comment saying so would help.

docs/operations/verifying-artifacts.md
```diff
@@ -374,8 +374,110 @@ using the `cosign` provider, matching the same OIDC issuer and identity used abo
...
+pin the same issuer and `attest.yml@refs/tags/<TAG>` identity as the commands above, and
+they expect the **new-bundle / OCI referrer** layout ADR-074 publishes (not a legacy
+`sha256-<digest>.sig` tag).
+
+| Verifier | Reads our format? | Sample |
+|---|---|---|
+| `cosign verify` / `verify-attestation` (v3.1.3) | Yes | this page |
+| Kyverno `ImageValidatingPolicy` (v1.19+) | Yes — referrer / bundle path | `config/samples/policy/kyverno-verify-images.yaml` |
+| Kyverno `ClusterPolicy` / `verifyImages` | **No** — looks for legacy `.sig` tags | do not use |
+| Sigstore policy-controller with `signatureFormat: bundle` | Yes — attestations in bundle form | `config/samples/policy/policy-controller-verify-images.yaml` |
```
ndipebot: I would hedge this row.

#272 was rescoped to one Kyverno sample and said policy-controller "should be re-tested before being ruled out", and as far as I can tell that re-test has not happened for the shape in this PR. An unqualified "Yes" here sits three paragraphs after we decline to publish a Flux manifest because "an untested verification config is exactly the kind of false assurance the rest of this page exists to avoid."

Either mark it the way we marked the Flux row, or split the policy-controller sample into its own PR once someone has actually run it. I do not mind which, but I do not think it should ship as a flat Yes.

docs/operations/verifying-artifacts.md
```diff
@@ -374,8 +374,110 @@ using the `cosign` provider, matching the same OIDC issuer and identity used abo
...
+
+# Edit the sample's subject pin to ${TAG}, then:
+kubectl apply -f config/samples/policy/kyverno-verify-images.yaml
+
+# Admitted: signed release index by digest
+kubectl run nvcre-admit --image="${IMAGE}@${DIGEST}" -n nvcre \
+  --dry-run=server --restart=Never -o name
+
+# Denied: signed under publish.yml@refs/heads/main (development image)
+kubectl run nvcre-deny-main --image="${IMAGE}:main" -n nvcre \
```
ndipebot: This does not test what the comment says it tests.

`publish.yml` only ever pushes `main-<sha7>` and asserts that shape (`.github/workflows/publish.yml:53-58`), and since the tag trigger was removed (`:12`) it publishes no floating `main` tag at all. So `${IMAGE}:main` is a 404, and it gets denied because the manifest cannot be resolved, not because the signing identity is `refs/heads/main`.

`${IMAGE}:not-a-release` on line 469 has the same problem. Both negative cases are denied for the same trivial reason, so neither one exercises the `subject:` pin, which is the entire point of these samples. This is also the only evidence we have for #272's "rejects an unsigned image and an image signed by a different identity" criterion, so as written that box still cannot be ticked.

Swapping in a real `main-<sha7>` tag would turn the first case into a genuine identity test, and it is a one-word change.

docs/operations/verifying-artifacts.md
```diff
@@ -374,8 +374,110 @@ using the `cosign` provider, matching the same OIDC issuer and identity used abo
...
+  --dry-run=server --restart=Never -o name
+
+# Denied: unsigned tag under our repository (not a real release digest)
+kubectl run nvcre-deny-unsigned --image="${IMAGE}:not-a-release" -n nvcre \
+  --dry-run=server --restart=Never -o name
+
+# Untouched: image outside the glob, even in the labeled namespace
```
ndipebot: True for Kyverno, false for policy-controller.

`no-match-policy` defaults to `deny` (`pkg/config/store.go:74`), so once line 455 applies `policy.sigstore.dev/include=true` this `busybox` create is rejected on a policy-controller cluster rather than passing untouched.

Line 451 offers policy-controller as a supported alternative for this whole block and line 455 applies both labels unconditionally, so an operator following the repro on that engine gets a different result from the documented one, on the step that is meant to reassure them about blast radius. Worth splitting the two engines through this block, or at minimum caveating this line.

RELEASE.md
```diff
@@ -51,6 +51,12 @@ people do not tag at the same time.
    made visible only by the `Verify release` job, after it has verified every published
    artifact. If that job fails, the release stays a draft — see Troubleshooting.
 5. Check the published release, then announce it.
+6. If you maintain the admission-policy samples, bump the pinned `subject:` tag in
```
ndipebot: Two things about step 6.

It sits after "Check the published release, then announce it", so by the time anyone reaches it the tag is already cut and the sample committed at that tag names the previous release. Someone who checks out `v0.3.0` and applies the sample as-is gets a policy pinned to `v0.2.0`, which is precisely the stale-pin outage this step exists to warn about.

And "If you maintain the admission-policy samples" makes it optional in a checklist where nothing else is. If we cannot automate the bump, I would rather see this as an unconditional step before the tag push, or as a release-blocking item, than as a conditional one after the announcement.

test/docspolicy/policy_samples_test.go
```diff
@@ -0,0 +1,363 @@
...
+	t.Run("kyverno", func(t *testing.T) {
+		doc := mustLoadYAML(t, kyvernoSample)
+
+		if got := asString(doc["kind"]); got != "ImageValidatingPolicy" {
+			t.Fatalf("kind = %q, want ImageValidatingPolicy — ClusterPolicy/verifyImages "+
+				"cannot see new-bundle referrer signatures", got)
+		}
+
+		spec := asMap(t, doc["spec"], "spec")
+		if got := asString(spec["failurePolicy"]); got != "Fail" {
```
ndipebot, line 48: `failurePolicy` is webhook-failure behaviour, not the enforcement switch. The field that decides whether a failed validation denies or merely audits is `spec.validationActions`, and nothing in this test reads it.

Someone could flip the sample to `validationActions: [Audit]` and every assertion in this subtest still passes while the policy quietly stops denying anything. Given the comment at the top of the file says the job is to stop the samples rotting back into the parked shape, that one seems worth pinning.

test/docspolicy/policy_samples_test.go
```diff
@@ -0,0 +1,363 @@
...
+				globs = append(globs, g)
+			}
+		}
+		assertNarrowManagerGlobs(t, globs)
+
+		authorities := asSlice(t, spec["authorities"], "spec.authorities")
+		if len(authorities) == 0 {
+			t.Fatal("spec.authorities is empty")
+		}
+		auth := asMap(t, authorities[0], "authorities[0]")
```
ndipebot: This reads `authorities[0]` and nothing asserts how many authorities there are.

policy-controller admits an image if any authority verifies it, so appending a second, looser authority would neuter the policy while `signatureFormat`, the keyless identities and both attestations all still check out against index 0. A `len(authorities) == 1` assertion, or a loop over all of them, would close it.

test/docspolicy/policy_samples_test.go
```diff
@@ -0,0 +1,363 @@
...
+			if strings.Contains(subject, ".+") || strings.Contains(subject, ".*") {
+				t.Errorf("subject = %q looks like a regexp; use subjectRegExp: for patterns "+
+					"(a regexp under subject: matches no SAN)", subject)
+			}
+			found = true
+		}
+		if subjectRE != "" {
+			// Optional loosening is fine in comments; a live subjectRegExp must
+			// still name attest.yml and refs/tags.
+			if !strings.Contains(subjectRE, "attest\\.yml") && !strings.Contains(subjectRE, "attest.yml") {
```
ndipebot: This branch only substring-checks for `attest.yml` and `refs/tags`. It never requires the `NVIDIA/cluster-readiness-engine` path and never requires anchoring, so an unanchored `subjectRegExp: "attest\\.yml@refs/tags/"`, which would accept any fork's or any attacker's `attest.yml`, satisfies it and sets `found`.

Related, and in the same helper: `found` is set by any identity in the list, so appending a second wide-open identity alongside the good one also passes. Both are cheap to tighten, and the second one matters more, since adding an identity is a much likelier edit than replacing one.

config/samples/policy/kyverno-verify-images.yaml
asivanadi0: Fixed in 5aa4e01. `validationConfigurations` is now set explicitly (`mutateDigest` / `verifyDigest` / `required: true`) with a comment that applying the sample also registers the mutating webhook. Defaults unchanged; documented as a decision. Also reflected in the "Why the samples look the way they do" table.

config/samples/policy/kyverno-verify-images.yaml
asivanadi0: Fixed in 5aa4e01. BEFORE YOU APPLY #4 on the Kyverno sample, BEFORE YOU APPLY #4 on the policy-controller sample, and a new apply-runbook step in verifying-artifacts.md all say to rewrite the `ghcr.io` globs on mirror/air-gap installs, and call out Kyverno's empty-list fail-open vs policy-controller's no-match deny.

config/samples/policy/kyverno-verify-images.yaml
asivanadi0: Fixed in 5aa4e01. Comments now name `attest.yml@refs/heads/main` (publish.yml routes signing through attest.yml). Matching Kind-repro comment in verifying-artifacts.md updated the same way.

config/samples/policy/kyverno-verify-images.yaml
asivanadi0: Fixed in 5aa4e01. Applied the containers+init+ephemeral concatenation for clearer deny messages, and added `pods/ephemeralcontainers` to `matchConstraints.resourceRules` so the ephemeral half is reachable. Agreed this was not an authorization bypass — `required: true` already denied unchecked init images; noted that in the sample comment and the docs table. Reply on the CodeRabbit thread points here too.

config/samples/policy/kyverno-verify-images.yaml
```diff
@@ -0,0 +1,144 @@
...
+          verifyImageSignatures(image, [attestors.nvcreRelease])).all(e, e > 0)
+      message: >-
+        NVCRE manager image signature verification failed. Expected a new-bundle
+        cosign signature under attest.yml@refs/tags/<pinned-tag>. Legacy
+        ClusterPolicy/verifyImages cannot see these signatures.
+    - expression: >-
+        images.containers.map(image,
+          verifyAttestationSignatures(image, attestations.slsaProvenance,
+            [attestors.nvcreRelease])).all(e, e > 0)
```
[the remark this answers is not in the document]
asivanadi0: Addressed with ndipebot's contextualization in 5aa4e01: concatenated containers/init/ephemeral as suggested, and also added `pods/ephemeralcontainers` to matchConstraints (required for the ephemeral half — the suggested diff alone left that unreachable). Severity: not an authorization bypass; `validationConfigurations.required` already denied unchecked init images (confusing message, not a hole).

config/samples/policy/policy-controller-verify-images.yaml
asivanadi0: Fixed in 5aa4e01. NAMESPACE SCOPING now states that labelling puts the namespace under chart-default `no-match-policy: deny` for every unmatched image — opt-in that denies everything else in the namespace unless `no-match-policy` is `allow`|`warn`. Docs table + Kind repro caveat updated the same way.

config/samples/policy/policy-controller-verify-images.yaml
asivanadi0: Fixed in 5aa4e01. Comment now names `pkg/apis/glob/glob.go` (`*` → `[^/]*`, no filepath.Match). Dropped the bare `.../manager` glob because `Match` runs against `ref.Name()` which always has a tag or digest. Kept `manager:*` and `manager@sha256:*`.

config/samples/policy/policy-controller-verify-images.yaml
asivanadi0: Fixed in 5aa4e01. Dropped `ctlog.url` on the policy-controller sample with a comment that bundle mode takes CT material from the TUF trusted root (Kyverno's `ctlog.url` stays — it is actually read there).

docs/operations/verifying-artifacts.md
asivanadi0: Fixed in 5aa4e01. Hedged the policy-controller table row the same way as Flux: schema/format yes, but this sample shape is unconfirmed end-to-end (issue #272 asked for a re-test before ruling it in). Keeping the sample as format guidance rather than splitting it out of this PR.

docs/operations/verifying-artifacts.md
asivanadi0: Fixed in 5aa4e01. Kind deny-main now uses a real published `main-c2d4d47` tag (`DEV_TAG`) so admission resolves the manifest and exercises the `subject:` identity pin; comment notes `publish.yml` does not publish a floating `:main`. The `:not-a-release` case is reframed honestly as the unresolvable / no-signature-material stand-in — we do not publish unsigned manager images under this path.

docs/operations/verifying-artifacts.md
asivanadi0: Fixed in 5aa4e01. Kind repro is now Kyverno-primary (label only the Kyverno opt-in). Added an explicit policy-controller caveat that chart-default `no-match-policy=deny` means the busybox create is DENIED (not untouched) once `policy.sigstore.dev/include=true` is applied.

RELEASE.md
asivanadi0: Fixed in 5aa4e01. Sample `subject:` bump is now unconditional step 3 *before* tagging; tag/push is step 4 on the commit that already carries the bumped pins. Removes the post-announce optional framing.

test/docspolicy/policy_samples_test.go
asivanadi0: Fixed in 5aa4e01. Structural test now requires `validationActions: [Deny]` (length 1) in addition to `failurePolicy: Fail`.

test/docspolicy/policy_samples_test.go
asivanadi0: Fixed in 5aa4e01. Test requires `len(authorities) == 1` and loops every authority for `signatureFormat`, identities, and attestations — a second looser authority can no longer sneak through.

test/docspolicy/policy_samples_test.go
asivanadi0: Fixed in 5aa4e01. `assertExactReleaseIdentity` now requires *every* identity to be tight (not "any one found"), and a live `subjectRegExp` must name `NVIDIA/cluster-readiness-engine`, `attest.yml`, `refs/tags`, and be anchored with `^...$`.

asivanadi0:
Addressed the open ndipebot review (and CodeRabbit `images.containers` with ndipebot's contextualization) in **5aa4e01**.

### Must-fix themes
1. **Kind Denied identity case** — uses real published `main-c2d4d47` (`DEV_TAG`); floating `:main` is not published. `:not-a-release` reframed as unresolvable / no-signature-material stand-in (we do not ship unsigned manager images).
2. **Mirror / air-gap globs** — BEFORE YOU APPLY + apply runbook tell operators to rewrite `ghcr.io` globs; documents Kyverno empty-list fail-open vs policy-controller no-match deny.
3. **policy-controller namespace label** — docs now say labelling is deny-by-default for unmatched images (`no-match-policy: deny`); Kind repro is Kyverno-primary with an explicit PC Untouched caveat.
4. **`validationConfigurations`** — set explicitly (`mutateDigest` / `verifyDigest` / `required: true`) and documented as also registering a mutating webhook.

### Also fixed
- CodeRabbit init/ephemeral: concatenation **and** `pods/ephemeralcontainers` in `matchConstraints` (suggested diff alone left ephemeral unreachable); not treated as authz bypass.
- `attest.yml@refs/heads/main` wording (not `publish.yml`).
- PC glob comment (`pkg/apis/glob`); dropped dead bare `.../manager` glob; dropped unused bundle-mode `ctlog.url`.
- Hedged policy-controller table row (format yes / e2e unconfirmed for this shape per #272).
- RELEASE.md: unconditional sample pin bump **before** tag.
- Structural tests: `validationActions: [Deny]`, `len(authorities)==1` + loop, every identity tight + anchored `subjectRegExp` with repo path.

`make lint` 0 issues; `go test ./test/docspolicy/...` and `make test` green. Docs Verify ```shell fences for cluster-only kubectl kept.

Fresh `/ok to test` would help re-run CI on the fork push.

All actionable ndipebot (+ CodeRabbit) threads replied + resolved; none deferred.

asivanadi0:
Rebased onto `main` (includes #330). Head `5aa4e01` → `6f8a6ee`. Kept ```shell``` fences for cluster-only kubectl samples and the ndipebot review fixes. `make lint`, `go test ./test/docspolicy/...`, and `make test` green locally. Fresh trustee `/ok to test` needed for NVIDIA runners on the new head.

config/samples/policy/policy-controller-verify-images.yaml
ndipebot: Correcting myself here, because you took my wording into the file and my wording was wrong.

I said the bare `.../manager` glob "can never match an image". That is too strong. `glob.Match` has a second chance I missed: when `re.MatchString(ref.Name())` fails and `ref.Name() != image`, it retries the pattern against the raw string. I ran the real logic against `go-containerregistry@v0.20.6` rather than reasoning about it a second time:

```
image .../manager   glob .../manager    -> MATCH via the raw fallback (ref.Name()=.../manager:latest)
image .../manager   glob .../manager:*  -> MATCH via ref.Name()
```

So a bare glob does match an untagged pod-spec reference.

Dropping it was still the right call and I would not put it back. `manager:*` already covers the untagged form through `:latest` normalisation, and I checked that the two remaining globs cover all four realistic reference shapes including tag-plus-digest. The only thing that needs changing is the justification now sitting at line 68, which says the bare form "never matches and is omitted". Redundant is the accurate word, not inert.

Sorry for the bum steer.

ndipebot (commented):
Second pass over `40011c9..6f8a6ee`. All fourteen of my earlier comments are properly addressed, and I checked the four highest-stakes ones against upstream source rather than taking the summary on trust.

1. The concatenated CEL compiles. Kyverno enables `cel.OptionalTypes()` at `pkg/cel/compiler/env.go:61`, so `images.?containers.orValue([])` against a `map[string][]string` is valid. I also compiled both expressions against the `cel-go` and `k8s.io/apiserver` versions pinned in Kyverno v1.19.0's `go.mod`, since a policy that fails to compile would have been the worst outcome here.
2. `pods/ephemeralcontainers` works. `pkg/cel/policies/ivpol/engine/engine.go:441` passes `request.RequestResource`, which for a subresource request is the parent GVR, so `getImageExtractorsFromGVR` still injects all three extractors.
3. Removing `ctlog` from the policy-controller sample is safe. In bundle mode `IgnoreTlog` only becomes true when an RFC3161 TSA is configured, and there is none here, so the transparency-log check is still required and `TrustedMaterial` comes from the TUF root.
4. `main-c2d4d47` really is published. GHCR returns 200 for it, while `:main` and `:not-a-release` are both 404, so the identity case now exercises the `subject:` pin for real.

Also confirmed: the `deployment.md#air-gapped-and-disconnected-environments` anchor resolves, the `RELEASE.md` renumbering is internally consistent with no stale step references left in the file, and `go vet` plus `go test ./test/docspolicy/...` are clean at this head.

I chased four other things this round and dropped all four, so nobody needs to reopen them: a supposed conflict between the new step 3 and the release-candidate rule, a claim that the repro still misses the unsigned half of #272 criterion 2, a `mutateDigest` persistence concern, and a green-`main` ordering concern. None of them held up.

I have also corrected one of my own round-one comments in a reply on the policy-controller glob thread. The justification I gave you was wrong, though the change you made off it was right.

What is left is one thing I would like fixed, on the Kyverno attestor assertions, plus four nits you can take or leave. Nice work on this round, the sample comments are genuinely useful to read now.

test/docspolicy/policy_samples_test.go
```diff
@@ -0,0 +1,397 @@
...
+	}
+}
+
+func kyvernoIdentities(t *testing.T, spec map[string]any) []any {
+	t.Helper()
+	atts := asSlice(t, spec["attestors"], "spec.attestors")
+	if len(atts) == 0 {
+		t.Fatal("spec.attestors is empty")
+	}
+	attestor := asMap(t, atts[0], "attestors[0]")
```
ndipebot: This is the one thing I would still like fixed before this goes in.

You closed the "a second looser entry neuters the pin" hole on the policy-controller side. `len(authorities) != 1` is fatal now and every identity has to be tight. The identical hole is still open here on the Kyverno side, which is the primary sample.

`kyvernoIdentities` reads `attestors[0]` and returns, nothing bounds `len(spec.attestors)`, and nothing checks which attestor name the CEL expressions actually hand to `verifyImageSignatures`. I reproduced it: append a second attestor with `issuerRegExp: ".*"` and `subjectRegExp: ".*"`, add it to the attestor list in both validations, and any image signed by any Fulcio identity verifies with a count of at least one and is admitted into the enforcing namespace. The suite stays green throughout, which is exactly the rot this file exists to catch.

Mirroring what you already did for policy-controller would cover it: fatal unless there is exactly one attestor, run `assertExactReleaseIdentity` over every attestor's identities, and assert the joined expressions reference `attestors.<name>` for that one attestor and no other `attestors.` selector.

test/docspolicy/policy_samples_test.go
```diff
@@ -0,0 +1,397 @@
...
+				t.Errorf("identities[%d].subjectRegExp = %q must be anchored with ^...$", i, subjectRE)
+			}
+		}
+	}
+}
+
+func resourceRulesInclude(rules []any, resource string) bool {
+	for _, r := range rules {
+		m, _ := r.(map[string]any)
+		if slices.Contains(asStringSlice(m["resources"]), resource) {
```
ndipebot: Nit, take it or leave it.

This only asks whether the string appears in some rule's `resources`. It never checks that the rule carrying it has `apiGroups: [""]`, `apiVersions: ["v1"]`, or the `UPDATE` operation that an ephemeral-container patch actually uses.

So a later tidy-up that relocates `pods/ephemeralcontainers` under the `apps` rule, or narrows that rule to `CREATE` only, leaves this assertion green while the API server never routes the subresource to the policy. Passing the expected group and required operations into the helper would close it, and asserting plain `pods` is in the same rule would be worth doing at the same time.

test/docspolicy/policy_samples_test.go
```diff
@@ -0,0 +1,397 @@
...
+			t.Error("validations must call verifyImageSignatures")
+		}
+		if !strings.Contains(joined, "verifyAttestationSignatures") {
+			t.Error("validations must call verifyAttestationSignatures for provenance")
+		}
+		if !strings.Contains(joined, "slsaProvenance") {
+			t.Error("validations must reference attestations.slsaProvenance")
+		}
+		for _, key := range []string{"initContainers", "ephemeralContainers"} {
+			if !strings.Contains(joined, key) {
```
ndipebot: Same class of nit, and cheap to fix while you are in here.

`joined` is every `spec.validations[].expression` concatenated, so one expression mentioning the keys satisfies this check for all of them. The provenance validation could quietly drop back to `images.containers` while the signature validation keeps the concatenation, and this still passes.

Checking each expression separately would match what the comment on the sample claims the coverage is.

config/samples/policy/kyverno-verify-images.yaml
```diff
@@ -0,0 +1,177 @@
...
+    - name: slsaProvenance
+      intoto:
+        # Same predicate the docs verify with --type slsaprovenance1.
+        type: "https://slsa.dev/provenance/v1"
+  validations:
+    # Concatenate containers + initContainers + ephemeralContainers. The three
+    # are distinct keys on images (pkg/cel/compiler/images.go); containers alone
+    # leaves init/ephemeral to validationConfigurations.required (deny with a
+    # "no policy performed a check" message). Explicit coverage keeps the deny
+    # message accurate. ephemeralContainers only appears when
```
ndipebot: Small accuracy thing in an otherwise good comment.

`ephemeralContainers` is not "only" present when the subresource is in `matchConstraints`. The key is always populated for the built-in GVRs and initialised to `[]`, and it can be non-empty on a plain `pods` UPDATE without the subresource rule.

What the subresource rule actually buys is interception: without it the API server never routes the `kubectl debug` request to this policy in the first place. Worth rewording, because the current phrasing would send someone looking in the wrong place if ephemeral coverage ever seemed to misbehave.

docs/operations/verifying-artifacts.md
```diff
@@ -374,8 +374,138 @@ using the `cosign` provider, matching the same OIDC issuer and identity used abo
...
+is server-side dry-run so the admission webhook actually runs:
+
+```shell
+TAG=v0.2.0
+IMAGE=ghcr.io/nvidia/cluster-readiness-engine/manager
+DIGEST="$(crane digest "${IMAGE}:${TAG}")"
+ID="https://github.com/NVIDIA/cluster-readiness-engine/.github/workflows/attest.yml@refs/tags/${TAG}"
+# Live main-<sha7> tag from GHCR (publish.yml does not publish a floating :main).
+# Replace with any current main-<sha7> if this example ages out.
+DEV_TAG=main-c2d4d47
```
ndipebot, line 464: This is the right fix and I am glad the case is real now. I checked and `main-c2d4d47` does resolve, GHCR returns 200 for it.

One thing though. A frozen dev SHA in published docs will drift, and the comment above tells the reader to "Replace with any current `main-<sha7>`" without showing them how to find one. When this tag stops resolving, the case degenerates into the same manifest-lookup denial as the `:not-a-release` case below it, which is the failure mode we just spent a round removing.

A one-line snippet that lists a current dev tag would make it self-healing rather than dated.

test/docspolicy/policy_samples_test.go
asivanadi0: Fixed in 99615d5. `assertKyvernoSingleTightAttestor` now fatals unless `len(attestors)==1`, runs `assertExactReleaseIdentity` over that attestor's identities, and `assertValidationsReferenceOnlyAttestor` requires every CEL `attestors.<name>` selector to be that one name — same class of pin as policy-controller's `len(authorities)!=1`.

test/docspolicy/policy_samples_test.go
asivanadi0: Fixed in 99615d5. Replaced the resource-substring helper with `resourceRuleCoversEphemeralContainers`: requires `pods/ephemeralcontainers` and `pods` on the same `\"\"/v1` rule with `UPDATE` in operations.

test/docspolicy/policy_samples_test.go
asivanadi0: Fixed in 99615d5. `initContainers` / `ephemeralContainers` are now asserted on each `validations[].expression` separately, not across the joined set.

config/samples/policy/kyverno-verify-images.yaml
asivanadi0: Fixed in 99615d5. Reworded both comments: the subresource rule is about API-server interception of `kubectl debug`; `images.ephemeralContainers` is always populated (as `[]` when empty) on built-in GVRs.

docs/operations/verifying-artifacts.md
asivanadi0: Fixed in 99615d5. Kind repro now shows how to list a current tag: `crane ls …/manager | grep '^main-' | head -n1` above the frozen `DEV_TAG=main-c2d4d47` example.

asivanadi0:
Pushed `99615d5` addressing the 2nd-pass ndipebot threads (Kyverno single-attestor pin + the four nits). SHA moved — fresh `/ok to test` when convenient.

docs/operations/verifying-artifacts.md
```diff
@@ -374,8 +374,139 @@ using the `cosign` provider, matching the same OIDC issuer and identity used abo
...
+is server-side dry-run so the admission webhook actually runs:
+
+```shell
+TAG=v0.2.0
+IMAGE=ghcr.io/nvidia/cluster-readiness-engine/manager
+DIGEST="$(crane digest "${IMAGE}:${TAG}")"
+ID="https://github.com/NVIDIA/cluster-readiness-engine/.github/workflows/attest.yml@refs/tags/${TAG}"
+# Live main-<sha7> tag from GHCR (publish.yml does not publish a floating :main).
+# List a current one when this example ages out:
+#   crane ls ghcr.io/nvidia/cluster-readiness-engine/manager | grep '^main-' | head -n1
```
ndipebot, line 463: This picks the oldest dev tag, not a current one, and the oldest ones are not signed.

GHCR returns tags in push order, oldest first. There are 165 `main-<sha7>` tags on the package, so `head -n1` gives `main-908cb14`, which is the initial commit from 2026-08-06.

I checked signature material on each by looking for the `sha256-<hex>` referrer fallback tag, since GHCR does not serve the Referrers API. Bisecting that list, the first signed one is `main-7c14f5b` at index 89, so the first 89 of 165 have nothing: no referrer tag, no `.sig`.

Which means a reader whose `DEV_TAG` has aged out runs this snippet, gets an unsigned image, and the deny lands on missing signature material rather than on the subject pin. That is the same degenerate case as `:not-a-release`, which is what we just spent a round removing.

`tail -n1` gives `main-76be643`, which is signed and current, so that alone would fix it. I would rather skip the registry though, since `publish.yml:53` already computes the tag deterministically as `value="main-${GITHUB_SHA::7}"`:

```
# List a current one when this example ages out:
#   DEV_TAG="main-$(git rev-parse --short=7 origin/main)"
```

No network, and it cannot drift.

docs/operations/verifying-artifacts.md
```diff
@@ -374,8 +374,139 @@ using the `cosign` provider, matching the same OIDC issuer and identity used abo
...
+kubectl run nvcre-admit --image="${IMAGE}@${DIGEST}" -n nvcre \
+  --dry-run=server --restart=Never -o name
+
+# Denied: signed under attest.yml@refs/heads/main (development image).
+# Uses a real published main-<sha7> tag so admission resolves the manifest and
+# exercises the subject: identity pin (a floating :main tag is not published).
+kubectl run nvcre-deny-main --image="${IMAGE}:${DEV_TAG}" -n nvcre \
+  --dry-run=server --restart=Never -o name
+
+# Denied: unpublished tag under our repository. We do not ship unsigned manager
```
ndipebot, line 484: This is not true, and it is carrying more weight than a comment should.

We do ship unsigned manager images. `:v0.1.0-rc.1` through `:v0.1.0-rc.8` are all published, all return 200, and none of them have signature material: no `sha256-<hex>` referrer tag, no `.sig`. Signing starts at `:v0.1.0-rc.9`.

The reason it matters is the line below it. `:not-a-release` is a 404, so that case exercises manifest resolution rather than signature absence, and this sentence is the stated justification for accepting that. #272 asks for a case that rejects an unsigned image, and a lookup failure is not one. It is the same objection I raised on the `:main` case last round, one layer down.

Swapping the image makes the criterion real and retires the sentence:

```
# Denied: published but unsigned. Pre-dates release signing, which starts at
# v0.1.0-rc.9, so admission resolves the manifest and fails on missing
# signature material rather than on the lookup.
kubectl run nvcre-deny-unsigned --image="${IMAGE}:v0.1.0-rc.1" -n nvcre \
  --dry-run=server --restart=Never -o name
```

These are old tags and could be garbage collected eventually, but that is a better failure than the one we have, and it is obvious when it happens.

test/docspolicy/policy_samples_test.go
```diff
@@ -0,0 +1,451 @@
...
+		}
+		assertValidationsReferenceOnlyAttestor(t, joined, attestorName)
+		// Per expression, not across the joined set: one validation covering
+		// init/ephemeral must not paper over another that dropped back to
+		// images.containers alone.
+		for i, raw := range vals {
+			expr := asString(asMap(t, raw, "validations[]")["expression"])
+			for _, key := range []string{"initContainers", "ephemeralContainers"} {
+				if !strings.Contains(expr, key) {
+					t.Errorf("validations[%d] must include images.%s (containers alone is incomplete)", i, key)
```
ndipebot, line 93: Nit, take it or leave it.

With the attestor list and the image lists both pinned now, the result predicate is the last unpinned piece of these expressions. I changed `.all(e, e > 0)` to `.all(e, e >= 0)` in both validations and the suite stayed green. At that point both validations are vacuously true for an image with zero verified signatures and zero verified provenance, which is the fail-open shape this file exists to catch, off a one-character edit.

The loop you already have here is the natural place for it. Requiring each expression to match something like `all\(\s*\w+\s*,\s*\w+\s*>\s*0\s*\)` would catch `>= 0`, a switch to `exists`, and a dropped `.all(...)` in one go.

test/docspolicy/policy_samples_test.go
```diff
@@ -0,0 +1,451 @@
...
+			}
+		}
+	}
+}
+
+// resourceRuleCoversEphemeralContainers requires the subresource on the core
+// v1 Pod rule with UPDATE (the operation kubectl debug uses), and pods in the
+// same rule so a tidy-up that relocates the subresource under apps cannot keep
+// the assertion green while the API server never routes it to the policy.
+func resourceRuleCoversEphemeralContainers(rules []any) bool {
```
ndipebot, line 384: Two small things about this helper, both take-or-leave.

It returns on the first rule carrying the subresource rather than looking for one that satisfies everything. The loop `continue`s only when `pods/ephemeralcontainers` is absent, so a rule that has the subresource but fails a later check hits `return false` and any good rule after it is never reached. I put a broken `apps` rule carrying the subresource ahead of the core rule and the suite failed, even though the core rule still covers it correctly. Turning each `return false` into `continue`, and returning `true` only once a rule satisfies all of it, would match what the API server actually does.

The other one is line 400: it requires `UPDATE` but never `CREATE`. I narrowed the core rule to `operations: ["UPDATE"]` and the suite stayed green, which would leave a directly created Pod unvalidated while this assertion still reads as covering Pod admission. Requiring both, or accepting `["*"]`, closes it.

test/docspolicy/policy_samples_test.go
```diff
@@ -0,0 +1,451 @@
...
+		}
+		cosign := asMap(t, attestor["cosign"], "attestors[].cosign")
+		keyless := asMap(t, cosign["keyless"], "attestors[].cosign.keyless")
+		ids := asSlice(t, keyless["identities"], "keyless.identities")
+		assertExactReleaseIdentity(t, ids)
+	}
+	return name
+}
+
+var attestorSelector = regexp.MustCompile(`attestors\.([A-Za-z_][A-Za-z0-9_]*)`)
```
ndipebot, line 319: Nit, and the smallest of these.

This only sees dotted selectors. I wired a second attestor in as `[attestors.nvcreRelease, attestors["anyFulcio"]]` and the scan did not notice it.

Not exploitable, because getting a second attestor into CEL means defining one, and the `len(atts) != 1` fatal at line 302 fires first. The pin holds either way. It is only that this check reads as though it independently covers the CEL side, and it does not. Widening it to allow the bracket form, or a line noting that the `len == 1` check is the load-bearing one and this is defence in depth, would both do.

ndipebot (commented):
Third pass over `6f8a6ee..99615d5`. All five second-pass items are fixed, and this round I ran my own exploit against the suite rather than reading the diff, since the point of that file is that a green run should mean something.

Confirmed by execution, so nobody needs to re-check these:

1. The attestor hole is closed. My exact round-two exploit, a second attestor with `issuerRegExp: ".*"` and `subjectRegExp: ".*"` wired into both validations, now fails with `spec.attestors has 2 entries, want exactly 1`. The three variants that worried me fail too: replacing the single identity with a wide-open one, appending a second wide-open identity inside the one attestor, and pointing the CEL at `attestors.somethingElse`. A consistent rename to another tight name still passes, so there is no false positive.
2. `resourceRuleCoversEphemeralContainers` does what it claims. Relocating the subresource under the `apps` rule, narrowing the core rule to `CREATE`, and dropping bare `pods` each fail.
3. The per-expression check works and names the right index. Reverting only the provenance expression to `images.containers` fails with `validations[1] must include images.initContainers`.
4. The `ephemeralContainers` rewording is accurate.
5. The deny case is a real identity test now. I decoded the certificate out of the bundle rather than trusting the 200: `main-c2d4d47` carries a SAN of `attest.yml@refs/heads/main`, and `v0.2.0` carries `attest.yml@refs/tags/v0.2.0`. That is the pin we wanted exercised.

Two things I would like fixed before this lands, both on the Kind repro block and both small. `head -n1` picks the oldest dev tag rather than a current one, and the oldest ones carry no signature material, so the snippet quietly recreates the failure mode we spent last round removing. And the line excusing the unsigned case is not true, which matters because it is the reason #272's unsigned criterion is still untested. Details inline, including the image that would make that case real.

I also need to correct my own correction. My round-two reply on the policy-controller glob thread was wrong and your original wording was right, so please do not make that change. Reasoning on that thread. Two bad calls on the same four lines is a poor showing and I am sorry for the churn.

Three nits on the test file after that, all take-or-leave.

config/samples/policy/policy-controller-verify-images.yaml
ndipebot: Retracting this. I was wrong, the original wording was right, please leave lines 66 to 68 as they are.

What I leaned on was the retry branch in `glob.Match`, upstream `pkg/apis/glob/glob.go:92`:

```go
match := re.MatchString(ref.Name())
if !match && ref.Name() != image {
    match = re.MatchString(image)
}
```

I read that as the raw, unqualified string getting a second chance, which would make a bare `.../manager` glob live. It does not, on the admission path. The caller has already normalised:

- `pkg/webhook/validator.go:1159` parses the container image into `ref`
- `pkg/webhook/validator.go:1166` passes `ref.Name()`, not the original string, into `GetMatchingPolicies`
- `pkg/apis/config/image_policies.go:142` hands that same value straight to `glob.Match` as `image`

So by the time `glob.go:87` re-parses, it is parsing a string that is already fully qualified, which is idempotent, and `ref.Name() != image` is false. The retry never fires for anything the webhook sees. It exists for callers that pass a raw string, and the webhook is not one.

Which leaves exactly what you wrote: patterns are only ever matched against the fully qualified form, so a glob with no tag or digest never matches and there is no point listing one. Your wording also names the right mechanism, `ref.Name()`, rather than describing the symptom.

I pinned v0.15.1 for all of this, matching the version the sample targets.

Sorry for the churn, two wrong calls from me on the same four lines.

asivanadi0:
Pushed **fc5bdd3** addressing the 3rd-pass Kind repro items.

1. `DEV_TAG` refresh is now `main-$(git rev-parse --short=7 origin/main)` — GHCR lists oldest-first, so `crane ls | head -n1` was returning unsigned `main-908cb14`.
2. Unsigned deny case is published `:v0.1.0-rc.1` (signing starts at `v0.1.0-rc.9`), not `:not-a-release` (404 / lookup failure).

Also took the three structural-test nits (`all(e > 0)`, CREATE+UPDATE helper, bracket attestor selectors). `go test ./test/docspolicy/...` green locally.

SHA moved — fresh trustee `/ok to test fc5bdd3` when convenient.

asivanadi0:
Rebased onto latest `main` (includes #341). Tip `fc5bdd3` → `6c57982`; clean rebase (RELEASE.md auto-merged sample subject-bump + attest-selftest rollout). Fresh `/ok to test` may be needed on the new SHA.

ndipebot (approved):

Issue #335, opened 2026-09-10
[Bug]: setup init [deps] fails on clusters that already have JobSet installed; setup reset deletes the pre-existing JobSet CRD


ndipebot:
### Prerequisites
- [x] I searched existing issues
- [x] I can reproduce this issue

### Code of Conduct
- [x] I agree to follow the project's Code of Conduct

### Bug Description

`nvcrectl setup init` (v0.2.0) dies in the `[deps]` phase on any cluster that already has JobSet installed as its own Helm release.

NVCRE installs Kubeflow Trainer 2.2.1 via `helm upgrade --install`, and that chart pulls JobSet in as a subchart. `jobset.install` defaults to `true`, and the chart's own values file documents the constraint we are violating:

```yaml
jobset:
  # -- Whether to install jobset as a dependency managed by trainer.
  # This must be set to `false` if jobset controller/webhook has already been installed into the cluster.
  install: true
  # -- String to fully override jobset release name.
  fullnameOverride: jobset
```

`fullnameOverride: jobset` means the subchart's objects take the same names a standalone `jobset` release already owns, so the cluster-scoped ones collide by name and Helm refuses to adopt them into `kubeflow-trainer`.

`installTrainerHelmRelease` (`pkg/setup/helm.go:209-229`) hardcodes the Helm arguments and never sets `jobset.install`. It also passes `jobset.controller.tolerations[0].operator=Exists`, so the install path assumes the subchart is being installed. `setup init` has no `--set` or `--values` passthrough (`pkg/setup/setup.go:114-125`), so the value cannot be reached from the CLI at all.

There are three parts to this:

1. The install always fails. There is no detection of a pre-existing JobSet and no way to set `jobset.install=false`.

2. The failure gets no guidance. `classifyHelmInstallFailure` (`pkg/setup/helm.go:391-403`) requires `conflict`, `.data.`, and `secret` to all appear in the transcript. The `invalid ownership metadata` message contains none of the first two, so it classifies as `failureClassOther` and `installDepsPhase` returns the raw Helm output (`pkg/setup/setup.go:338-339`, `362-365`). Nothing points the user at `--skip-phases=deps`.

[...]

Issue #145, opened 2026-08-12
helm upgrade --install does not update CRDs — schema drift after CRE upgrades


ndipebot:
### Prerequisites

- [x] I searched existing issues
- [x] I can reproduce this issue

### Code of Conduct

- [x] I agree to follow the project's Code of Conduct

### Bug Description
`pkg/setup/helm.go` uses `helm upgrade --install` to deploy CRE. Helm does not update CRDs on upgrade by design (see Helm docs: CRDs are only applied on initial install). When CRE is upgraded to a version with new or modified CRD schemas, the cluster's installed CRDs remain at the old version. Controllers may attempt to write fields that the old CRD schema rejects, or new CRD validation rules are not enforced for resources that were created before the upgrade.

**Location:** `pkg/setup/helm.go:124–137`

**Suggested fix:** Before calling `helm upgrade --install`, apply the CRD manifests directly with `kubectl apply -f helm/cluster-readiness-engine/crds/` (or use `kubectl replace --force` for fields that require replacement). Alternatively, adopt the Helm CRD install hooks pattern or use a pre-upgrade Job.

#

### Component

Helm / Deployment

### Steps to Reproduce
1. Install CRE v1.
2. Upgrade to CRE v2 (with a new CRD field) using `ncrectl install upgrade`.
3. Inspect the installed CRD: `kubectl get crd certifications.cre.nvidia.com -o yaml` — the new field is absent.

#### Environment
- CRE version: all
- Kubernetes version: all
- Deployment method: ncrectl install / Helm

PR #336, opened 2026-09-10
docs: propose JobSet ownership and cleanup design

kaynetu:
## Summary

Propose ADR-078 for issue #335: Trainer installation collides with independently installed JobSet resources, and setup reset can delete the shared JobSet CRD and its workloads.

The design separates controller ownership from CRD provenance. It reuses an identified external controller, preserves bundled upgrades, and restores a bundled controller after reset when inspection confirms a CRD-only state with no JobSets or conflicting resources. A missing CRD short-circuits ownership probing; unknown ownership prints the manual-install plus `--skip-phases=deps` fallback.

Reset and recovery retain JobSet CRDs. Reset reports them without deletion commands; schema upgrades remain operator-managed pending a separate design. Recovery protects unrelated namespace resources while excluding Events and EndpointSlices. ADR-073 explicitly records the proposed changes to its recovery procedure and namespace-cleanup policy.

Require complete Helm collision transcript fixtures and a dedicated setup lifecycle UAT suite in the implementation PR, covering bundled upgrade, reset/init, and external JobSet preservation on an isolated Kind cluster.

ADR-078 remains **Proposed**. This PR contains design documentation only; runtime changes and the new UAT suite will follow separately.

## Related Issue

Related to #335. This design PR does not close the bug.

## Type of Change

- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected

- [ ] API / CRDs
- [ ] Catalog / Workloads
- [ ] CLI (nvcrectl)
- [ ] Helm / Deployment
- [x] Documentation / CI
- [ ] Other: ____________

## Testing

- [x] Manual documentation review completed
- [x] Documentation links validated
- [x] No runtime changes

For review-fix commit `94fbff9`: reviewed the revised decisions and validation requirements, verified the JobSet values excerpt and LWS defaults against the Trainer 2.2.1 chart archive, and ran `make verify-doc-links` (279 relative links) and `git diff --check`. Commit signature and DCO sign-off verified.

The original proposal at `a571aa4` passed CI, including build, lint, tests, and catalog UAT. That suite was not rerun locally for this documentation-only follow-up. The proposed setup lifecycle UAT and diagnostics fixtures are implementation requirements, not tests added or executed by this PR.

## Risk Assessment

Documentation only; no runtime behavior changes. The proposal explicitly accepts deletion of Events and EndpointSlices during namespace recovery, requires access to inspect protected resources, and leaves shared JobSet CRD schema upgrades to operators. These boundaries remain subject to design review before implementation.

## Checklist

- [x] Self-review completed
- [ ] `make manifests generate` run (if `*_types.go` was modified)
- [ ] Golden files updated (if integration test output changed)
- [x] Documentation updated
- [x] Ready for review

CRD generation and golden updates are not applicable to this documentation-only change.


ndipebot (commented):
I went through this against the code on `main`, the published Trainer 2.2.1 chart, and Helm's source. The problem statement is accurate, every code reference in it checks out, and the core mechanism works. But I do not want to accept it as written. Decision 3 combined with decisions 1 and 2 breaks the ordinary `reset` then `init` cycle, and decision 4 has grown into a subsystem that will refuse recovery in practice.

## 1. `setup reset` followed by `setup init` leaves a bundled cluster with no JobSet controller

Decision 3 retains the CRD "in every ownership state, including bundled installs". Decision 1's table then classifies "CRD present, no controller, no release" as `external`, and decision 2 turns `external` into `--set jobset.install=false`. Trace that on an ordinary cluster where NVCRE installed JobSet itself:

1. `setup reset` runs `uninstallDepsPhase` (`pkg/setup/setup.go:736`). `helm uninstall` removes the bundled JobSet controller, its webhook configurations and its cluster-scoped RBAC, because all of those are in the release manifest. Under this ADR the CRD stays.
2. `setup init` runs the new detector. It finds `jobsets.jobset.x-k8s.io` with no controller resources and no release record, which is the `external` row.
3. It passes `--set jobset.install=false`. Trainer installs, no JobSet controller is installed, and TrainJobs create JobSets that nothing reconciles.

The ADR chose this deliberately. Decision 1 says "A CRD-only installation is conservatively reused as external", decision 2 says a post-reset init "may see a CRD-only installation and require operator cleanup or restoration of the external controller; it must warn rather than claim that a controller was detected", and Consequences repeats it. So this is a stated tradeoff rather than an oversight, and I disagree with the tradeoff. We would be degrading the documented remedy for a wedged install into a silently non-functional one on the common path, in order to protect the uncommon one. Decision 3 also forbids printing the CRD deletion as a recovery step, so the operator gets a warning and no instruction out.

It also removes the only way to clear a stale JobSet CRD schema. `helm upgrade` installs `crds/` only when `--install` falls through to a fresh install: `pkg/action/install.go` `RunWithContext` is the only CRD path in both v3.18.4 and v4.2.4, and `upgrade.go` in both versions declares only a `SkipCRDs` field, whose own comment says it "skips installing CRDs when install flag is enabled during upgrade", and never calls `installCRDs`. On our side `applyChartCRDs` (`pkg/setup/crds.go:70`) is wired only to the NVCRE chart inside `installHelmRelease` (`pkg/setup/helm.go:156-160`), never to Trainer. Today `setup reset` is the only thing that clears it.

Concrete alternative I would like considered in decision 2: when inspection succeeds and finds a CRD but no controller resources and no JobSet instances cluster-wide, keep `jobset.install=true` instead of folding that case into `external`. It is safe:

1. `installCRDs` creates `crds/` resources with `KubeClient.Create` and continues on `IsAlreadyExists` (v3.18.4 `pkg/action/install.go:160-181`, v4.2.4 `pkg/action/install.go:185-215`), so the pre-existing unlabeled CRD is skipped rather than adopted. Decision 1 already relies on this behaviour.
2. The templated subchart resources that would collide on ownership were removed by the uninstall, so there is nothing to adopt.
3. Only a genuinely orphaned CRD reaches this branch. The upstream JobSet chart has no toggle to disable its webhook configurations (`charts/jobset/values.yaml` exposes only cert-manager and metrics toggles), so a live install always leaves cluster-scoped webhook objects for the detector to find.

## 2. Decision 4 reverses an explicitly accepted ADR-073 consequence, and the amendment note does not say so

ADR-073 Consequences: "Anything a user manually placed in `kubeflow-system` is removed by recovery ... arbitrary foreign objects (a user's ConfigMap) are not individually protected. Accepted: mirroring the field recovery beats partial cleanup that leaves the conflicted Secrets behind."

Decision 4 here: "Unrelated foreign objects such as a user ConfigMap also block recovery."

That is a direct reversal, and I think it is the right call, but nothing in the record says it is one. The Context paragraph claims "What this record supersedes is the assumption that JobSet CRDs are always inside the `kubeflow-trainer` blast radius", and the Status line this PR adds to ADR-073 mentions only "JobSet CRD deletion and JobSet-instance recovery gating". Please widen both.

## 3. The namespace inventory as specified will refuse recovery on most real clusters

Decision 4 requires inventorying "every discoverable, listable namespaced resource type" in `kubeflow-system`, classifying each against the release manifest with UID owner-reference traversal, and blocking on anything unclassified, and `--auto-approve` explicitly does not bypass it. The certain consequence is that a single object placed in that namespace by any cluster add-on, or by an operator, permanently disables automatic recovery with no override.

The allowlist predicates are where I expect this to bite. "Controller-generated Events, EndpointSlices, and leader-election Leases whose relationship to verified release resources is established" is hard to make sound: an Event whose `involvedObject` has already been garbage collected cannot be tied to a verified release resource, and rollout churn in that namespace produces exactly those.

Either scope the inventory to what a user could plausibly care about losing, and drop Events and EndpointSlices from consideration rather than trying to prove their lineage, or say in Consequences that automatic recovery is now expected to be rare. A sentence on the restricted-kubeconfig case would help too, since full-namespace discovery plus list is a large ask and "incomplete namespace inventory blocks" turns a missing permission into a permanent block.

## 4. `unknown` is a dead end with no operator override

Decision 2 stops before any Helm mutation on `unknown`, and decision 6 says "Detection and the targeted subchart override are the interface", so there is no way for an operator to assert the answer. A transient API read failure then makes `setup init` unrunnable in a way it is not today. Either add an explicit override such as `--jobset-mode=bundled|external`, or state in decision 2 that `unknown` degrades to the hand-install plus `--skip-phases=deps` path, and print that path.

Related: please short-circuit the probe when no JobSet CRD exists. One `Get` proves `absent`, and it keeps the fresh-cluster path, which is the common one, from gaining a new failure class. As written the `absent` row requires "No JobSet CRD, controller resources, or bundled release manifest entries exist", so it reads as the full inspection.

## 5. Validation asks for a lifecycle test that has nowhere to run

"Include a Helm-backed lifecycle check before considering implementation complete: start with bundled JobSet and its actual unlabeled CRD, exercise the upgrade, and verify the controller stays installed."

Nothing in `.github/workflows/`, the `Makefile`, `test/`, or `hack/` invokes `nvcrectl setup init`. The UAT harness installs Trainer with its own `helm upgrade --install` in `test/uat/Tiltfile:54-55`, and `test/helm/render_test.go` only runs `helm template` against `helm/cluster-readiness-engine`. Please say where this check runs, whether that is a new UAT case, a manual pre-merge gate, or explicitly out of scope, otherwise it will quietly disappear at implementation time.

The rest of the validation table maps cleanly onto the existing `pkg/setup/testdata/init-recovery/` cases and looks realistic to me.

## 6. Decision 6 declines the issue's stated minimum without answering it

Issue #335 asks for detection "or at minimum exposes a way to pass Trainer chart values". Decision 6 rejects that outright, while Alternatives softens it to "Rejected as the primary fix", which reads as leaving it open as a secondary. Please pick one and address the reporter's minimum directly. Decision 3 does this well for the other half of the request, where it says plainly that conditional CRD deletion is being declined and why.

## Smaller

1. Context says `classifyTrainerInstallFailure` "requires `conflict`, `.data.`, and `secret`". There is a fourth condition, the namespace string: `classifyHelmInstallFailure` (`pkg/setup/helm.go:438-451`) also has `case !strings.Contains(output, namespace)`. Worth being exact, given decision 5 adds a sibling classifier.

2. The `values.yaml` block in Context is presented as a quote but is rewrapped and drops two lines. The 2.2.1 text is:

   ```yaml
   jobset:
     # -- Whether to install jobset as a dependency managed by trainer.
     # This must be set to `false` if jobset controller/webhook has already been installed into the cluster.
     install: true
     # -- String to fully override jobset release name.
     fullnameOverride: jobset
   ```

   Issue #335 quotes it verbatim. Either match it or mark the block as abridged.

3. ADR-073's Status line is now the only multi-sentence Status in `docs/designs/`. The existing idiom is a separate blockquote line, `> **Supersedes:** ...` as in ADR-034, or a dated inline note as in ADR-074. I would keep `> **Status:** Accepted` and put `> **Amended by:** [ADR-078](078-jobset-ownership.md) (proposed), JobSet CRD retention and recovery gating` underneath, so Status stays greppable.

4. The chart carries a second subchart with the identical shape: `charts/lws/crds/leaderworkerset.x-k8s.io_leaderworkersets.yaml`, `dataCache.lws.install: true`, `fullnameOverride: lws`. It is gated behind `dataCache.enabled: false` and we never set it, so there is no live equivalent today. One line in Notes would save the next reader the check.

## What I verified

Every code reference in the ADR is correct. `trainerHelmUpgradeArgs` (`pkg/setup/helm.go:262`) sets `jobset.controller.tolerations[0].operator=Exists` and never `jobset.install`; the unconditional JobSet `deleteCRDsByGroup` calls are at `pkg/setup/setup.go:513` and `pkg/setup/setup.go:752`; `trainerRecoveryGate` (`pkg/setup/setup.go:458`) blocks on any JobSet instance; reset does not delete `kubeflow-system`.

On the chart, pulling `oci://ghcr.io/kubeflow/charts/kubeflow-trainer:2.2.1` gives the digest the ADR cites, `sha256:8e8e3de0257be804f60dbc8e6784eaddb64b2e158a46cade30ca18b6030183b1`. The CRD is at `charts/jobset/crds/jobset.x-k8s.io_jobsets.yaml` and carries no Helm ownership metadata, only a `controller-gen.kubebuilder.io/version` annotation, so decision 1's premise holds. The Helm v4.2.4 link resolves and `installCRDs` is where the ADR says it is.

I also rendered the chart both ways. `--set jobset.install=false` removes exactly the 12 JobSet subchart documents, and the remaining 12 parent-chart documents are byte-identical between the two renders. That is the evidence for decision 2's mechanism and for decision 6's claim that no generic passthrough is needed. Leaving `jobset.controller.tolerations[0].operator=Exists` set with the subchart disabled is accepted by Helm without error, as decision 2 assumes.

078 is the right next number, and `make verify-doc-links` passes at 278 links.

asivanadi0 (commented):

Read ADR-078, the PR body, issue #335, ADR-073 on this branch, and `pkg/setup/{helm,setup,crds}.go` on `main`. Problem statement and code citations check out; separating controller ownership from CRD provenance is the right axis; retaining JobSet CRDs under missing provenance is the conservative cleanup call. I am not ready to accept the record as written — a few design gaps below are distinct from @ndipebot's pass (I agree with several of those, noted where they overlap).

### What looks solid

- Decision 1's premise matches the published chart: JobSet CRD under `charts/jobset/crds/` has no Helm ownership metadata; Helm's `installCRDs` path does not add it. Requiring full `meta.helm.sh/release-name` + `release-namespace` + `app.kubernetes.io/managed-by=Helm` for live bundled resources is the right bar.
- Decision 2's mechanism is minimal and correct: `--set jobset.install=false` only, leave the existing toleration `--set` alone, never disable a subchart already in the release merely because the CRD is unlabeled.
- Decision 5's "diagnostics shared / recovery eligibility unchanged" split is the right way to avoid accidentally arming ADR-073 SSA recovery on fresh ownership failures.
- Index entry 078 and the ADR-073 References link are in the right places; `make verify-doc-links` claim is plausible at this docs-only scope.

### Concrete gaps before I would accept this ADR

### 1. Forever-stuck JobSet CRD schema — missing Consequence, and adjacent to in-tree `#145` precedent

Decision 3 retains the JobSet CRD in every ownership state and forbids printing CRD deletion as a recovery step. Helm does not apply `crds/` on upgrade (only on fresh install) — the same fact `applyChartCRDs` documents for the NVCRE chart (`pkg/setup/crds.go`, issue #145). Trainer/JobSet has **no** equivalent reconcile today (`applyChartCRDs` is wired only into `installHelmRelease` for the NVCRE chart).

So under this ADR, a JobSet CRD that lands once (bundled or external) keeps that schema through Trainer upgrades **and** through reset/re-init, with no NVCRE-mediated refresh path. That is broader than the post-reset controller hole: it is a permanent schema-freeze for the shared CRD. Consequences should say so explicitly. Implementation should either (a) add a Trainer/JobSet CRD reconcile path analogous to `applyChartCRDs` when mode is `absent`/`bundled`, or (b) defer schema refresh as operator-manual / a follow-on ADR and name that deferral. Silent forever-freeze is not an acceptable implicit Consequence.

### 2. `printRetainedResources` cleanup one-liner is a footgun for this CRD

Today every retained row prints a ready-to-run `kubectl delete …` (`pkg/setup/setup.go` `retainedResource.cleanup`). Decision 3 extends that helper to the JobSet CRD and simultaneously says deletion "destroys JobSets across all namespaces" and must be "an operator decision after inspecting cluster-wide use, not as a required recovery step".

Those two collide: reset output that lists

```text
- CRD jobsets.jobset.x-k8s.io
    kubectl delete crd jobsets.jobset.x-k8s.io
```

is exactly the copy-paste trap the ADR is trying to remove from recovery plans. Please specify the retained-row contract for this CRD: warn about cluster-wide cascade, and either omit the one-liner, gate it behind an explicit "after you have inspected …" preface, or point at a documented manual cleanup page — do not reuse the namespace/secret cleanup shape unchanged.

Related: when `--skip-phases=deps` skips uninstall, the Trainer namespace row is already suppressed (`printRetainedResources` only mentions it if deps ran). Say whether the JobSet CRD retained line follows the same rule (only when deps actually ran) or always appears; as written it is ambiguous and the "we chose to retain it" framing is wrong on the skip path.

### 3. ADR-073 Decision body still mandates deleting `jobsets` — Status-only amendment is not enough

This PR rewrites ADR-073's Status line and adds a References bullet. The Decision text that implementers will follow is unchanged:

- Decision 3(a) still: uninstall → delete the four `trainer.kubeflow.org` / `jobset.x-k8s.io` CRDs → delete namespace → reinstall
- Decision 3(c) still prints that same procedure
- `trainerRecoveryCRDs` / `printManualTrainerRecovery` / `printTrainerRecoveryPlan` on `main` still list `jobsets.jobset.x-k8s.io`

Please amend those Decision bullets (or add an explicit "Amendments from ADR-078" block that rewrites them) so a reader of ADR-073 alone cannot ship the old four-CRD recovery. Same for the Context step-5 narrative that still presents JobSet CRD deletion as part of the field-validated procedure.

### 4. Decision 5 matcher needs a pinned real transcript, not the abbreviated Context quote

Context quotes only the annotation-validation clause. Live Helm ownership failures usually look like:

```text
Error: Unable to continue with install: ClusterRole "…" exists and cannot be imported …
invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-name" must equal "kubeflow-trainer": current value is "jobset"
```

Decision 5 requires `invalid ownership metadata` **and** a concrete JobSet resource kind/name from the subchart. That is the right narrowness, but the exact multi-line shape is load-bearing for a sibling classifier next to `classifyHelmInstallFailure`. Please add a Diagnostics golden derived from a real (or chart-rendered) collision transcript — including cluster-scoped objects that carry `fullnameOverride: jobset` names — so implementation cannot drift into matching bare `"jobset"` / `"ownership"` substrings, and so the Context quote is marked abridged or replaced with the full form.

Also: Context still says `classifyTrainerInstallFailure` "requires `conflict`, `.data.`, and `secret`". `classifyHelmInstallFailure` (`pkg/setup/helm.go:438-451`) has a fourth gate — the namespace string. Worth being exact before decision 5 adds a sibling.

### 5. Post-reset CRD-only → `external` needs a first-class Install/upgrade validation row (and a decided behavior)

I agree with @ndipebot's §1 on the substance: Decision 1 folding "CRD present, no controller, no release" into `external` + Decision 2's `jobset.install=false` turns the ordinary bundled `reset` → `init` cycle into a Trainer install with no JobSet controller. Decision 2's prose already admits this ("may see a CRD-only installation and require operator cleanup") — that is a stated tradeoff I also disagree with for the common path.

Beyond agreeing: the Validation table currently buries this under Ownership ("CRD-only installation") and Plans/docs ("reset/re-init … reports controller uncertainty"). It does **not** appear as an Install/upgrade case that forces a decided `jobset.install` value. Please add an explicit row such as "post-reset CRD-only, no JobSet instances, no controller evidence → ___" with the chosen `jobset.install` outcome. Without that row the most important behavioral fork can stay prose-ambiguous through implementation. (I would want that cell to be "keep enabled", matching ndipebot's proposed safe alternative, but the ADR must pick one and test it.)

### Smaller / clarifying

1. **Decision 6 vs issue #335's minimum.** Issue asks for detection "or at minimum exposes a way to pass Trainer chart values". Decision 6 rejects passthrough; Alternatives says "Rejected as the primary fix". Please pick one answer in Decision 6 the way Decision 3 does for conditional CRD deletion — either reject the minimum outright with rationale, or keep a narrow escape hatch (even if detection remains primary). Overlaps ndipebot §6; flagging because the issue's acceptance criteria otherwise stay unmet on paper.

2. **`unknown` with no operator assertion.** Decision 2 hard-stops; Decision 6 says detection is the interface. A transient API read failure then makes `setup init` worse than today. Either an explicit `--jobset-mode=bundled|external` (or equivalent) or a documented degrade to the hand-install + `--skip-phases=deps` path. Same concern as ndipebot §4; I would also short-circuit `absent` on a single CRD `Get`.

3. **ADR-073 Status idiom.** Prefer `> **Status:** Accepted` plus a separate `> **Amended by:** …` line (ADR-034-style) so Status stays greppable.

I am not approving on this pass — items 1–5 should be resolved (or explicitly deferred with named follow-ons) in the ADR text before acceptance. Happy to re-review a revision.

kaynetu:
Addressed the review feedback in [94fbff9](https://github.com/NVIDIA/cluster-readiness-engine/pull/336/commits/94fbff9c628021b6e9f69de5a6f261cc6a61ba51). Key changes:

- **Reset/init:** a verified CRD-only state keeps JobSet installation enabled, restoring the bundled controller while retaining the CRD. Existing JobSets without an identifiable controller or conflicting evidence block installation.
- **Detection and fallback:** a missing CRD short-circuits the probe. Unknown ownership prints the manual Trainer installation plus `nvcrectl setup init --skip-phases=deps` path. Generic values passthrough is explicitly outside this change.
- **Recovery boundaries:** protect unrelated namespace resources, exclude Events and EndpointSlices from inventory/lineage checks, and explain restricted-kubeconfig failures. ADR-073 now includes explicit proposed amendments to its recovery procedure and namespace-cleanup policy.
- **CRD retention:** require warning-only retained output with no deletion command, omitted when deps is skipped. Existing schemas remain unchanged by NVCRE; schema upgrades are operator-managed, with automatic reconciliation deferred to a separate design.
- **Validation:** require a dedicated setup lifecycle UAT job on an isolated Kind cluster and complete real Helm ownership-collision transcript fixtures in the implementation PR. Also corrected the classifier description, chart excerpt, status formatting, and LWS note.

`make verify-doc-links` passed (279 links), and `git diff --check` passed. This remains a proposed, documentation-only ADR; the runtime changes, UAT suite, and diagnostic fixtures will follow in the implementation PR.

asivanadi0 (approved):
Re-reviewed 94fbff9 against my prior COMMENT on a571aa4. All five design gaps are closed in the ADR text; CI is green. Approving the design record — point implementers at ADR-078 (and the ADR-073 amendment block) rather than re-deriving policy from this thread.

### Gap checklist (prior → 94fbff9)

1. **Forever-stuck JobSet CRD schema / `#145` / `applyChartCRDs`** — **fixed.** Decision 3 now has an explicit **Schema refresh is operator-managed** contract: schema stays frozen through upgrades/recovery/reset-init; NVCRE `applyChartCRDs` is not extended to Trainer/JobSet here; automatic shared-CRD reconcile is deferred to a separate design (also in Alternatives + Consequences). Silent forever-freeze is no longer implicit.

2. **`printRetainedResources` delete footgun** — **fixed.** **Retained-output contract** is warning-only (cluster-wide cascade), no `kubectl delete crd` one-liner, and omitted entirely under `--skip-phases=deps` (same rule as the Trainer namespace row). Validation has a Retained output row.

3. **ADR-073 Decision body still mandating `jobsets` deletion** — **fixed.** ADR-073 now has Status **Amended by** ADR-078, a Context **Historical procedure** callout on step 5, and an explicit **Proposed amendments from ADR-078** block that rewrites 3(a)/3(c)/5 (retain JobSet CRD; omit it from every recovery plan; update `trainerRecoveryCRDs` / `printManualTrainerRecovery` / `printTrainerRecoveryPlan` together). A reader of ADR-073 alone cannot ship the old four-CRD recovery without ignoring that block.

4. **Decision 5 pinned Helm transcript** — **fixed.** Context marks the ownership quote as abbreviated; `classifyTrainerInstallFailure` is correctly described with all four gates (incl. namespace). Decision 5 + Diagnostics validation require a complete real collision golden (chart-derived cluster-scoped kind/name with `fullnameOverride: jobset`, original framing/line breaks, versions + repro) in the implementation PR, with negative cases against bare-word matching.

5. **Post-reset CRD-only Install validation** — **fixed.** First-class `crd-only` state (not folded into `external`); Decision 2 keeps `jobset.install` enabled; Install/upgrade validation row covers `crd-only` → retain CRD + install bundled controller; Plans/docs row requires that report on reset/re-init. Matches the safe common-path outcome we wanted.

Smaller items from the prior pass (Decision 6 vs issue minimum; `unknown` manual escape / `absent` CRD-Get short-circuit; ADR-073 Status idiom) are also addressed.

No merge from me — acceptance/implementation still to follow the ADR as written.

ndipebot (commented):
Folded into the request-changes review below, which carries both findings inline at the lines they apply to.

ndipebot (commented):
Folded into the request-changes review below.

docs/designs/078-jobset-ownership.md
```diff
@@ -0,0 +1,190 @@
...
+   - When the bundled JobSet controller would be uninstalled, any JobSet instance blocks automatic recovery, even though its CRD is retained.
+   - When JobSet is external, instances outside `kubeflow-system` do not alone block recovery. Any JobSet instance inside `kubeflow-system` does block it because namespace deletion destroys that instance.
+   - Verify that no external JobSet controller or supporting resource within the protected inventory resides in `kubeflow-system`. Inspect namespaced resources and their ownership, including webhook service targets. Inventory namespaced resource types representing workloads, configuration, or storage, including Pods, Services, Secrets, ConfigMaps, PVCs, and custom resources. Exclude Events (both core and `events.k8s.io`) and EndpointSlices (`discovery.k8s.io`) entirely: do not list them for this gate or prove their lineage. Other discoverable namespaced resource types remain protected by default. Refuse namespace deletion when foreign or unclassified protected resources could be removed; do not rely only on default controller names or labels. Unrelated foreign objects such as a user ConfigMap also block recovery.
```
ndipebot, line 88: **Core `Endpoints` is missing from the exclusion, and it blocks recovery on every cluster.** Same shape as the Events and EndpointSlices problem this bullet just fixed.

`kubeflow-system` always has at least one Service. `templates/manager/service.yaml` renders `kubeflow-trainer-controller-manager` unconditionally, with a selector, even at `jobset.install=false`. The endpoints controller then creates a matching core `Endpoints` object, and that controller is still enabled by default: `newEndpointsControllerDescriptor` in `cmd/kube-controller-manager/app/core.go:385-391` sets no `isDisabledByDefault`.

That object is unclassifiable under this bullet:

1. Helm never renders `Endpoints`, so it is not in the release manifest.
2. It has no owner references at all. The controller builds it as `ObjectMeta{Name: service.Name, Labels: service.Labels}` (`pkg/controller/endpoint/endpoints_controller.go:460`, identical at v1.36.4 and v1.37.0, zero occurrences of `OwnerReferences` in the file). The UID traversal on line 89 has nothing to follow.
3. It is not in the allowlist on line 90.
4. It does carry `app.kubernetes.io/managed-by: Helm`, copied verbatim from the Service, but line 89 already rules that out: "a copied label ... is not sufficient". It gets no `meta.helm.sh/release-name`, because annotations are not copied.

So "Refuse namespace deletion when foreign or unclassified protected resources could be removed" fires on every run and automatic recovery never happens.

Add core `Endpoints` to the same exclusion as EndpointSlices. I would also state the rule the exclusion list is drawing, something like derived objects that no controller owns and that are rebuilt from primary resources, so the next kind of this shape is covered by the principle instead of another patch.

docs/designs/078-jobset-ownership.md
```diff
@@ -0,0 +1,190 @@
...
+
+Render the published Trainer 2.2.1 chart with JobSet enabled and disabled to verify resource identities and subchart behavior. Fake-client tests establish decisions and explicit deletion boundaries; they do not establish Helm metadata behavior or Kubernetes namespace garbage collection.
+
+**Automated setup lifecycle UAT is required in the implementation PR.** Add a separate Go test package under `test/uat/setup/`, a `make test-uat-setup` entry point, and a dedicated job in `.github/workflows/uat.yml`. The job builds the current `nvcrectl`, provisions its own disposable Kind cluster with an explicit kubeconfig/context, runs the lifecycle tests, exports CLI and cluster diagnostics on failure, and cleans up the cluster even when tests fail. Use real Helm releases and running controllers. Do not run destructive setup tests against the existing Tilt-managed catalog UAT cluster or simulate controller readiness with KWOK. The existing catalog UAT does not satisfy this requirement because it installs Trainer directly through Helm.
+
+The new suite invokes the built CLI's dependency phase, using phase selection to avoid unrelated setup requirements, and covers:
```
ndipebot, line 134: **Phase selection cannot isolate the dependency phase.** `--skip-phases=helm` is a no-op today.

`setup init` and `setup reset` both advertise a `helm` phase in the confirmation summary (`pkg/setup/setup.go:195` and `:682`), but nothing ever reads `skip["helm"]`. The only skip lookups in the file are `skip[phaseDeps]` at 160, 351, 737, 804 and `skip[phaseCR]` at 696. `installHelmRelease` runs unconditionally at `:225`, `uninstallHelmRelease` at `:704`, and `parseSkipPhases` (`:875-887`) does no validation, so the flag is accepted silently and only changes the printed summary. Pre-existing bug rather than something this ADR introduced, but the plan leans on it.

The knock-on is concrete. `LDFLAGS ?= -s -w -X main.version=$(VERSION)` with `VERSION ?= $(shell git describe --tags --always --dirty)` means a source-built `nvcrectl` is a dev build, and `resolveHelmChartVersion` (`pkg/setup/helm.go:83-92`) rejects it with `dev build %q has no published chart; pass --version <chart-version>`. `installDepsPhase` at `:218` completes first, then init returns a `[helm]` error and exits non-zero, so any scenario asserting a clean `setup init` fails.

Pick one and say which here:

1. Have the implementation PR make `skip["helm"]` work. Small fix, and worth doing anyway since the confirmation summary currently misreports it.
2. Pass `--chart-ref` at a chart packaged from the working tree plus `--version`, and accept that the job installs the NVCRE release too.
3. Drive the dependency phase through the package API and give up end-to-end CLI coverage.

1 and 2 together is what I would do, given that real Helm releases and running controllers are the stated point of the suite.

docs/designs/078-jobset-ownership.md
```diff
@@ -0,0 +1,190 @@
...
+
+- **Bundled upgrade:** seed a real bundled Trainer release with its actual unlabeled JobSet CRD at a chart version that exercises the CLI's upgrade path rather than its pinned-version skip. Run `setup init` and verify the intended upgrade occurred, the JobSet controller remains installed and becomes ready, and the CRD retains its UID. Pin the seed chart version in the test.
```
ndipebot, line 136: nit: seeding at a non-pinned chart version makes `planTrainerPhase` return `trainerActionAttemptRecover` rather than a plain install, so this case exercises the attempt-then-classify arm rather than a straight upgrade path. Probably what you want, but say it here or the first failure will read as a test bug.

docs/designs/078-jobset-ownership.md
```diff
@@ -0,0 +1,190 @@
...
+- **External JobSet:** seed a standalone JobSet Helm release in a separate namespace using a pinned chart version. Run `setup init` and `setup reset`; after each command, verify that the external release, controller, and CRD remain, their controller/CRD UIDs and Helm ownership metadata are unchanged, and the external controller is ready. Verify that Trainer installation disabled its JobSet subchart.
+
+Run these destructive scenarios serially with isolated cluster state and bounded readiness waits. Wire the dedicated CI job into the UAT workflow's PR-mirror push, main push, scheduled, and manual triggers. Passing this automated suite is required before the implementation is considered complete; manual testing supplements it rather than replacing it.
+
+Review existing golden diffs before requesting permission to regenerate them. After implementation, run the repository's required verification suite and documentation link checks.
+
+## Rationale
```
ndipebot, line 144: nit: `crd-only` is the largest change in this revision and Rationale has no bullet for it. The reasoning is in decision 2, but this section still reads as though ambiguity always means stop. A bullet on why installing a controller against an orphaned CRD is the safe direction, and disabling one is not, would close that.

ndipebot (changes requested):
Converting to request-changes. Two blocking items, both new in 94fbff9 rather than carried over from my earlier pass.

The good news first, so nobody re-litigates it. All six of my earlier points are genuinely addressed, and two of them are better than what I suggested. `crd-only` is tighter than my version: requiring successful controller, RBAC and webhook inspection plus a clean cluster-wide JobSet list, with anything short of that falling to `unknown`, is the right bar. And the note that the `absent` short circuit proves CRD absence but not absence of leftover colliding objects closes a hole I left open. The ADR-073 amendment block, the historical-procedure marker on step 5, and naming `trainerRecoveryCRDs`, `printManualTrainerRecovery` and `printTrainerRecoveryPlan` as things that must move together are all exactly right.

Two things I want fixed before this is accepted:

1. Core `Endpoints` is missing from the recovery inventory exclusion, so automatic recovery blocks on every cluster. Same failure mode the Events and EndpointSlices exclusion was just added to fix.
2. The UAT plan's isolation mechanism does not exist. `--skip-phases=helm` is silently a no-op today, so phase selection cannot isolate the dependency phase.

Neither is a design disagreement. The decisions are right. These are two places where the record specifies something the cluster or the code will not actually do. Reasoning inline, plus two nits.

I checked the rest against current sources rather than the tags cited, since the new UAT job will run on a recent Kind image. The endpoints controller behaves identically at v1.36.4 and v1.37.0 and is still on by default; the UAT trigger list matches `.github/workflows/uat.yml`; `test/uat/Tiltfile:54-55` does install Trainer through its own Helm invocation, so the claim that the catalog UAT does not cover this is correct; the classifier description now matches `classifyHelmInstallFailure`; the LWS note matches the chart; `verify-doc-links` passes at 279.

The `crd-only` reset/init round trip is the part I care most about and it is now specified well. Nice work on that.

docs/designs/078-jobset-ownership.md
kaynetu: Addressed in 4079a63. ADR-078 now excludes core Endpoints alongside Events and EndpointSlices throughout Decision 4, Implementation, Validation, and Consequences; a future kind remains protected until a fixture and rationale prove the same ownerless, derived, automatically rebuilt shape. Evidence: https://github.com/NVIDIA/cluster-readiness-engine/blob/4079a63ce38065e01283faf78140725eab665565/docs/designs/078-jobset-ownership.md#L83-L91

docs/designs/078-jobset-ownership.md
kaynetu: Addressed in 4079a63. We chose option 1 only: make helm a real phase and skip pull-secret setup plus NVCRE install on init, and NVCRE uninstall plus CRD cleanup on reset. Every new UAT scenario asserts dependency-phase behavior, so packaging and installing NVCRE would add an unrelated failure surface without additional coverage; using the package API remains rejected so CLI parsing and phase sequencing stay under test. Evidence: https://github.com/NVIDIA/cluster-readiness-engine/blob/4079a63ce38065e01283faf78140725eab665565/docs/designs/078-jobset-ownership.md#L110 and https://github.com/NVIDIA/cluster-readiness-engine/blob/4079a63ce38065e01283faf78140725eab665565/docs/designs/078-jobset-ownership.md#L129-L138

docs/designs/078-jobset-ownership.md
kaynetu: Addressed in 4079a63. The bundled-upgrade UAT now states that a deployed non-pinned release exercises trainerActionAttemptRecover, not trainerActionInstall, and that failures must be interpreted against the attempt-then-classify arm. Evidence: https://github.com/NVIDIA/cluster-readiness-engine/blob/4079a63ce38065e01283faf78140725eab665565/docs/designs/078-jobset-ownership.md#L142

docs/designs/078-jobset-ownership.md
kaynetu: Addressed in 4079a63. Rationale now explains why installing against a successfully proven crd-only state is safe, why disabling the controller is not, and why any incomplete or conflicting evidence remains unknown. Evidence: https://github.com/NVIDIA/cluster-readiness-engine/blob/4079a63ce38065e01283faf78140725eab665565/docs/designs/078-jobset-ownership.md#L150-L155

asivanadi0 (approved):
Re-reviewed **4079a63** against my APPROVE on 94fbff9 and @ndipebot's CHANGES_REQUESTED (plus the four inline threads). Both blocking CR items and both nits are closed in the ADR text; the five prior design gaps still hold. CI is green. Approving the design record again — point implementers at ADR-078 (and the ADR-073 amendment block) rather than re-deriving policy from this thread.

### ndipebot CR checklist (94fbff9 → 4079a63)

1. **Core `Endpoints` recovery exclusion** — **fixed.** Decision 4 now excludes Events (core + `events.k8s.io`), core `Endpoints`, and EndpointSlices under one rule (ownerless, derived from a primary resource, automatically rebuilt), with the Service→Endpoints always-present evidence and the “future kind stays protected until a fixture + rationale prove the same shape” bar. Implementation, Namespace-inventory validation, and Consequences carry the same list; missing list permissions for those exclusions do not block.

2. **UAT `--skip-phases=helm` no-op** — **fixed.** Chose option 1 only: Implementation + Validation **Phase selection** require a real `helm` phase constant, gating the full NVCRE Helm block on init (pull-secret + install) and reset (uninstall + NVCRE CRD cleanup), plus `parseSkipPhases` validation against known names. The UAT section correctly documents today's silent no-op (verified on `main`: only `skip[phaseDeps]` / `skip[phaseCR]` are read; `installHelmRelease` / `uninstallHelmRelease` run unconditionally) and makes fixing that a prerequisite of the suite rather than assuming it works. Rejecting package-API drive-by and not requiring NVCRE `--chart-ref` packaging for dependency-phase scenarios is the right scope call.

Nits also closed: bundled-upgrade UAT names `trainerActionAttemptRecover` / attempt-then-classify; Rationale has the `crd-only` install-vs-disable asymmetry bullet.

### Prior five design gaps (still hold on 4079a63)

1. **Schema refresh operator-managed / `#145`** — still explicit in Decision 3, Alternatives, Consequences; no silent forever-freeze.
2. **`printRetainedResources` footgun** — warning-only retained-output contract, no `kubectl delete crd`, omitted under `--skip-phases=deps`; Retained output validation row intact.
3. **ADR-073 Decision body** — Status **Amended by**, Historical procedure on step 5, and **Proposed amendments from ADR-078** rewriting 3(a)/3(c)/5 with the `trainerRecoveryCRDs` / `printManualTrainerRecovery` / `printTrainerRecoveryPlan` move-together still present.
4. **Decision 5 pinned collision transcript** — complete real golden + negatives still required in the implementation PR; Context quote still marked abbreviated; four-gate classifier description intact.
5. **Post-reset `crd-only` Install/upgrade row** — first-class state; keep `jobset.install` enabled; Install/upgrade + Plans/docs validation rows still force the decided outcome.

Minor non-blocking note: ADR-073's **Amended by** / amendment-block prose still names “Events and EndpointSlices” while ADR-078 now also excludes core `Endpoints`. Harmless because those lines say “as specified in ADR-078”; worth a one-line sync in a follow-up polish if anyone is grepping ADR-073 alone for the exclusion list.

No merge from me — acceptance/implementation still follow the ADR as written.

asivanadi0 (approved):
Re-reviewed **3cc878b** against my APPROVE on 4079a63 and @ndipebot's resolved CHANGES_REQUESTED items. Diffed `4079a63...3cc878b` (3 branch commits + merge of main `#330` PDB/affinity). The ADR delta is additive and does **not** reopen the prior CR items or the five design gaps. CI is green. Approving the design record again — point implementers at ADR-078 (and the ADR-073 amendment block) rather than re-deriving policy from this thread.

### What changed since 4079a63

1. **`db2b421` — Trainer recovery concurrency safeguards.** Decision 4 now requires operator-maintained quiescence, repeated inventories at destructive boundaries (post-confirmation before uninstall, workload rechecks before each Trainer CRD deletion, final namespace inventory before namespace delete), evidence-based comparison (not identical-inventory), UID delete precondition, and honest partial-recovery refusal without claiming rollback. Admission write barrier and retain-namespace cleanup are deferred with rationale (CRE 1.29+ / VAP 1.30). ADR-073 amendment block and Consequences updated to match.
2. **`d8e1b8d` — merge `main`.** Brings `#330` helm PDB + default preferred anti-affinity + ops docs/tests. Orthogonal to JobSet ownership; no ADR policy change.
3. **`3cc878b` — missing JobSet CRD + manual fallback.** Decision 2 requires a CRD `Get` before pinned skip or install/upgrade; absent CRD on an **existing** Trainer release stops for operator repair (Helm does not apply `crds/` on upgrade). Fresh `absent` still selects bundled install. Printed `--skip-phases=deps` fallback must state NVCRE still installs, require released CLI or `--version` (dev builds), optional `--chart-ref`, and must **not** recommend current `--skip-phases=deps,helm`. Validation rows: Missing CRD, Recovery timing, Inventory comparison; Plans/docs tightened.

Also closes my prior non-blocking note: ADR-073 **Amended by** / Safety rails now name Events, core `Endpoints`, and EndpointSlices (synced with ADR-078).

### ndipebot CR checklist (still hold on 3cc878b)

1. **Core `Endpoints` recovery exclusion** — still present end-to-end (Decision 4, Implementation, Namespace-inventory validation, Consequences); ADR-073 now matches.
2. **UAT `--skip-phases=helm` no-op** — Phase selection still requires a real `helm` gate; 3cc878b strengthens this by forbidding current `--skip-phases=deps,helm` as a workaround and documenting chart/`--version` prerequisites for the manual fallback.

Nits from that pass (attempt-then-classify UAT naming; `crd-only` Rationale asymmetry) remain closed from 4079a63.

### Prior five design gaps (still hold)

1. **Schema refresh operator-managed / `#145`** — Decision 3 / Alternatives / Consequences unchanged; missing-CRD repair is operator-manual, not silent auto-reconcile.
2. **`printRetainedResources` footgun** — warning-only contract, no `kubectl delete crd`, omitted under `--skip-phases=deps`; Retained output row intact.
3. **ADR-073 Decision body** — Status **Amended by**, Historical procedure, and **Proposed amendments** now cover decisions 1, 3(a), 3(c), and 5 (pinned-skip CRD check, quiescence/rechecks, partial refusal, Endpoints exclusion); `trainerRecoveryCRDs` / `printManualTrainerRecovery` / `printTrainerRecoveryPlan` move-together intact.
4. **Decision 5 pinned collision transcript** — complete real golden + negatives still required in the implementation PR.
5. **Post-reset `crd-only` Install/upgrade** — first-class state; keep `jobset.install` enabled; Install/upgrade + Plans/docs rows still force the decided outcome.

### New CodeRabbit threads (non-blocking on this pass)

- **Missing-CRD / chart-capable fallback** — addressed in 3cc878b (repair refusal + `--version`/`--chart-ref` contract).
- **`absent` short-circuit vs orphan controller in another NS** — already an explicit Decision 2 tradeoff (short circuit proves CRD absence; leftover collisions rely on Helm ownership / decision 5). New commits narrow risk for *existing* releases; they do not reopen the intentional fresh-path short circuit.
- **`helmStateUnknown` → `trainerActionInstall`** — existing ADR-073 / `planTrainerPhase` behavior on `main`; Decision 5 still routes those failures through shared diagnostics. Out of ADR-078's amendment scope; not a regression from 4079a63→HEAD.

No merge from me — acceptance/implementation still follow the ADR as written.

kaynetu:
Thanks for the review @asivanadi0 . I was addressing the coderabbit comments. Hopefully, the latest commit addresses all.

asivanadi0 (approved):
Re-reviewed **04cfd5d** against my APPROVE on 3cc878b. Diffed `3cc878b...04cfd5d` (single docs commit). Closes the two open CodeRabbit Majors from 3cc878b without reopening @ndipebot's CR items or the five design gaps. CI green on the tip (UAT still pending). Approving the design record again — point implementers at ADR-078 (and the ADR-073 amendment block) rather than re-deriving policy from this thread.

### What changed in 04cfd5d

1. **Missing-CRD `absent` is no longer a one-`Get` short circuit.** Decision 1 now requires a complete cluster-scoped scan of `ValidatingWebhookConfiguration` / `MutatingWebhookConfiguration` / `ClusterRole` for explicit `jobset.x-k8s.io` rules after CRD `NotFound`. Clean scan → `absent` (bounded to the verified chart pattern); any match or incomplete list → `unknown` with named objects (no live-controller claim). Rationale + Consequences state the detection contract and permission cost; Validation adds a surviving-external-controller / deleted-CRD UAT.
2. **`helmStateUnknown` refuses before Helm.** Decision 2 + Implementation + Validation + ADR-073 amendment block: unreadable release state stops skip/mutation, preserves the state-query cause, and prints the manual fallback. Fresh install only from `not installed` / `uninstalled`. Replaces ADR-073's `unknown-state-attempts` golden in the implementation PR.

### CR + CodeRabbit checklist

- **ndipebot CR (Endpoints exclusion; real `helm` phase)** — still hold; not touched by this tip.
- **CodeRabbit: controller leftover after CRD deletion** — closed by the fingerprint scan + UAT row.
- **CodeRabbit: `helmStateUnknown` → install** — closed by the evidence-dependent refusal.
- All review threads on the PR are resolved.

Prior five design gaps and nits from 4079a63 still hold. No merge from me.

docs/designs/078-jobset-ownership.md
```diff
@@ -0,0 +1,233 @@
...
+
+   **Reset inspection contract:** reset does not need the JobSet ownership detector or the recovery namespace inventory to authorize its existing Trainer release uninstall. It uninstalls only the configured `kubeflow-trainer` release in `kubeflow-system`, leaves that namespace intact, and never deletes JobSet CRDs. Ownership-inspection failures therefore do not block reset or trigger fallback cleanup. The retained-resource lookup is best effort: a failed lookup prints that the JobSet CRD *may remain*, with the lookup error, rather than claiming it is absent. Helm uninstall and Trainer CRD cleanup failures still propagate normally. `--skip-phases=deps` continues to bypass dependency removal. This is an explicit reset operation, not automatic recovery; the recovery safety gate does not apply to it.
```
ndipebot: **"Helm uninstall and Trainer CRD cleanup failures still propagate normally" is not true of the Trainer uninstall, and line 144 pins the version we do not have.**

`uninstallTrainerHelmRelease` warns and returns `nil`:

```go
if err := runHelm(helmPath, args, out); err != nil {
    fmt.Fprintf(out, "[deps] Warning: failed to uninstall %s: %v\n", trainerReleaseName, err)
}
return nil
```

(`pkg/setup/helm.go:294-297`). So the caller's error check at `pkg/setup/setup.go:742` is reachable only through `ensureHelm()`, never through a failed uninstall. After `helm uninstall kubeflow-trainer` fails, reset carries straight on to `deleteCRDsByGroup` at `:749` and `:752` and then prints "NVCRE reset successfully." at `:732`, exit 0.

The CRD half of the sentence is correct, for what that is worth: `deleteCRDsByGroup` returns its errors at `pkg/setup/setup.go:1186` and `:1199` and the caller wraps them at `:749-751`. The NVCRE uninstall is correct too, at `helm.go:221`. It is specifically the Trainer uninstall that swallows.

This sentence has been in the record since `a571aa4` and has survived every revision since, so I read it as an oversight rather than a decision. What makes it worth a blocking-sized comment rather than a nit is that it now points two parts of the ADR at each other. Line 130 tells the implementer to keep the existing code: "Reset keeps the existing exact-release uninstall and Trainer CRD cleanup". Line 144 requires a validation case where "actual uninstall/Trainer cleanup failures propagate". The code line 130 preserves cannot pass the case line 144 requires.

"actual" does not get us out of it either. `--ignore-not-found` is already passed at `helm.go:288`, so an absent release is not an error at the Helm level at all, which means "actual failure" and "any error returned by `runHelm`" are the same set here.

Which way do you want it? If 144 wins, that is a deliberate behaviour change: `setup reset` starts exiting non-zero on a failed Trainer uninstall where it exits 0 today, and the warning at `helm.go:295` becomes a returned error. That is fine by me, but it should be stated as a change in the record rather than described as existing behaviour, and 130 needs to stop saying "keeps the existing".

docs/designs/078-jobset-ownership.md
```diff
@@ -0,0 +1,233 @@
...
...`discovery.k8s.io`) entirely: do not list them for this gate or prove their lineage. These share one shape: the endpoints controller and the event machinery create them from a primary resource, they carry no owner references (core `Endpoints` copies the Service's labels, including `app.kubernetes.io/managed-by: Helm`, but not its annotations), and they are rebuilt automatically when the primary resource returns. The Trainer chart renders the `kubeflow-trainer-controller-manager` Service unconditionally, even with `jobset.install=false`, so a core `Endpoints` object is present in `kubeflow-system` on every cluster; inventorying it would block recovery everywhere. The same evidentiary rule governs any future exclusion, but an unknown kind remains protected until a concrete fixture and rationale establish that it is ownerless, derived from a protected primary resource, and automatically rebuilt. Other discoverable namespaced resource types remain protected by default. Refuse namespace deletion when foreign or unclassified protected resources could be removed; do not rely only on default controller names or labels. Unrelated foreign objects such as a user ConfigMap also block recovery.
```
ndipebot: **The general rule is stated, which closes my ask, but the rule as written is false for one of the three kinds it covers.**

"they carry no owner references ... and they are rebuilt automatically when the primary resource returns" is right for Events and right for core `Endpoints`. I re-checked the latter rather than take my own earlier word for it: the endpoints controller builds the object as `ObjectMeta{Name: service.Name, Labels: service.Labels}`, and `OwnerReferences` does not appear anywhere in `pkg/controller/endpoint/endpoints_controller.go`.

EndpointSlices are the other way round on both legs.

1. Control-plane slices do carry an owner reference. From the Kubernetes EndpointSlices docs, under Ownership: "In most use cases, EndpointSlices are owned by the Service that the endpoint slice object tracks endpoints for. This ownership is indicated by an owner reference on each EndpointSlice as well as a `kubernetes.io/service-name` label." The controller sets it with `metav1.NewControllerRef(service, gvk)` in `staging/src/k8s.io/endpointslice/utils.go`, assigned to `OwnerReferences` a few lines down.
2. The ownerless EndpointSlices are the manually managed ones, and those are precisely the ones that nothing rebuilds.

So no EndpointSlice satisfies all three legs of the test this line hands to future kinds. Control-plane slices fail "ownerless"; manually managed ones fail "derived" and "rebuilt automatically". The rule is fine as a filter, it just does not admit the kind it was written to admit, which means the next reviewer applying it by principle gets the wrong answer for EndpointSlices.

Line 93 opens decision 4 with the same characterization, "Derived objects that no controller owns and that Kubernetes rebuilds from primary resources", so both sites need the same edit.

To be clear about the scope: I am not asking to move EndpointSlices out of the exclusion. Excluding them is right, and 100, 101, 130, 149, 194 and 195 are all correct as they stand. Line 149 already handles this carefully, attributing "no owner references" only to the core `Endpoints`. It is just the justification at 93 and 98 that needs to match. Something along the lines of: Events and core `Endpoints` are ownerless and rebuilt from their primary resource; EndpointSlices are either owned by their Service through a controller reference and likewise rebuilt, or manually managed, in which case losing them is accepted under the consequence already stated at 194 rather than allowed to block recovery on every cluster.

Smaller, same line: creation is attributed to "the endpoints controller and the event machinery", which leaves out the endpoint slice controller.

docs/designs/078-jobset-ownership.md
```diff
@@ -0,0 +1,233 @@
...
+| Release state | `helmStateUnknown` from a missing `helm` binary, a `helm status` failure other than not-found, or unparseable status output stops before skip or Helm mutation, preserves and reports the specific state-query cause, and prints the manual fallback; a successful separate manifest read does not bypass this refusal; `not installed` and `uninstalled` select fresh install; the ADR-073 `unknown-state-attempts` case is replaced by this refusal |
+| Reset | Bundled, external, unlabeled, and uncertain CRDs are retained; external controller resources remain; ownership detector is not required; failed retained-CRD lookup prints “may remain” and does not block reset; actual uninstall/Trainer cleanup failures propagate; namespace is never deleted; `--skip-phases=deps` still bypasses removal |
+| Retained output | Present JobSet CRD produces a cluster-wide cascade warning with no deletion command; failed lookup produces “may remain”, the error, and the warning without a deletion command; confirmed absence omits the entry; skipped deps omits both the JobSet CRD entry and lookup |
```
ndipebot: nit: this row gives `deps` the treatment and leaves the new `helm` phase without it, even though line 89 states the principle generally.

Line 89: "With `--skip-phases=deps`, omit the JobSet CRD entry and its lookup, just as the Trainer namespace entry is omitted: the dependency phase made no retention decision." Once `helm` is a phase that can actually be skipped, the identical reasoning applies to the NVCRE entries, and nothing in the ADR says so.

In the code, `printRetainedResources` is called unconditionally at `pkg/setup/setup.go:731`, and the NVCRE namespace and pull-secret entries at `:789-800` are gated on existence alone. Only the Trainer namespace entry is skip-aware, at `:804`, with the comment "Only mention the Trainer namespace when the deps phase actually ran". So `setup reset --skip-phases=helm` against a cluster with NVCRE installed would print

```
Retained resources (not removed by reset):
  - Namespace nvcre
      kubectl delete namespace nvcre
```

for a namespace that still holds the running release the operator just asked reset to leave alone. `[helm]` is a documented reset phase (`setup.go:627`) and the flag help reads "Comma-separated phases to skip (e.g., deps)" rather than "only deps" (`:642`), so that is a supported invocation the moment the gate lands.

It is not reachable from the UAT commands at line 160, for the record. The only two things that create the `nvcre` namespace are `--create-namespace` in the NVCRE upgrade args (`helm.go:183`) and `EnsureNamespace` inside `setupControllerSecret` (`setup.go:595`), and line 128 puts both inside the gate, so in the ADR's own flow the namespace never exists to be printed.

Suggest extending this row with "skipped helm omits the NVCRE namespace and pull-secret entries", and adding a `skip-helm-omits-nvcre-entries` case next to the existing `pkg/setup/testdata/print-retained-resources/skip-deps-omits-trainer-namespace/`, which already pins the deps half as a controlled differential against `all-retained`.

Two things for whoever implements line 160, neither worth its own row. There is no `[helm] Skipped.` line to match the three that exist at `:352`, `:697` and `:738`, and `printPhaseList` runs only inside `if !autoApprove` (`:188`, `:675`), so under `--auto-approve` a skipped helm phase leaves no trace in the transcript at all. And `asymmetricChartRefsWarning`'s last arm (`helm.go:550-554`) would start telling operators "so the `[helm]` phase pulls the NVCRE chart from ghcr.io" about a pull that no longer happens, because only the deps arm short circuits (`:540-543`) and the call site passes just `skip[phaseDeps]` (`setup.go:160`).

docs/designs/078-jobset-ownership.md
```diff
@@ -0,0 +1,233 @@
...
+
+   For `absent`, proceed with the normal bundled Helm install path only when the release-state decision actually selects a fresh installation; an existing release's upgrade path requires the repair above. The CRD `Get` alone establishes CRD absence, not controller absence: a standalone controller survives deletion of its CRD, watches cluster-wide, and if its rendered object identities do not collide with the bundled subchart's identities, so a fresh install would recreate the CRD and add a second controller reconciling every JobSet. That is why `absent` also requires the fingerprint scan in decision 1; matching JobSet-related resources select `unknown` and the operator path, leaving the standalone installation untouched. A clean scan authorizes this path only within decision 1's stated installation-pattern assumption. Where a leftover object does collide by name, Helm's ownership checks reject the install and decision 5 supplies targeted guidance; do not use `--take-ownership`. The short circuit only skips ownership probing for install selection. It does not bypass the existing release-state check or decision 4's safety checks before destructive recovery.
```
ndipebot: nit: this sentence lost its main clause in `04cfd5d`. "a standalone controller survives deletion of its CRD, watches cluster-wide, and if its rendered object identities do not collide with the bundled subchart's identities, so a fresh install would recreate the CRD and add a second controller" leaves the `if` with nothing to attach to.

Deleting the `so` before "a fresh install" fixes it and does not change the meaning. The text this replaced, in `3cc878b`, read fine.

docs/designs/073-setup-retry-convergence.md
```diff
@@ -25,6 +29,16 @@ So the reporting gap is closed by #188, but `setup init` itself still retries in
    - **(c) Fail fast with the exact procedure.** When the failure is conflict-classified but the safety gate does not pass — or classification is ambiguous — `setup init` fails with the manual recovery procedure from issue #180 printed verbatim (uninstall, the four CRD deletions, namespace deletion, pinned re-init) plus the reason automatic recovery was refused.
    - Option (b), Helm-level `--force` or `--take-ownership`, is rejected outright (see Alternatives Considered).
 
+   > **Proposed amendments from ADR-078 (not yet accepted):** upon acceptance and implementation, the following replace the corresponding instructions in decisions 1, 3(a), 3(c), and 5:
```
ndipebot: nit: the preamble says these replace the instructions in "decisions 1, 3(a), 3(c), and 5", but five bullets follow and the second is "Unknown release state (4)".

`3cc878b` added the preamble with four bullets and `04cfd5d` added the fifth without widening the list. Decision 4's `unknown` row at line 50 is exactly what ADR-078 line 59 amends, so it belongs in the enumeration.

The same undercount is in two other places: the "Amended by" header at line 5 and the References entry at line 117 both name three topics (CRD retention, recovery gating, protection of unrelated resources) and leave out the release-state change, which is the one that replaces a row in this record's own decision table.

ndipebot (commented):
Third pass, against `04cfd5d`. Both blocking items are closed and so are both nits. I checked each one against the code rather than against the evidence links, and I also went through `db2b421`, `3cc878b` and `04cfd5d`, which had had no review from me at all.

Closure first, so none of it gets re-litigated:

1. Core `Endpoints` is now excluded at every site. It appears at lines 93, 98, 100, 101, 130, 149, 194 and 195, and the ADR-073 amendment carries it at line 38. There is no site left that enumerates Events and EndpointSlices while omitting it.
2. Phase selection took option 1. The defect is described accurately at line 158, the fix is at line 160, and it is carried into Implementation at line 128 and Validation at line 151. Worth naming the dependency out loud: the UAT is safe from the dev-build problem only because it never installs the NVCRE release, which is true only once the gate actually exists.
3. The bundled-upgrade note at line 164 is there and it is right. `planTrainerPhase` (`pkg/setup/setup.go:324-336`) returns `trainerActionAttemptRecover` for a deployed release at a non-pinned version, and `trainerActionInstall` only for `helmStateNotInstalled`, `helmStateUninstalled` and `helmStateUnknown`.
4. The `crd-only` Rationale bullet is at line 176.

The three unreviewed commits are good work and I want to say so. The quiescence prerequisite, the recheck after confirmation, and the sentence admitting that these checks do not prove the absence of concurrent writers are the honest version of what decision 4 can and cannot promise. I would rather have that than a safety guarantee we cannot keep. The bounded-detection contract in `04cfd5d` is the same kind of honesty, and the accepted-consequences list at 188 now discloses the refusal classes instead of burying them.

Five things left, none of them design disagreements:

1. Line 91 says Helm uninstall failures propagate. For the Trainer release they do not, and have not since `a571aa4`. It also conflicts with the validation row at 144.
2. The general rule I asked for is stated at 98, which closes my ask, but the rule as written is false for one of its own three kinds. That is the remainder of item 1 rather than a new objection.
3. `073` line 32 undercounts its own bullets.
4. The retained-output row at 145 never applies its own principle to the new `helm` phase.
5. Line 71 lost a clause in `04cfd5d`.

Reasoning inline. I will clear the changes-requested once the 91 and 144 conflict has an answer, since that one needs a decision from you rather than an edit from me. The other four are small enough to land either way.

## What I verified this time

- Every line number cited above, read at `04cfd5d` rather than trusted from the reply threads.
- `uninstallTrainerHelmRelease`, `uninstallHelmRelease`, `uninstallDepsPhase`, `deleteCRDsByGroup`, `printRetainedResources`, `printPhaseList`, `planTrainerPhase` and `asymmetricChartRefsWarning` in `pkg/setup/`, plus every `skip[...]` lookup in the package.
- The EndpointSlice and core `Endpoints` owner-reference behaviour against upstream Kubernetes source and docs rather than from memory.
- `git log -L` on lines 71, 91, 93, 98 and 130 of `078` and line 32 of `073`, to check whether each item is new text or text that has been sitting there through several revisions.
- The Trainer 2.2.1 chart rendered at `jobset.install=true` and `false`, to re-check the claims in decisions 1 and 2. They hold.

docs/designs/078-jobset-ownership.md
ndipebot: Confirmed closed. Core `Endpoints` is now excluded at lines 93, 98, 100, 101, 130, 149, 194 and 195, and in the ADR-073 amendment at line 38. I checked each of those rather than the summary, and there is no site left that lists Events and EndpointSlices without it.

The general rule is stated at 98, which is the second half of what I asked for. It has one problem, which I have left at that line: the rule is false for EndpointSlices, which do carry a controller owner reference to their Service. The exclusion itself is right and I am not asking to change it, only the justification at 93 and 98.

docs/designs/078-jobset-ownership.md
ndipebot: Confirmed closed, option 1. The defect description at 158 matches the code (no `phaseHelm` constant, `"helm"` only as a literal in the two `printPhaseList` calls at `setup.go:195` and `:682`, no `skip["helm"]` lookup anywhere in the package, and `parseSkipPhases` validating nothing). The fix is at 160 with Implementation at 128 and Validation at 151.

One thing to keep in view: this closes as a plan, and the plan's own UAT commands depend on the gate existing. The dev-build problem I raised is avoided only because the gate stops the NVCRE release from being installed at all, so the UAT is not evidence for the gate, it is downstream of it.

I have left a related nit at line 145: the gate changes what `printRetainedResources` should print, and the Validation table only covers the deps half.

docs/designs/078-jobset-ownership.md
ndipebot: Confirmed closed, line 164. Checked against `planTrainerPhase` (`pkg/setup/setup.go:324-336`): a deployed release at a non-pinned version returns `trainerActionAttemptRecover`, and `trainerActionInstall` comes back only for `helmStateNotInstalled`, `helmStateUninstalled` and `helmStateUnknown`. The note says the right thing.

docs/designs/078-jobset-ownership.md
ndipebot: Confirmed closed, Rationale bullet at line 176.

docs/designs/078-jobset-ownership.md
kaynetu: Chose option 1 (“144 wins”) in 01695fb. The [decision](https://github.com/NVIDIA/cluster-readiness-engine/blob/01695fbedaa3393bd94d8b5cef22038a3455ae8c/docs/designs/078-jobset-ownership.md#L91) now explicitly changes the shared helper from warning-only to error propagation: reset stops before Trainer CRD deletion; recovery also stops before namespace deletion and reinstall. It preserves Helm diagnostics, acknowledges partial outcomes, and retains `--ignore-not-found` without promising retry convergence. [Implementation](https://github.com/NVIDIA/cluster-readiness-engine/blob/01695fbedaa3393bd94d8b5cef22038a3455ae8c/docs/designs/078-jobset-ownership.md#L130), [reset/recovery validation requirements](https://github.com/NVIDIA/cluster-readiness-engine/blob/01695fbedaa3393bd94d8b5cef22038a3455ae8c/docs/designs/078-jobset-ownership.md#L144), and the [rejected warning-and-continue alternative](https://github.com/NVIDIA/cluster-readiness-engine/blob/01695fbedaa3393bd94d8b5cef22038a3455ae8c/docs/designs/078-jobset-ownership.md#L203) agree. This commit updates the ADR only; implementation and failure-path tests remain required.

docs/designs/078-jobset-ownership.md
kaynetu: Corrected both sites in 01695fb: [Decision 4 introduction](https://github.com/NVIDIA/cluster-readiness-engine/blob/01695fbedaa3393bd94d8b5cef22038a3455ae8c/docs/designs/078-jobset-ownership.md#L93) and [per-kind rationale](https://github.com/NVIDIA/cluster-readiness-engine/blob/01695fbedaa3393bd94d8b5cef22038a3455ae8c/docs/designs/078-jobset-ownership.md#L98). Controller-managed EndpointSlices normally have Service controller owner references; manually managed endpoint objects are not guaranteed to be rebuilt. Also corrected the Events claim: deleted historical Events are not reconstructed. The exclusions remain unchanged, with those losses explicitly accepted. Future exclusions require a fixture and per-kind rationale covering regeneration and loss consequences, not a universal “ownerless and rebuilt” test.

docs/designs/078-jobset-ownership.md
kaynetu: Specified in 01695fb: skipped Helm omits both NVCRE namespace/pull-secret entries and their lookups ([retained-output contract](https://github.com/NVIDIA/cluster-readiness-engine/blob/01695fbedaa3393bd94d8b5cef22038a3455ae8c/docs/designs/078-jobset-ownership.md#L89)), with the requested `skip-helm-omits-nvcre-entries` case required in [validation](https://github.com/NVIDIA/cluster-readiness-engine/blob/01695fbedaa3393bd94d8b5cef22038a3455ae8c/docs/designs/078-jobset-ownership.md#L145). [Phase selection](https://github.com/NVIDIA/cluster-readiness-engine/blob/01695fbedaa3393bd94d8b5cef22038a3455ae8c/docs/designs/078-jobset-ownership.md#L128) also requires `[helm] Skipped.` under `--auto-approve` and skip-aware chart-pull warnings; both are covered by the [phase validation requirements](https://github.com/NVIDIA/cluster-readiness-engine/blob/01695fbedaa3393bd94d8b5cef22038a3455ae8c/docs/designs/078-jobset-ownership.md#L151). These are implementation requirements, not claims that the CLI gate or tests already exist.

docs/designs/078-jobset-ownership.md
kaynetu: Fixed in 01695fb: removed the stray “so” from the [missing-CRD explanation](https://github.com/NVIDIA/cluster-readiness-engine/blob/01695fbedaa3393bd94d8b5cef22038a3455ae8c/docs/designs/078-jobset-ownership.md#L71). No policy change.

docs/designs/073-setup-retry-convergence.md
kaynetu: Fixed all three sites in 01695fb: the [amendment preamble](https://github.com/NVIDIA/cluster-readiness-engine/blob/01695fbedaa3393bd94d8b5cef22038a3455ae8c/docs/designs/073-setup-retry-convergence.md#L32) now includes decision 4, and the [Amended by header](https://github.com/NVIDIA/cluster-readiness-engine/blob/01695fbedaa3393bd94d8b5cef22038a3455ae8c/docs/designs/073-setup-retry-convergence.md#L5) and [References entry](https://github.com/NVIDIA/cluster-readiness-engine/blob/01695fbedaa3393bd94d8b5cef22038a3455ae8c/docs/designs/073-setup-retry-convergence.md#L117) explicitly include unknown-release-state refusal and the pinned-skip CRD check. The historical ADR-073 body remains unchanged.

docs/designs/078-jobset-ownership.md
```diff
@@ -0,0 +1,236 @@
...
+
+   When the CRD exists, resolve the JobSet controller mode from the exact Trainer release's stored manifest and live JobSet controller, RBAC, and webhook resources. Use the pinned subchart's resource identities and webhook service references; do not assume the external Helm release is named `jobset`. Read the release manifest before uninstalling it, including failed/partial release state. Values alone are insufficient: `jobset.install=true` does not prove that the release created a pre-existing CRD.
```
ndipebot, line 41: Non-blocking, but this is the one I would want resolved before implementation rather than during it.

The name-independent scan is bound to the missing-CRD path, and that leaves `crd-only` reachable on a cluster that has a live external controller.

Line 37 introduces the `jobset.x-k8s.io` rule scan as the follow-up to a `NotFound` `Get`, and closes it with "This path reads no release manifests, namespaced controller resources, or JobSet instances." Implementation line 127 pins the same gating: "on CRD `NotFound`, run the cluster-scoped webhook-configuration and ClusterRole scan". This line is the CRD-present path, and it specifies discovery differently, as "Use the pinned subchart's resource identities and webhook service references".

So when the CRD exists, discovery is scoped to the subchart's identities, and an external install whose rendered names do not collide with `jobset-*` is invisible to it. On a cluster with the CRD present and no JobSets yet, which is the ordinary state right after an operator installs JobSet and before the first workload, every `crd-only` condition at line 50 holds: no controller found, no supporting resources found, no release evidence, empty cluster-wide list. Line 63 then keeps the subchart enabled and `setup init` adds a second cluster-wide JobSet controller.

That is the hazard line 71 describes on the absent-CRD path, and the Rationale at 176 depends on the CRD-present inspection being identity-independent: "Installing the bundled controller against a CRD that successful inspection shows has no controller ... adds a controller where none exists."

This line already anticipates the shape, with "do not assume the external Helm release is named `jobset`", so I read it as unstated rather than decided. One sentence saying whether the decision 1 scan also runs when the CRD exists would close it.

The same gap leaves one Validation case underivable. The Ownership row at 140 requires a "non-Helm controller" case, and line 43's test is Helm metadata only, so nothing in decision 1 says whether such a controller lands in `external` or `crd-only`. Rows 64 and 65 take opposite actions, so it matters.

docs/designs/073-setup-retry-convergence.md
```diff
@@ -25,6 +29,16 @@ So the reporting gap is closed by #188, but `setup init` itself still retries in
...
+   > - **Unknown release state (4):** `helmStateUnknown` stops the dependency phase before skip or Helm mutation, reports the preserved state-query cause, and prints ADR-078's manual-install fallback, instead of `upgrade --install`. The install decision is now evidence-dependent, and an unreadable release cannot be classified as fresh or existing; the `unknown-state-attempts` case is replaced by that refusal.
+   > - **Automatic recovery (3a):** require ADR-078 decision 4's operator-maintained quiescence and repeat its complete gate after confirmation before uninstalling Trainer. Recheck workload blockers before each Trainer CRD deletion, retain `jobsets.jobset.x-k8s.io`, and perform fresh discovery and a final protected namespace inventory before namespace deletion. Compare against expected cleanup using preserved ownership and UID evidence; delete with the inspected namespace UID precondition, wait, then reinstall using the selected JobSet mode preserved under ADR-078 decision 2. Keep the existing SSA eligibility, confirmation, and one-attempt limit.
```
ndipebot, line 36: nit: the uninstall abort is missing from this bullet, and this bullet is what an implementer works from.

ADR-078 line 91 makes a failed Trainer uninstall stop automatic recovery "before CRD deletion, namespace deletion, or reinstall". This bullet rewrites the 3(a) sequence step by step, from the post-confirmation gate through to reinstall, and never mentions it. The `Amended by` list at line 5 does not either, even after this round widened it.

Same shape as the undercount I raised last round. ADR-078's own decision 4 has the same hole: the enumeration at 105 stops recovery on "Any blocker or required inspection failure", and a failed `helm uninstall` is neither of those.

Worth a clause here and at line 5, since this is the record that stands as accepted until ADR-078 is.

docs/designs/078-jobset-ownership.md
```diff
@@ -0,0 +1,236 @@
...
+
+   **The manual fallback still installs NVCRE.** Every printed `nvcrectl setup init --skip-phases=deps` fallback must state that it skips Trainer setup only. Use a released CLI whose matching NVCRE chart is available, or supply `--version <nvcre-chart-version>` explicitly (required when the CLI reports a dev version, including an untagged source build). The selected chart version must be reachable at the default chart location or at an optional `--chart-ref <nvcre-chart-reference>`, with any required registry access. `--chart-ref` alone does not satisfy a dev build's version requirement. For that build, print `nvcrectl setup init --skip-phases=deps --version <nvcre-chart-version>`, with the optional chart-reference override explained separately. These flags select the NVCRE chart, not the manually installed Trainer chart. Do not recommend `--skip-phases=deps,helm` as a current workaround: the existing CLI does not gate the Helm phase. The phase-gating implementation below is a prerequisite for the future dependency-only UAT commands.
```
ndipebot, line 69: nit: "the existing CLI does not gate the Helm phase" is repealed by this ADR's own Implementation bullet at 128, which adds the `helm` phase constant and gates the whole NVCRE Helm block. The build that prints this fallback is the build that has the gate, so "current" is stale on arrival, and Validation 152 inherits the word.

The instruction is still right, just for a different and durable reason: the fallback exists to install NVCRE, so skipping `helm` defeats its purpose. That reason does not expire.

docs/designs/078-jobset-ownership.md
```diff
@@ -0,0 +1,236 @@
...
...h a Service controller owner reference. Controller-managed endpoint objects can be regenerated from their primary resources, but manually managed Endpoints and EndpointSlices are not guaranteed to be rebuilt. Their loss is explicitly accepted by this exclusion, not evidence that they are reconstructible. The Trainer chart renders the `kubeflow-trainer-controller-manager` Service unconditionally, even with `jobset.install=false`, so a core `Endpoints` object is present in `kubeflow-system` on every cluster; inventorying it would block recovery everywhere. An unknown kind remains protected until a concrete fixture and per-kind rationale establish why exclusion is appropriate, identify whether and how objects are regenerated, and explicitly document the consequences of losing non-reconstructible or manually managed data. Neither absence nor presence of owner references alone justifies exclusion. Other discoverable namespaced resource types remain protected by default. Refuse namespace deletion when foreign or unclassified protected resources could be removed; do not rely only on default controller names or labels. Unrelated foreign objects such as a user ConfigMap also block recovery.
```
ndipebot, line 98: nit: the two Kubernetes claims that carry the `Endpoints` exclusion are stated as fact and are not sourced anywhere in References.

Specifically "The endpoints controller reconciles core `Endpoints` for selector-based Services without owner references (copying the Service's labels, including `app.kubernetes.io/managed-by: Helm`, but not its annotations)", and "The EndpointSlice controller normally creates slices with a Service controller owner reference."

The first is doing real work. The copied label with no annotations is what makes the object unclassifiable under the rule at 99, which is what would block recovery on every cluster, which is the entire justification for excluding the kind and accepting the loss. References covers `NamespaceLifecycle`, CRD deletion, RBAC, `ValidatingAdmissionPolicy`, the Helm CRD lifecycle, and Helm `install.go`, and has nothing for either claim.

Both check out, for what it is worth. `OwnerReferences` does not appear in `pkg/controller/endpoint/endpoints_controller.go`, and the slice owner reference is `metav1.NewControllerRef(service, gvk)` in `staging/src/k8s.io/endpointslice/utils.go`. A References entry would let the next reader confirm that without re-deriving it.

ndipebot (approved):
Fourth pass, against `4394c25`. All five of my items from the `04cfd5d` round are closed, and I checked each against the code rather than against the evidence links.

1. Trainer uninstall propagation. Line 91 now states this as a change rather than as existing behavior, and it describes today's code correctly: `uninstallTrainerHelmRelease` warns and returns `nil` at `pkg/setup/helm.go:294-297`. Implementation 130 no longer says "keeps the existing", Validation 144 and 146 match it, and the Rationale bullet at 180, the Alternatives entry at 203, and the Consequence at 192 all landed. The contradiction between 130 and 144 is gone.
2. EndpointSlice justification. The universal ownerless-and-rebuilt rule is gone from both 93 and 98, replaced by per-kind statements that are correct for all three kinds, and the endpoint slice controller is now named as a creator. The bar handed to future kinds is a fixture plus a per-kind rationale. The exclusions themselves are unchanged, which is what I asked for.
3. Retained output. Line 89 and Validation 145 cover skipped helm and name the `skip-helm-omits-nvcre-entries` differential case. The two items I mentioned in passing also landed: `[helm] Skipped.` under `--auto-approve`, and skip-aware `asymmetricChartRefsWarning`, at 128 and 151.
4. The sentence at 71 parses again.
5. ADR-073 now reads "decisions 1, 3(a), 3(c), 4, and 5" against five bullets, and the `Amended by` header and the References entry both name the release-state refusal. The diff is 15 insertions and 0 deletions, so the historical record is annotated rather than rewritten.

Approving. Four non-blocking comments inline. The one on decision 1 is the only one I would want looked at before implementation starts rather than during it: I think the `crd-only` arm is reachable on a cluster that has a live external controller, and the fix reads like one sentence.

/lgtm
/approve

docs/designs/078-jobset-ownership.md
kaynetu: will address in implementation

Issue #252, opened 2026-09-01
[Feature]: Emit Normal events on phase transitions across all six reconcilers


ndipebot:
Follow-up to #150. Failure-path Warning events now exist at every tier: a331c85 wired the Job and Workflow reconcilers, and #250 wired Certification, GoodputMeasurement, BandwidthMeasurement, and WorkloadRun. What remains from #150 is the other half: events on key state transitions (InProgress, Succeeded, Failed) so `kubectl describe` tells the lifecycle story without controller log access.

This needs real design, which is why it was split out:

- Transition events must fire once per actual transition, not once per 15s requeue. Every reconciler funnels condition changes through `setExclusiveCondition` (`pkg/controller/status.go`), which already knows whether the condition actually changed, so that is the natural hook for dedup.
- The integration harness (`cmd/integration/integration_test.go`) currently wires a recorder only for the Workflow reconciler; asserting on emitted events would need harness support and golden coverage.
- Decide which transitions matter per tier. A Normal event per phase at six tiers is roughly 18 sites; some (measurement Complete, Job iteration restarts) may not be worth the etcd traffic.

Events RBAC is already in place, and every reconciler now has a nil-safe event helper, so this is purely about placement and dedup.

Also worth picking up here: the two Certification build/validation failure paths before Workflow creation (around `certification_controller.go` lines 188 and 252) still emit no Warning event. Both go terminal via `setCertificationFailed`, so a one-shot event there is safe and could reuse `ReasonWorkflowValidationFailed`.

Issue #150, opened 2026-08-12
[Bug]: Reconcilers do not emit Kubernetes Events — failures and transitions are invisible without log access


ndipebot:
### Prerequisites

- [x] I searched existing issues
- [x] I can reproduce this issue

### Code of Conduct

- [x] I agree to follow the project's Code of Conduct

### Bug Description
None of the six CRE reconcilers (Job, Workflow, Certification, GoodputMeasurement, BandwidthMeasurement, WorkloadRun) configure an `EventRecorder` in their `struct` definitions or call `r.Recorder.Event(...)` on state transitions. As a result, `kubectl describe certification` and `kubectl describe job` show no Events section. Operators investigating failures must access controller logs directly rather than the standard Kubernetes diagnostic interface.

**Location:** All reconciler structs in `pkg/controller/`

**Suggested fix:** Add an `EventRecorder` field to each reconciler struct, inject it via `mgr.GetEventRecorderFor("cre-<reconciler>")` in `SetupWithManager`, and emit `Warning`/`Normal` events on key transitions (InProgress→Failed, Job timeout, dependency cleanup failures).

#

### Component

Controller / Reconcilers

### Steps to Reproduce
1. Run a Certification and observe a failure.
2. Run `kubectl describe certification <name>` — the Events section is empty.
3. The only diagnostic information is in controller pod logs.

#### Environment
- CRE version: all
- Kubernetes version: all
- Deployment method: all

Issue #67, opened 2026-08-08
ValidationFailed can never reach a WorkloadRun: the Workflow condition it mirrors is never set


dims:
## Summary

`WorkloadRunValidationFailed` can never be set. The condition type is declared, the code that would mirror it exists, but the Workflow condition it reads is never written by anything.

## The chain

The WorkloadRun controller mirrors validation failures from its Workflow:

```go
// pkg/controller/workloadrun_controller.go:143-152
// Mirror validation failed (independent condition).
if condIsTrue(workflow.Status.Conditions, burninv1alpha1.WorkflowValidationFailed) {
	msg := condMsg(workflow.Status.Conditions, burninv1alpha1.WorkflowValidationFailed)
	meta.SetStatusCondition(&run.Status.Conditions, metav1.Condition{
		Type:    burninv1alpha1.WorkloadRunValidationFailed,
		...
	})
}
```

`WorkflowValidationFailed` is declared at `api/v1alpha1/workflow_types.go:29`. **Nothing anywhere in the tree ever sets a Workflow condition of that type.** The Workflow controller sets exactly three:

```
workflow_controller.go:1851   WorkflowInProgress
workflow_controller.go:1856   WorkflowSucceeded
workflow_controller.go:1861   WorkflowFailed
```

So the `condIsTrue` above is permanently false.

**A second, independent reason the same code is unreachable.** A threshold violation sets `JobValidationFailed` on the Job, and `hasValidationFailures` (`:1653`) turns that into a Workflow **`Failed`** with reason `JobValidationFailed` (`:1607`). The WorkloadRun controller checks `WorkflowFailed` *first* and returns:

```go
// :137-141
if condIsTrue(workflow.Status.Conditions, burninv1alpha1.WorkflowFailed) {
	r.setWorkloadRunCondition(run, burninv1alpha1.WorkloadRunFailed, ReasonWorkflowFailed, msg)
	return ctrl.Result{}, r.Status().Update(ctx, run)
}
```

So even if the condition type were written, the mirror below would not be reached in the case it exists for.

## Observed behaviour

On hardware, a WorkloadRun with an unmeetable threshold reports `Failed=WorkflowFailed`. A consumer looking for `ValidationFailed` to tell "the run worked but missed its numbers" apart from "the run broke" cannot distinguish them.

## What I am unsure about — hence an issue rather than a PR

Three ways to resolve it, and the choice is a maintainer call:

[...]

PR #338, opened 2026-09-10
docs: design phase transition events in ADR-080

kaynetu:
## Summary

Add proposed ADR-080 for phase-transition events so `kubectl describe` can explain resource lifecycles without controller logs. The design covers four lifecycle tiers, independent Job verdicts, Workflow-driven Job timeouts, and both Certification validation failures. It specifies retry-aware duplicate suppression, emission after successful status writes, and best-effort delivery.

Define recorder wiring, event collection, and tests that distinguish transition duplicates from legitimate repeated action events. Add the ADR to the design index. This PR contains the design only; controller implementation remains separate.

## Related Issue

Refs #252. Keep the issue open for implementation.

## Type of Change
- [x] 📚 Documentation

## Component(s) Affected
- [x] Documentation / CI

## Testing
- [x] Tests pass locally
- [x] Manual testing completed: reviewed the ADR against issue #252 and current controller status-write paths.
- [x] No breaking changes

Validation: `make manifests generate`, `make lint-fix`, `make build`, `make test`, `make verify-doc-links`, `make check-agents-sync`, and `git diff --check`.

## Risk Assessment

Documentation-only change; no runtime, API, chart, or golden-file changes. The ADR explicitly leaves the existing timeout-condition exclusivity behavior unchanged and documents that events are best-effort diagnostics.

## Checklist
- [x] Self-review completed
- [ ] Golden files updated (not applicable; no integration output changes)
- [x] Documentation updated
- [x] Ready for review

### DIFF
--- a/docs/designs/080-phase-transition-events.md
+++ b/docs/designs/080-phase-transition-events.md
@@ -124,11 +124,13 @@ Both facts drive the scope decision below.
    no flip and emits nothing. Detection is not done from the caller's stale
    copy.
 
-3. **Emit after the write succeeds, never on a failed write.** The event is
+3. **Emit transition events after the write succeeds, never on a failed write.** The event is
    emitted by the tier wrapper after `setExclusiveStatusCondition` returns
    `nil`, using the transition it reports. A write that exhausts conflict
-   retries or fails for any other reason emits no event: an event must never
-   claim a phase the object does not have.
+   retries or fails for any other reason emits no transition event: an event
+   must never claim a phase the object does not have. This does not prohibit
+   action-failure Warnings that describe an observed error rather than a
+   persisted phase; decision 5 preserves those diagnostics on failed writes.
 
    This makes events **best-effort notification with transition-based
    duplicate suppression**, not an exactly-once record. A controller crash in
@@ -206,7 +208,8 @@ Both facts drive the scope decision below.
    hand-placed Warning immediately followed by a Failed transition with the
    same reason is not two rows but one `Event` with `count: 2`, which the
    single-failure cases in this design forbid. Where that happens, the transition owns the
-   notification and the hand-placed emission is removed. Three sites match
+   notification on success, and the unconditional pre-write emission becomes
+   a fallback on status-write failure. Three sites match
    today:
 
    - WorkloadRun `BuildFailed`: `warnf` followed by
@@ -215,6 +218,24 @@ Both facts drive the scope decision below.
      `setWorkflowFailed("HeterogeneousPlatform")`.
    - Workflow `OverrideError`: `eventf(Warning)` followed by
      `setWorkflowFailed("OverrideError")`.
+
+   At each of these three sites, first attempt the Failed status write. If it
+   succeeds, only the transition hook emits; no fallback is emitted. If the
+   status operation returns an error, emit one action Warning for that failed
+   operation, after any internal retries have finished, using a distinct
+   reason: `BuildFailedStatusUpdateFailed`,
+   `HeterogeneousPlatformStatusUpdateFailed`, or
+   `OverrideErrorStatusUpdateFailed`, respectively. Its message includes the
+   original build/platform/override error and the status-update error, and
+   states that recording the Failed condition was unsuccessful. It must not
+   assert that the resource entered Failed. Preserve existing error returns
+   and logging. The fallback uses the same object and nil-safe recorder.
+
+   A later successful reconcile can emit the normal Failed-transition Warning
+   with the original reason; the distinct fallback reason prevents those two
+   facts from being aggregated together. Repeated failed status operations
+   may repeat the fallback, just as retained action-failure Warnings may
+   repeat. Fallbacks are outside the transition count and volume guarantees.
 
    Where the reasons differ, both stay, because they describe two facts:
    Certification `WorkflowCreationError` (the Create that was rejected)
@@ -328,11 +349,15 @@ Both facts drive the scope decision below.
     before `meta.SetStatusCondition`, and after `Status().Update` returns
     `nil` emit `Warning / JobTimedOut` regarding the Job when it was not.
     Leave the write itself, including its effect on `InProgress`, unchanged.
-  - Remove the hand-placed `eventf(Warning)` at the `HeterogeneousPlatform`
-    and `OverrideError` sites; the Failed transition now emits them.
+  - Replace the unconditional `eventf(Warning)` at the `HeterogeneousPlatform`
+    and `OverrideError` sites with decision 5's fallback in the status-error
+    branch. Add their distinct fallback reason constants. Successful writes
+    notify through the Failed-transition hook only.
 - `pkg/controller/workloadrun_controller.go`
-  - Remove the hand-placed `warnf(BuildFailed)`; the Failed transition now
-    emits it with the same reason and message.
+  - Replace the unconditional `warnf(BuildFailed)` with decision 5's fallback
+    when the build guard's status update fails, using the new
+    `BuildFailedStatusUpdateFailed` reason constant. On success, the Failed
+    transition emits with the original reason and message.
   - At each call site that runs `setWorkloadRunCondition` then
     `Status().Update`, capture the previously-true execution type before the
     mutation and emit after a successful update when it differs. Prefer one
@@ -368,11 +393,18 @@ Both facts drive the scope decision below.
   reconciles produces one recorded event; a reason-then-message change within
   the phase produces zero further events; and a flip to a new phase produces
   exactly one more.
-- Failed-write tests using the existing `interceptor.Funcs{SubResourceUpdate}`
+- Failed-write tests of the transition hooks using the existing `interceptor.Funcs{SubResourceUpdate}`
   pattern in `status_test.go`: a status update that returns a non-conflict
   error emits nothing; a conflict sequence that exhausts `retry.DefaultRetry`
   emits nothing; a conflict followed by success emits exactly once. These pin
-  the "no event on failed write" rule from decision 3.
+  the "no transition event on failed write" rule from decision 3.
+- Recorder-level tests for each of decision 5's three fallback sites:
+  successful status persistence emits one transition and no fallback; a
+  non-conflict status error emits one fallback and no transition; Workflow
+  conflict exhaustion emits one fallback after retries, not one per attempt.
+  A failed reconcile followed by successful persistence emits the fallback
+  and then one transition under distinct reasons. Assert that fallback
+  messages contain both errors and do not claim a persisted Failed phase.
 - Nil-recorder pins for any new or generalized event helper, mirroring
   `TestJobWarnfNilRecorder`.
 - A recorder-level case applying multiple overrides asserts one retained
@@ -410,8 +442,9 @@ Both facts drive the scope decision below.
     Warning and the outcome Warning appear with their distinct reasons;
   - the three same-reason sites from decision 5 (WorkloadRun `BuildFailed`,
     Workflow `HeterogeneousPlatform`, Workflow `OverrideError`), each
-    asserting a single row with `count: 1` after the hand-placed emission is
-    removed. A `count: 2` here is the regression these tests exist to catch.
+    asserting a single transition row with `count: 1` and no fallback after
+    successful status persistence. A `count: 2` here is the regression these
+    tests exist to catch.
 - Existing goldens must show zero diffs, since event collection is opt-in
   through the `events` list.
   Any diff in a case that did not opt in is a defect, not a regeneration
@@ -435,8 +468,10 @@ maintainer approval.
   set; comparing which member is `True` before and after is the only signal
   that means "phase changed". Reading it inside the callback is what makes the
   conflict-retry path correct without a second read.
-- **Emit-after-write is the safety invariant.** Events are observational; an
-  event that precedes or survives a failed write lies about the object.
+- **Emit-after-write is the transition safety invariant.** A transition event
+  must describe a persisted phase. An action fallback instead describes the
+  observed failure and unsuccessful status update, preserving diagnostics
+  without claiming that the phase changed.
 - **Failed as Warning matches how operators filter.** `kubectl get events
   --field-selector type=Warning` is the standard first cut. Burying a terminal
   failure under Normal would make the events feature miss the case #150 was
@@ -465,7 +500,7 @@ maintainer approval.
   InProgress, and any additional actual phase flips add emissions. The
   timeout hook supplies the Job's failure event, not an extra event on top
   of it. This estimate excludes retained action and informational events,
-  which can repeat, and is not an estimate of API writes or aggregated Event
+  including status-error fallbacks, which can repeat, and is not an estimate of API writes or aggregated Event
   objects. Events expire according to the API server's TTL.
 - `kubectl describe` on a Job, Workflow, Certification, or WorkloadRun now
   shows its lifecycle in the Events section. Measurement objects continue to
@@ -476,8 +511,9 @@ maintainer approval.
   needs to read the reason.
 - Where an action Warning precedes a Failed transition with a different
   reason, two Warning events result. This is documented, not deduplicated.
-  Where the reasons were identical, the hand-placed emission is removed at
-  three sites and the transition emits instead. At WorkloadRun `BuildFailed`
+  Where the reasons were identical, the unconditional emission is replaced at
+  three sites: the transition emits on success, and a distinct action
+  fallback emits if status persistence fails. On success, at WorkloadRun `BuildFailed`
   and Workflow `HeterogeneousPlatform` the message is unchanged. At Workflow
   `OverrideError` the message changes from `Override failed: …` to the
   condition's `Failed to apply overrides: …`; the reason is the same, and the
@@ -550,9 +586,12 @@ reason and object (and at two of them the message as well). The recorder
 correlates on reason, not message, so each pair collapses into one `Event`
 with `count: 2`, which would fail the single-failure case's expected `count: 1`
 and would show operators an inflated count for a single failure. Ownership
-moves to the transition. The reason an operator filters on is unchanged at
-all three sites; the message text changes only at `OverrideError`, to the
-wording the condition already carries.
+moves to the transition on successful status persistence, while a distinct
+fallback Warning preserves the observed error on failed persistence. On
+success the reason an operator filters on is unchanged at all three sites;
+the message text changes only at `OverrideError`, to the wording the
+condition already carries. Removing these action diagnostics entirely would
+lose visibility when the status write fails.
 
 ### Route the Workflow timeout write through the shared exclusive helper
 
--- a/docs/designs/080-phase-transition-events.md
+++ b/docs/designs/080-phase-transition-events.md
@@ -6,8 +6,9 @@
 
 Issue #150 asked for Kubernetes Events so that `kubectl describe` explains a
 resource's history without controller log access. It landed in two halves.
-Commit `a331c85` wired an `EventRecorder` into the Job and Workflow
-reconcilers, and #250 wired the remaining four (Certification,
+The Workflow reconciler already had an `EventRecorder`; commit `a331c85`
+populated the Job reconciler's, which `main.go` had left nil, and #250 wired
+the remaining four (Certification,
 GoodputMeasurement, BandwidthMeasurement, WorkloadRun) and gave each a nil-safe
 `warnf` helper. Both changes emit only **Warning events tied to a failed action
 on the current reconcile pass**: a child Create rejected by the API server, an
@@ -17,9 +18,11 @@ recurs only when it fails again.
 
 Issue #252 is the deliberately split-off other half: events on the phase
 transitions themselves (InProgress, Succeeded, Failed). #250 excluded them
-because they need dedup logic against the poll. Every reconciler requeues on a
-configurable interval (15s in production), and a naive event at the status
-write would fire once per pass, not once per transition.
+because they need dedup logic against the poll. Every reconciler requeues:
+Job, Workflow, and Certification on a configurable interval (15s in
+production), WorkloadRun on a fixed 15s package constant, and
+GoodputMeasurement on its 60s default sample interval. A naive event at the
+status write would fire once per pass, not once per transition.
 
 Events RBAC and every recorder are already in place. The remaining problem is
 placement and dedup. Two facts about the code shape the design.
@@ -28,7 +31,7 @@ placement and dedup. Two facts about the code shape the design.
 `setExclusiveStatusCondition` ([status.go](../../pkg/controller/status.go))
 returns `true` whenever any condition in the exclusive set differs after the
 write: a reason or message update within the same phase (Certification
-`WaitingForNodes` to `Running`, a progress message rewritten each pass), an
+`WaitingForNodes` to `WorkflowCreated`, a progress message rewritten each pass), an
 `ObservedGeneration` bump after a spec edit, or a caller-supplied `extra`
 mutation. Emitting on `changed` would produce a Normal event on every message
 refresh. Dedup has to key on which condition type flipped to `True`.
@@ -58,6 +61,12 @@ Both facts drive the scope decision below.
    phases. Exclude the two measurement tiers.** Job, Workflow, Certification,
    and WorkloadRun each emit at most one event when their exclusive condition
    set flips to `InProgress`, `Succeeded`, or `Failed`. Twelve sites.
+
+   This amends the scope in #252's title, "across all six reconcilers": the
+   two measurement reconcilers are deliberately excluded for the reasons
+   below, and an implementation matching this record should not be bounced
+   for not matching the issue's title. The trigger to revisit is a future
+   `Failed` condition on a measurement CRD.
 
    GoodputMeasurement and BandwidthMeasurement emit no transition events. Their
    CRDs define only `Measuring` and `Complete`; there is no `Failed` phase to
@@ -109,6 +118,14 @@ Both facts drive the scope decision below.
    callback of `setJobHardwareFailed` and `setJobValidationStatus`, emit after
    the write succeeds, never on a failed write. With the Workflow-driven
    timeout write from decision 3, sixteen sites in total.
+
+   WorkloadRun's additive `ValidationFailed` condition is **status-only** and
+   gets no event. It is a mirror of the Workflow's `ValidationFailed`, which
+   is itself a roll-up of the Job verdict, so the same threshold finding
+   already has a Warning on the Job (`ThresholdViolation`) and a Failed
+   transition on the Workflow (`JobValidationFailed`); a third copy on the
+   WorkloadRun would add a row without adding information. The WorkloadRun's
+   own `Failed` transition is what its Events section shows.
 
 2. **Dedup on the true-type flip, detected inside the write callback.** A
    transition is "the condition type that is `True` after the write differs
@@ -192,22 +209,55 @@ Both facts drive the scope decision below.
    the Events section as in the Conditions block, and the existing tier-prefixed
    reason vocabulary is reused rather than doubled.
 
-5. **The two Certification pre-Workflow validation failures are covered by the
-   Failed transition; no separate site is added.** #252 also asks for a Warning
-   on the two build/validation failure paths before Workflow creation
-   ([certification_controller.go](../../pkg/controller/certification_controller.go),
-   the two `setCertificationFailed(..., ReasonWorkflowValidationFailed, ...)`
-   calls). Both go terminal through `setCertificationFailed`, so under
-   decision 4 they now emit `Warning / WorkflowValidationFailed` with the
-   validation error as the message, at most once. Adding a hand-placed `warnf`
-   there as well would produce a redundant pair with the same reason.
+5. **The Certification catch-alls are covered by the Failed transition, once
+   they stop overwriting a more specific reason.** #252 also asks for a
+   Warning on the two `setCertificationFailed(..., ReasonWorkflowValidationFailed, ...)`
+   calls in
+   [certification_controller.go](../../pkg/controller/certification_controller.go).
+   They are not dedicated validation paths; they are generic catch-alls
+   around the whole `createWorkflowForCategory` step. Today that matters: on
+   a rejected Workflow `Create`, `createWorkflowForCategory` writes
+   `Failed / WorkflowFailed` and returns the error, and the caller's catch-all
+   then rewrites the same condition to `Failed / WorkflowValidationFailed`.
+   Under emit-on-flip the second write is a reason-only change and emits
+   nothing, so Events would say `WorkflowFailed` while Conditions settle on
+   `WorkflowValidationFailed`, breaking decision 4's same-words promise.
+
+   **Prerequisite code change:** the catch-alls must not clobber a terminal
+   reason already written on this pass. `createWorkflowForCategory` returns a
+   typed Create-rejected error carrying the original Create error and any
+   failure from its `setCertificationFailed` call, preserving both causes for
+   error inspection. Both callers recognize it before the generic catch-all,
+   skip the second Failed write, and return that error unchanged. A failed
+   status write therefore remains an error-driven retry, not terminal success;
+   it emits no transition event, while the existing `WorkflowCreationError`
+   action Warning remains. When the inner status write succeeds, the persisted
+   reason is `WorkflowFailed`, and a later reconcile sees the terminal status.
+   The typed error identifies the Create-rejected path, not proof of successful
+   status persistence.
+
+   The existing collision handler is a precedent for bypassing the generic
+   catch-all, but writes its own collision-specific Failed condition. Other
+   errors still reach the generic `WorkflowValidationFailed` write, including
+   build/validation errors, failed reads of an existing Workflow, and a
+   terminating foreign Workflow. Reclassifying those remaining errors is out
+   of scope. Both failure paths emit through the Failed transition only after
+   successful persistence. Adding a hand-placed `warnf` at the catch-alls would
+   produce a same-reason pair.
 
    The existing action-failure Warnings are kept **unless they share a
-   reason with the Failed transition on the same object.** The recorder
-   correlates on `(type, reason, regarding)`, not on message, so a
+   reason with the Failed transition on the same object.** The `events/v1`
+   recorder correlates on `(type, action, reason, reportingController,
+   reportingInstance, regarding, related)`, and the message is not part of
+   the key. Every emit helper in this repository passes the reason as the
+   `action` argument (`Eventf(obj, nil, eventType, reason, reason, ...)`),
+   and **this record makes that an explicit constraint**: the transition hook
+   and every fallback pass `action == reason` too. Under that constraint a
    hand-placed Warning immediately followed by a Failed transition with the
    same reason is not two rows but one `Event` with `count: 2`, which the
-   single-failure cases in this design forbid. Where that happens, the transition owns the
+   single-failure cases in this design forbid. If any emitter passed a
+   different action, the pair would split into two rows and this analysis
+   would not hold. Where that happens, the transition owns the
    notification on success, and the unconditional pre-write emission becomes
    a fallback on status-write failure. Three sites match
    today:
@@ -217,10 +267,17 @@ Both facts drive the scope decision below.
    - Workflow `HeterogeneousPlatform`: `eventf(Warning)` followed by
      `setWorkflowFailed("HeterogeneousPlatform")`.
    - Workflow `OverrideError`: `eventf(Warning)` followed by
-     `setWorkflowFailed("OverrideError")`.
+     `setWorkflowFailed("OverrideError")`. A second `OverrideError` site,
+     the early `applyOverrides` guard in `reconcileJob`, writes
+     `Failed / OverrideError` with no hand-placed emission. It is covered by
+     the same rule so one reason is treated evenly: the transition Warning on
+     success, and the same `OverrideErrorStatusUpdateFailed` fallback when its
+     status write fails.
 
-   At each of these three sites, first attempt the Failed status write. If it
-   succeeds, only the transition hook emits; no fallback is emitted. If the
+   There are three existing dual-emission sites and four fallback sites,
+   counting the early `OverrideError` guard. At all four fallback sites, first
+   attempt the Failed status write. If it succeeds, only the transition hook
+   emits; no fallback is emitted. If the
    status operation returns an error, emit one action Warning for that failed
    operation, after any internal retries have finished, using a distinct
    reason: `BuildFailedStatusUpdateFailed`,
@@ -239,11 +296,15 @@ Both facts drive the scope decision below.
 
    Where the reasons differ, both stay, because they describe two facts:
    Certification `WorkflowCreationError` (the Create that was rejected)
-   followed by `Failed / WorkflowFailed` (the outcome). Where an action
-   Warning is followed by a returned error and no Failed write (Job and
-   WorkloadRun `*CreationError`), or is informational with no phase change
-   (Workflow `CordonedNodesExcluded`, `HeterogeneousGPU`,
-   `InsufficientGPUCapacity`), nothing changes.
+   followed by `Failed / WorkflowFailed` (the outcome), and Workflow
+   `InsufficientGPUCapacity` followed by `Failed / PartitionError` when every
+   surviving node is capacity-excluded. Nothing changes at the remaining
+   action Warnings: Job `WorkloadCreationError` and WorkloadRun
+   `WorkflowCreationError` return an error with no Failed write; Job
+   `MeasurementCreationError` is deliberately non-fatal and the reconcile
+   continues to the workload status update; and Workflow
+   `CordonedNodesExcluded` and `HeterogeneousGPU` are Warning-typed
+   exclusion notices with no Failed write of their own.
 
 6. **Integration coverage asserts real Event objects, opt-in per case.** The
    integration harness runs envtest, a real API server, so emitted events are
@@ -257,18 +318,32 @@ Both facts drive the scope decision below.
    sorted by the first five. Event names, UIDs, timestamps, source, and
    reporting instance are omitted; they are not stable across runs.
 
-   `count` is load-bearing. The `events/v1` recorder correlates events on
-   `(type, action, reason, reportingController, reportingInstance,
-   regarding)`; the message is **not** part of that key. Repeated emissions
-   for the same object and reason therefore do not appear as extra rows: they
-   collapse into one `Event` whose `series.count` increments. A projection
-   without `count` cannot distinguish one emission from fifty. The collector
-   emits `count` as `series.count` when a series exists and `1` otherwise.
-   For each new phase or verdict event whose case drives one qualifying
-   transition, the expected count is `1`; a higher count is a dedup defect,
-   not a regeneration candidate. This rule does not apply to retained action
-   or informational events: repeated failed actions or multiple applied
-   overrides may legitimately produce a count above `1`.
+   `count` is load-bearing but coarse. The `events/v1` recorder correlates
+   on `(type, action, reason, reportingController, reportingInstance,
+   regarding, related)`; the message is **not** part of that key, and
+   `action == reason` by the decision 5 constraint. Repeated emissions for
+   the same object and reason therefore collapse into one `Event` rather
+   than new rows, but only the **second** emission promptly patches the API
+   object (`series.count: 2`); the third and later bump an in-memory counter
+   that is flushed by periodic refresh or idle-series finalization. Refresh
+   runs every 30 minutes; cleanup runs every 6 minutes and finalizes series
+   idle for more than 6 minutes. Closure therefore occurs on an eligible
+   cleanup pass, not exactly 6 minutes after the last emission. Before either
+   flush, and after asynchronous delivery, the API-visible `count` provides
+   only a coarse signal: `1` versus at-least-two (`2`), not an exact total of
+   later emissions. After a series is removed from the broadcaster cache, a
+   repeat lands as a fresh row with `count: 1`. The collector emits `count` as
+   `series.count` when a series exists and `1` otherwise. For each new phase
+   or verdict event whose case drives one qualifying transition, the expected
+   result is one row with `count: 1`; a `2` is a dedup defect, not a
+   regeneration candidate. Give opted-in cases a cumulative 3-minute deadline
+   from recorder startup through event collection, with a fresh broadcaster
+   per case, so periodic refresh and idle cleanup cannot change that signal.
+   Each wait uses the smaller of its configured timeout and the remaining
+   budget; fail the case if collection exceeds the deadline. Existing
+   configurable per-wait timeouts do not guarantee this total duration.
+   This rule does not apply to retained
+   action or informational events, which may legitimately show `2`.
 
    Delivery is asynchronous: the recorder hands events to a broadcaster
    goroutine, so observing the status transition does not mean the `Event`
@@ -315,6 +390,19 @@ Both facts drive the scope decision below.
 ## Implementation
 
 - `pkg/controller/status.go`
+  - Add a pure `conditionFlip(before, after, types)` helper returning the
+    watched conditions whose presence or Status changed, including old/new
+    presence and Status and the resulting condition. Ignore reason, message,
+    and ObservedGeneration-only edits. Snapshot the watched values before
+    mutation; do not retain an alias to the mutated condition slice.
+    For exclusive phases, callers select a condition becoming True; for Job
+    verdicts, select absent/False to True for Warnings and absent to False for
+    ThresholdsMet. For timeout, watch only JobFailed becoming True, regardless
+    of InProgress. The helper reports changes; callers retain event policy.
+    Compute fresh results inside each retry callback and replace any previous
+    attempt's result, including clearing it on a no-op retry. WorkloadRun and
+    timeout callers instead snapshot immediately before their direct mutation
+    and compare afterward. Every caller emits only after a successful write.
   - Extend `setExclusiveStatusCondition` to report the transition: the type
     that was `True` before the mutation and the type that is `True` after, read
     inside the mutate callback on the object state the write is computed from.
@@ -340,10 +428,16 @@ Both facts drive the scope decision below.
     condition flipped to `True`, and the `Normal / ThresholdsMet` when
     `ValidationFailed` flipped from absent to `False`. A write that leaves the
     condition's status unchanged emits nothing.
-  - Job and Certification currently expose only `warnf`/`normalf`; add or
-    generalize to an `eventf(obj, eventType, reason, fmt, args...)` matching
-    the Workflow tier so the wrapper can pass the type. Keep all helpers
+  - Job exposes only `warnf`, and Certification `warnf`/`normalf`; add or
+    generalize each to an `eventf(obj, eventType, reason, fmt, args...)`
+    matching the Workflow tier so the wrapper can pass the type. Every helper
+    passes the reason as the `action` argument (decision 5). Keep all helpers
     nil-safe.
+  - Prerequisite from decision 5: `createWorkflowForCategory` returns a typed
+    Create-rejected error and both Certification callers skip their
+    `setCertificationFailed` catch-all for it. Preserve and return both the
+    Create error and any status-write error; `WorkflowFailed` is persisted only
+    when the inner status write succeeds. Do not return success when it fails.
 - `pkg/controller/workflow_controller.go`, `updateStatusFromJobs`
   - At the `timeoutPerJob` write, read whether `JobFailed` is already `True`
     before `meta.SetStatusCondition`, and after `Status().Update` returns
@@ -351,7 +445,9 @@ Both facts drive the scope decision below.
     Leave the write itself, including its effect on `InProgress`, unchanged.
   - Replace the unconditional `eventf(Warning)` at the `HeterogeneousPlatform`
     and `OverrideError` sites with decision 5's fallback in the status-error
-    branch. Add their distinct fallback reason constants. Successful writes
+    branch. Add the same fallback to the early `applyOverrides` guard that
+    currently has no preceding event. Add their distinct fallback reason
+    constants. Successful writes
     notify through the Failed-transition hook only.
 - `pkg/controller/workloadrun_controller.go`
   - Replace the unconditional `warnf(BuildFailed)` with decision 5's fallback
@@ -364,6 +460,9 @@ Both facts drive the scope decision below.
     small helper that wraps the pair so the rule lives in one place.
 - `cmd/integration/integration_test.go`
   - Wire recorders for all six reconcilers.
+  - Enforce decision 6's cumulative 3-minute deadline for opted-in cases from
+    fresh recorder startup through collection, sharing the remaining budget
+    across waits. Leave cases without event collection unchanged.
   - Add an `events` list to `waitConfig` with `involvedKind`, `involvedName`,
     `namespace`, and `expect: [{type, reason}]`. Readiness waits for every
     `expect` row; serialization then emits all rows for the object in the
@@ -380,12 +479,20 @@ Both facts drive the scope decision below.
 
 ### Testing plan
 
-- `pkg/controller/status_test.go` (table tests are the accepted idiom in this
-  file): the transition result for absent to InProgress, InProgress to
-  Succeeded, InProgress to Failed, same phase with a new reason (no
-  transition), same phase with a new message (no transition), `extra`-only
-  mutation (no transition), and a conflict retry where the refetched object
-  already holds the target phase (no transition).
+- Transition-result cases as `testutil.TestCaseParser` goldens under
+  `pkg/controller/testdata/`, since the result is a multi-field structure and
+  `pkg/controller/` is a required-golden package: absent to InProgress,
+  InProgress to Succeeded, InProgress to Failed, same phase with a new reason
+  (no transition), same phase with a new message (no transition),
+  `extra`-only mutation (no transition), and a conflict retry where the
+  refetched object already holds the target phase (no transition). Tables
+  stay only where the convention allows them: nil-recorder pins and the
+  single-value `FakeRecorder` call counts below.
+- Shared-detector goldens cover absent-to-False verdicts, False-to-True
+  verdicts, unchanged verdict status with changed details, and JobFailed
+  becoming True while InProgress remains True. Exercise the direct-write
+  consumers as well as retry callbacks; a failed attempt's result must not
+  survive a no-op retry.
 - Emission tests at the recorder level, using client-go's
   `events.FakeRecorder` injected as the wrapper's `Recorder`, since the golden
   projection observes only the aggregated `Event` and the fake sees every
@@ -398,13 +505,23 @@ Both facts drive the scope decision below.
   error emits nothing; a conflict sequence that exhausts `retry.DefaultRetry`
   emits nothing; a conflict followed by success emits exactly once. These pin
   the "no transition event on failed write" rule from decision 3.
-- Recorder-level tests for each of decision 5's three fallback sites:
-  successful status persistence emits one transition and no fallback; a
+- Recorder-level tests for all four fallback sites: WorkloadRun BuildFailed,
+  Workflow HeterogeneousPlatform, and both Workflow OverrideError guards.
+  Successful status persistence emits one transition and no fallback; a
   non-conflict status error emits one fallback and no transition; Workflow
   conflict exhaustion emits one fallback after retries, not one per attempt.
   A failed reconcile followed by successful persistence emits the fallback
   and then one transition under distinct reasons. Assert that fallback
   messages contain both errors and do not claim a persisted Failed phase.
+- For both Certification callers, inject Create rejection with successful
+  status persistence, a non-conflict status error, and exhausted status
+  conflicts. Assert no generic catch-all overwrite; successful persistence
+  leaves WorkflowFailed and emits its transition, while failure returns the
+  Create and status errors, retains the action Warning, and emits no
+  transition. A retry with another rejected Create and a successful status
+  write emits the transition once.
+- Harness deadline tests cover multiple waits sharing one budget and collection
+  after expiry. An individual wait timeout must not reset the total budget.
 - Nil-recorder pins for any new or generalized event helper, mirroring
   `TestJobWarnfNilRecorder`.
 - A recorder-level case applying multiple overrides asserts one retained
@@ -438,6 +555,13 @@ Both facts drive the scope decision below.
     shows `Normal / WorkloadCompleted` then `Normal / ThresholdsMet`;
   - a checkpoint-restart Job case asserting the restart does not emit a second
     InProgress event for the same Job when the phase does not flip;
+  - the second `OverrideError` site (the early `applyOverrides` guard): the
+    transition Warning on success and the `OverrideErrorStatusUpdateFailed`
+    fallback on a failed status write, matching the `eventf` site;
+  - a Certification rejected-Create case asserting the persisted reason stays
+    `WorkflowFailed` (the catch-all no longer overwrites it) and Events and
+    Conditions agree; and a genuine build error asserting
+    `WorkflowValidationFailed` on both;
   - a Certification `WorkflowCreationError` case asserting both the action
     Warning and the outcome Warning appear with their distinct reasons;
   - the three same-reason sites from decision 5 (WorkloadRun `BuildFailed`,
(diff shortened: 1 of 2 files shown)


docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,674 @@
...
+   and logging. The fallback uses the same object and nil-safe recorder.
+
+   A later successful reconcile can emit the normal Failed-transition Warning
+   with the original reason; the distinct fallback reason prevents those two
+   facts from being aggregated together. Repeated failed status operations
+   may repeat the fallback, just as retained action-failure Warnings may
+   repeat. Fallbacks are outside the transition count and volume guarantees.
+
+   Where the reasons differ, both stay, because they describe two facts:
+   Certification `WorkflowCreationError` (the Create that was rejected)
```
ndipebot, line 241: This example doesn't survive the code as it stands. `createWorkflowForCategory` writes `Failed / WorkflowFailed` and returns the error (`certification_controller.go:609`), but both callers' catch-alls then call `setCertificationFailed(..., ReasonWorkflowValidationFailed, err.Error())` (`certification_controller.go:204`, `:282`), and `meta.SetStatusCondition` rewrites the reason. So after a rejected Workflow `Create` the persisted condition reason is `WorkflowValidationFailed`, while under emit-on-flip the only transition event carries `WorkflowFailed`. That breaks decision 4's same-words-in-Events-and-Conditions promise on this exact path, and the integration case in the test plan asserts an outcome the controller never leaves in place.

Decision 5's framing of those two calls as 'the two build/validation failure paths before Workflow creation' has the same root cause: they're generic catch-alls around the whole ensure step, not dedicated validation paths. I'd fix the model first (either make the catch-alls stop clobbering a more specific terminal reason, called out as a prerequisite code change, or describe the overwrite honestly), then rework this example and the planned test on top of it.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,674 @@
...
+   ([certification_controller.go](../../pkg/controller/certification_controller.go),
+   the two `setCertificationFailed(..., ReasonWorkflowValidationFailed, ...)`
+   calls). Both go terminal through `setCertificationFailed`, so under
+   decision 4 they now emit `Warning / WorkflowValidationFailed` with the
+   validation error as the message, at most once. Adding a hand-placed `warnf`
+   there as well would produce a redundant pair with the same reason.
+
+   The existing action-failure Warnings are kept **unless they share a
+   reason with the Failed transition on the same object.** The recorder
+   correlates on `(type, reason, regarding)`, not on message, so a
```
ndipebot, line 207: The correlation key here is wrong, and the design silently leans on the difference. events/v1 correlates on `eventType`, `action`, `reason`, `reportingController`, `reportingInstance`, `regarding`, and `related` (client-go v0.36.4, `tools/events/event_broadcaster.go`), not `(type, reason, regarding)`. The collapse into one row with `count: 2` only happens today because all eight helpers pass the reason as the `action` argument (`Eventf(obj, nil, eventType, reason, reason, ...)`). If the new transition hook passes any other action string, say `PhaseTransition`, the same-reason pairs become two rows, and both this analysis and the `count: 2` regression signature in the test plan stop holding.

The `action == reason` convention needs to be stated as an explicit constraint on the implementation, and the key corrected both here and in the fuller statement under the dedup discussion below, which also omits `related`.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,674 @@
...
+   `message`, `involvedObject.kind`, `involvedObject.name`, and `count`,
+   sorted by the first five. Event names, UIDs, timestamps, source, and
+   reporting instance are omitted; they are not stable across runs.
+
+   `count` is load-bearing. The `events/v1` recorder correlates events on
+   `(type, action, reason, reportingController, reportingInstance,
+   regarding)`; the message is **not** part of that key. Repeated emissions
+   for the same object and reason therefore do not appear as extra rows: they
+   collapse into one `Event` whose `series.count` increments. A projection
+   without `count` cannot distinguish one emission from fifty. The collector
```
ndipebot, line 265: This overstates what `count` can show. In the events/v1 client only the second isomorphic emission promptly patches the API object (`Series{Count: 2}`); the third and later emissions just bump the in-memory cache (`event_broadcaster.go:179-191`, `Series.Count++; return nil`), which is only flushed every 30 minutes, or 6 minutes after the series goes idle. Inside an integration-test window the API-visible count distinguishes one emission from at-least-two, and nothing finer; it will never show fifty. The collector and the fixture constraints built on `series.count` are resting on a weaker signal than this sentence suggests, and any golden expecting an exact count above 2 could never pass inside the test window.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,674 @@
...
+GoodputMeasurement, BandwidthMeasurement, WorkloadRun) and gave each a nil-safe
+`warnf` helper. Both changes emit only **Warning events tied to a failed action
+on the current reconcile pass**: a child Create rejected by the API server, an
+exec-framework build guard, an unresolvable LogProfile. Each site is safe
+against the steady-state requeue because the action either does not recur or
+recurs only when it fails again.
+
+Issue #252 is the deliberately split-off other half: events on the phase
+transitions themselves (InProgress, Succeeded, Failed). #250 excluded them
+because they need dedup logic against the poll. Every reconciler requeues on a
```
ndipebot, line 20: This isn't true for two of the six. `GoodputMeasurement` defaults to 60s (`goodputmeasurement_controller.go:35`, `defaultSampleInterval`, with no override in `main.go`), and `WorkloadRun`'s 15s is a package constant (`workloadrun_controller.go:46`), not a configurable field. Anyone sizing event volume or dedup windows from this paragraph starts from the wrong numbers.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,674 @@
...
+   projection for deterministic test fixtures: `type`, `reason`,
+   `message`, `involvedObject.kind`, `involvedObject.name`, and `count`,
+   sorted by the first five. Event names, UIDs, timestamps, source, and
+   reporting instance are omitted; they are not stable across runs.
+
+   `count` is load-bearing. The `events/v1` recorder correlates events on
+   `(type, action, reason, reportingController, reportingInstance,
+   regarding)`; the message is **not** part of that key. Repeated emissions
+   for the same object and reason therefore do not appear as extra rows: they
+   collapse into one `Event` whose `series.count` increments. A projection
```
ndipebot, line 264: 'Do not appear as extra rows' only holds while the series is alive. `finishSeries` drops the cache entry once `LastObservedTime` is more than 6 minutes stale (`event_broadcaster.go:145-160`), and event names carry a `UnixNano` suffix, so a retained Warning that recurs after the series closes creates a second row with `count: 1`. Two consequences for later sections: a count above 1 is not a reliable signature of repetition (the repeat may land as a fresh row instead), and since the collector serializes all rows and an unexpected extra row fails the golden, an opted-in case whose reconcile spans that window can fail on a legitimate repeat the design says cannot appear.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,674 @@
...
+Rejected. envtest is a real API server; listing `Event` objects tests the
+production path, exercises the recorder's aggregation, and requires no
+test-only interface on the reconcilers. Nondeterministic fields are handled by
+projection rather than by faking the sink.
+
+### Rely on the recorder's built-in aggregation for dedup
+
+Rejected. The events library aggregates identical events into a `count` bump
+within a window; it does not suppress them, and the aggregation window is
+shorter than a long-running InProgress phase. Dedup must be a property of the
```
ndipebot, line 638: The mechanism described here is wrong, even though the rejection stands. The aggregation window isn't fixed: `finishSeries` only closes a series when `LastObservedTime` is more than 6 minutes old, and every isomorphic emission refreshes it (`event_broadcaster.go:152`, `:182`). With a 15s requeue the series never closes, so naive per-pass emission produces one row with a growing count, not periodic new rows; the window is never 'shorter than a long-running InProgress phase'. Count inflation alone is still a good reason to reject this alternative, but the stated mechanism is the part a reader would rely on elsewhere.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,674 @@
...
+   A later successful reconcile can emit the normal Failed-transition Warning
+   with the original reason; the distinct fallback reason prevents those two
+   facts from being aggregated together. Repeated failed status operations
+   may repeat the fallback, just as retained action-failure Warnings may
+   repeat. Fallbacks are outside the transition count and volume guarantees.
+
+   Where the reasons differ, both stay, because they describe two facts:
+   Certification `WorkflowCreationError` (the Create that was rejected)
+   followed by `Failed / WorkflowFailed` (the outcome). Where an action
+   Warning is followed by a returned error and no Failed write (Job and
```
ndipebot, line 243: 'Followed by a returned error' is not accurate for `MeasurementCreationError`. Those failures are deliberately non-fatal: the reconciler logs, emits the Warning, and continues to `updateStatusFromWorkload` (`job_controller.go:310-326`; the comment there says a measurement problem must not block workload status updates). Only Job `WorkloadCreationError` and WorkloadRun `WorkflowCreationError` behave as described here.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,674 @@
...
+   notification on success, and the unconditional pre-write emission becomes
+   a fallback on status-write failure. Three sites match
+   today:
+
+   - WorkloadRun `BuildFailed`: `warnf` followed by
+     `setWorkloadRunCondition(Failed, BuildFailed)` with the same message.
+   - Workflow `HeterogeneousPlatform`: `eventf(Warning)` followed by
+     `setWorkflowFailed("HeterogeneousPlatform")`.
+   - Workflow `OverrideError`: `eventf(Warning)` followed by
+     `setWorkflowFailed("OverrideError")`.
```
ndipebot, line 220: There are two `setWorkflowFailed("OverrideError")` sites, and this inventory only covers the one with the preceding `eventf`. The early `applyOverrides` guard in `reconcileJob` (`workflow_controller.go:142-147`) writes `Failed / OverrideError` with no hand-placed emission at all. Under this design that site emits the transition Warning on success but gets no `OverrideErrorStatusUpdateFailed` fallback when its status write fails, which applies decision 5's own preserve-diagnostics rule unevenly across two sites sharing one reason. Either add it to the inventory or say explicitly why it's out of scope.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,674 @@
...
+write would fire once per pass, not once per transition.
+
+Events RBAC and every recorder are already in place. The remaining problem is
+placement and dedup. Two facts about the code shape the design.
+
+**`changed` is not a phase transition.** The shared helper
+`setExclusiveStatusCondition` ([status.go](../../pkg/controller/status.go))
+returns `true` whenever any condition in the exclusive set differs after the
+write: a reason or message update within the same phase (Certification
+`WaitingForNodes` to `Running`, a progress message rewritten each pass), an
```
ndipebot, line 31: `Running` isn't a Certification reason. The InProgress reasons are `WaitingForNodes`, `WorkflowCreated`, and `WorkflowRunning` (`pkg/controller/helpers.go:50-52`), and `setCertificationInProgress` is never called with `Running`. The example should name a pair that exists, for instance `WaitingForNodes` to `WorkflowCreated`.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,674 @@
+# ADR-080: Phase Transition Events Across the Lifecycle Tiers
+
+> **Status:** Proposed
+
+## Context
+
+Issue #150 asked for Kubernetes Events so that `kubectl describe` explains a
+resource's history without controller log access. It landed in two halves.
+Commit `a331c85` wired an `EventRecorder` into the Job and Workflow
```
ndipebot, line 9: `a331c85` didn't do this. Its own commit message says `WorkflowReconciler` already had a recorder and the commit only populated the `JobReconciler` recorder that `cmd/manager/main.go` had left nil. The References entry ('Job and Workflow tier events') repeats the misattribution. Worth correcting, since the Context section is the history future readers will trust.

ndipebot (commented):
I went through this in detail and checked its claims against the controller code and the pinned client-go (v0.36.4) sources. The direction is good and a lot of it held up under scrutiny: the retry semantics, the timeout write's terminal guard, `mgr.GetEventRecorder`, the `events.FakeRecorder` choice, and the volume math all check out. I left inline comments where the text contradicts the code or the events library. The three I'd want resolved before this merges:

1. The `Failed / WorkflowFailed` example: that reason never survives the callers' catch-alls, so decision 4's same-words promise and the planned integration case are built on a state the controller doesn't persist.
2. The events/v1 correlation key is wrong as stated, and the `count: 2` analysis silently depends on our helpers passing `reason` as the `action` argument.
3. The `series.count` semantics: only the second emission promptly reaches the API server and a series closes after 6 idle minutes, which affects the collector design, the golden guarantees, and one rejected alternative.

The rest are smaller accuracy fixes, plus one design question about hooking `updateStatusWithRetry` directly.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,674 @@
...
+   facts from being aggregated together. Repeated failed status operations
+   may repeat the fallback, just as retained action-failure Warnings may
+   repeat. Fallbacks are outside the transition count and volume guarantees.
+
+   Where the reasons differ, both stay, because they describe two facts:
+   Certification `WorkflowCreationError` (the Create that was rejected)
+   followed by `Failed / WorkflowFailed` (the outcome). Where an action
+   Warning is followed by a returned error and no Failed write (Job and
+   WorkloadRun `*CreationError`), or is informational with no phase change
+   (Workflow `CordonedNodesExcluded`, `HeterogeneousGPU`,
```
ndipebot, line 245: 'Informational' undersells these three. All of them are emitted as `corev1.EventTypeWarning` (`workflow_controller.go:386`, `:400`, `:442`), and `InsufficientGPUCapacity` can be immediately followed by a terminal `setWorkflowFailed(ReasonPartitionError, ...)` when every surviving node is capacity-excluded (`workflow_controller.go:460`). 'No phase change' is not how that last path actually behaves, and it's the path someone auditing this categorization would care about.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,674 @@
...
+    line and `recordJobStatus` on `changed`.
+  - In `setJobHardwareFailed` and `setJobValidationStatus`, read whether the
+    condition was already `True` inside the mutate callback (in the former,
+    alongside the existing `isFirstFailure` computation, which is keyed on
+    `failedNodes` emptiness rather than on the condition and is left for the
+    metrics it drives). Emit the Warning after a successful write when the
+    condition flipped to `True`, and the `Normal / ThresholdsMet` when
+    `ValidationFailed` flipped from absent to `False`. A write that leaves the
+    condition's status unchanged emits nothing.
+  - Job and Certification currently expose only `warnf`/`normalf`; add or
```
ndipebot, line 343: Job has no `normalf`. `JobReconciler` defines only `warnf` (`job_controller.go:1510`, and its doc comment says every Job-tier event is a warning); only Certification has both helpers (`certification_controller.go:1032`, `:1043`). The bullet's plan still works, it just needs to start from the helper that actually exists on the Job tier.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,674 @@
...
+- Documentation
+  - `docs/operations/troubleshooting.md` (or the closest existing
+    operations page): a short section on reading the Events section of
+    `kubectl describe` per tier, the Normal/Warning split, the one-hour TTL,
+    and that measurement tiers report through their Job.
+  - Add the new page or section to `docs/index.yml` if a new page is created.
+
+### Testing plan
+
+- `pkg/controller/status_test.go` (table tests are the accepted idiom in this
```
ndipebot, line 383: This reads the `CLAUDE.md` exception more broadly than it's written. Tables in `pkg/controller/` are licensed for nil-safety pins, concurrency guards, and single-value assertions, and `status_test.go` is named only as the retry-budget pin (it contains exactly one table test today), not as a file where tables are the idiom. A seven-scenario table returning multi-field transition results looks like exactly the hand-maintained structured-output table the convention rejects. I'd either specify these as `testutil.TestCaseParser` goldens or shrink the assertions to single values so they stay inside the exception. The `FakeRecorder` count tests and the nil pins elsewhere in the plan are fine as tables.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,674 @@
...
+**The six reconcilers do not share one status path.** Three patterns exist:
+
+| Tier | Status pattern |
+|---|---|
+| Job, Workflow, Certification | per-tier `setExclusiveCondition` wrapper over the shared `setExclusiveStatusCondition` |
+| GoodputMeasurement, BandwidthMeasurement | bespoke `setComplete` / raw `meta.SetStatusCondition` with first-terminal-write-wins (ADR-072) |
+| WorkloadRun | in-memory `setWorkloadRunCondition` followed by a direct `Status().Update` |
+
+A single hook is therefore not available; the design has to say which tiers get
+one and why.
```
ndipebot, line 45: I don't think this dismissal holds up, and Alternatives Considered never evaluates the option it dismisses. Twelve of the sixteen emission sites already funnel through `updateStatusWithRetry`: the three tier wrappers and `setJobFailed` reach it via `setExclusiveStatusCondition` (`status.go:69`, which calls it at `:79`), and both verdict setters call it directly (`job_controller.go:825`, `:1096`). A per-condition flip detector at that choke point would collapse most of the per-seam prose in the Implementation section, and if the deferred alternative of routing the Workflow timeout through the shared helper ever lands, it would absorb that special case too. The two detection alternatives the ADR does reject are weaker versions of this idea. If there's a reason the choke-point hook doesn't work, I'd like to see it evaluated and rejected explicitly; five hand-specified seams are a lot to keep in sync during implementation and review.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,674 @@
...
+
+## References
+
+- [Issue #252: Emit Normal events on phase transitions across all six reconcilers](https://github.com/NVIDIA/cluster-readiness-engine/issues/252)
+- [Issue #150: Reconcilers do not emit Kubernetes Events](https://github.com/NVIDIA/cluster-readiness-engine/issues/150)
+- [PR #250: wire event recorders into the four recorder-less reconcilers](https://github.com/NVIDIA/cluster-readiness-engine/pull/250)
+- Commit `a331c85`: Job and Workflow tier events
+- [ADR-072: Freeze GoodputMeasurement Status at Job Terminal State](072-goodput-terminal-freeze.md)
+- [ADR-000: CRD hierarchy](000-adr.md)
+- [Kubernetes Events API and `--event-ttl`](https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/event-v1/)
```
ndipebot, line 671: This link doesn't back the flag: it's the event-v1 API reference, which doesn't mention `--event-ttl`. The flag is documented on the kube-apiserver reference page (https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/). The 1h default claimed earlier in the doc is correct, it just needs the citation that actually contains it.

asivanadi0 (commented):

Read ADR-080 at HEAD `63aee68`, the PR body, issue #252 / #150 context, and the cited controller surfaces on `main` (`pkg/controller/{status,job,workflow,certification,workloadrun,goodputmeasurement}_controller.go`, `cmd/manager/main.go`, `cmd/integration/integration_test.go`, client-go `tools/events`). I also read ndipebot's pass and re-checked the three items they flagged against the code rather than relying on the review text.

Direction is right: flip-not-`changed`, emit-after-successful-write, Job verdicts outside the exclusive set, Workflow-driven `JobTimedOut` as its own site, and decision 5's distinct-reason fallbacks for the same-reason action/transition pairs. Several claims I checked independently held: measurement CRDs really are Measuring/Complete only; production already wires six `mgr.GetEventRecorder`s (harness today only Workflow); the timeout write sets `JobFailed` without clearing `InProgress`; existing `eventf` already passes `reason` as both reason and action.

I am not ready to accept the record as written — concrete gaps below. ndipebot's three merge blockers are real; I add a few that their pass did not cover.

### What looks solid

- Dedup keyed on which exclusive condition is `True`, read inside the mutate callback, is the only correct retry story given `setExclusiveStatusCondition`.
- Keeping measurement tiers out of transition events (Job verdicts are the operator-facing fact) matches the type model and avoids an ADR-072 first-terminal-write interaction for little gain.
- Sixteen-site arithmetic (4 tiers × 3 phases + HardwareFailed + ValidationFailed True/False + JobTimedOut) checks out.
- Opt-in integration `events:` collection so existing goldens stay untouched is the right harness shape.
- Decision 5's fallback reasons (`*StatusUpdateFailed`) are the right answer to CodeRabbit's "preserve a warning when the status write fails" note (already addressed in `63aee68`).

### Concrete gaps before I would accept this ADR

### 1. `Failed / WorkflowFailed` from Create does not survive the callers (ndipebot #1 — verified)

`createWorkflowForCategory` does write `Failed / WorkflowFailed` + `warnf(WorkflowCreationError)` on Create rejection (`certification_controller.go` ~607–612), but both callers' catch-alls immediately re-call `setCertificationFailed(..., ReasonWorkflowValidationFailed, err.Error())` for any non-collision, non-`errNoNodesMatch` error (~204 and ~282). Same reconcile, same exclusive type: under this ADR's flip rule the second write is a reason-only change and emits nothing, so the Events section would say `WorkflowFailed` while Conditions settle on `WorkflowValidationFailed`. Decision 4's "same words in Events and Conditions" and the planned `WorkflowCreationError` + outcome integration case need to name the reason that actually persists (or the Create path needs to stop double-writing). I would not merge the ADR while that example is wrong.

### 2. events/v1 correlation + `series.count` semantics (ndipebot #2/#3 — verified)

Agree with the inline notes. Our helpers pass `reason` as the `action` argument today (`workflow_controller.go` `eventf`: `Eventf(obj, nil, eventType, reason, reason, ...)`), so the written `(type, reason, regarding)` key is incomplete, and the `count: 2` regression story depends on that helper quirk plus the broadcaster's "second emission patches Series; later ones stay in-memory until finishSeries (~6m idle)". Collector / golden / rejected-alternative text should describe the library that ships with the pinned client-go, not an idealized key.

### 3. README index skips ADR-078 and ADR-079

This branch's `docs/designs/README.md` jumps 077 → 080. #336 (ADR-078 JobSet ownership) and #337 (ADR-079 workload labels) are still open. Same class of gap I flagged on #337: either land/coordinate the numbers, or leave an explicit placeholder row so the index does not look like 078/079 were skipped by mistake.

### 4. Issue #252 title vs measurement exclusion — state the scope amendment

#252's title still says "across all six reconcilers". The exclusion rationale is good; the ADR should say out loud that it amends the issue's six-reconciler scope (and that the trigger to revisit is a future measurement `Failed` condition), so a later reader does not bounce the implementation PR for "doesn't match the issue".

### 5. `WorkloadRunValidationFailed` is a silent sibling of the Job verdict story

WorkloadRun already mirrors `WorkflowValidationFailed` into its own additive `ValidationFailed` condition *before* the terminal mirror (`workloadrun_controller.go` ~154–165, issue #67). Decision 1 gives Job `HardwareFailed` / `ValidationFailed` their own events because they sit outside the exclusive set, but says nothing about this WorkloadRun twin. Either add a one-line "WorkloadRunValidationFailed is status-only; the operator-facing verdict events live on the Job / Workflow Failed reason" or give it the same flip-to-True Warning treatment. Leaving it unmentioned will surprise anyone describing a WorkloadRun that missed thresholds.

### 6. Small accuracy nits I independently hit (several overlap ndipebot's inlines)

- Context's "Certification `WaitingForNodes` to `Running`" — `Running` is not a Certification reason; InProgress reasons are `WaitingForNodes` / `WorkflowCreated` / `WorkflowRunning`.
- "Job and Certification currently expose only `warnf`/`normalf`" — Job has only `warnf` (doc comment: every Job-tier event is a warning). Only Certification has both.
- Decision 5's OverrideError inventory covers the `eventf` + `setWorkflowFailed` site; the early `applyOverrides` guard in `reconcileJob` fails with no preceding `eventf`. Transition-only coverage is fine there — say so, so implementers do not hunt a fourth dual-emit.
- "Informational" for `CordonedNodesExcluded` / `HeterogeneousGPU` / `InsufficientGPUCapacity` undersells Warning-typed events; `InsufficientGPUCapacity` can precede a Failed write.

### Not blocking, but worth a sentence

Hooking emission at `updateStatusWithRetry` (ndipebot's design question) is the interesting alternative the record dismisses too quickly given twelve of sixteen sites already go through it. I do not need that chosen — I do need Alternatives Considered to evaluate it honestly if the dismissal stays.

Happy to re-review a revision that fixes (1)–(2) and lands (3)–(5) as explicit text. Until then, COMMENT only.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,635 @@
...
+   validation error as the message, at most once. Adding a hand-placed `warnf`
+   there as well would produce a redundant pair with the same reason.
+
+   The existing action-failure Warnings are kept **unless they share a
+   reason with the Failed transition on the same object.** The recorder
+   correlates on `(type, reason, regarding)`, not on message, so a
+   hand-placed Warning immediately followed by a Failed transition with the
+   same reason is not two rows but one `Event` with `count: 2`, which the
+   single-failure cases in this design forbid. Where that happens, the transition owns the
+   notification and the hand-placed emission is removed. Three sites match
```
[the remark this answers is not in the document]
kaynetu: Decision 5 retains distinct status-write-failure warnings, with the fallback contract and tests clarified in 58ff951.

docs/designs/080-phase-transition-events.md
kaynetu: 58ff951 makes preserving `WorkflowFailed` a prerequisite, specifies retryable status-write failures, and adds coverage for both callers.

docs/designs/080-phase-transition-events.md
kaynetu: 58ff951 corrects both correlation-key descriptions and explicitly requires `action == reason` for transitions and fallbacks.

docs/designs/080-phase-transition-events.md
kaynetu: 58ff951 describes pre-flush counts as a one-versus-at-least-two signal and uses recorder-level tests for exact emission counts.

docs/designs/080-phase-transition-events.md
kaynetu: 58ff951 distinguishes GoodputMeasurement’s 60s default and WorkloadRun’s fixed 15s interval from configurable polling.

docs/designs/080-phase-transition-events.md
kaynetu: 58ff951 documents periodic idle cleanup and fresh rows after closure, and requires a cumulative three-minute event-collection deadline.

docs/designs/080-phase-transition-events.md
kaynetu: 58ff951 rejects aggregation for count inflation and correctly describes a series staying open while emissions continue.

docs/designs/080-phase-transition-events.md
kaynetu: 58ff951 clarifies that `MeasurementCreationError` is non-fatal and reconciliation continues to the workload status update.

docs/designs/080-phase-transition-events.md
kaynetu: 58ff951 includes both `OverrideError` guards and specifies the same fallback and recorder-level test matrix for all four fallback sites.

docs/designs/080-phase-transition-events.md
kaynetu: 58ff951 changes the example to `WaitingForNodes` → `WorkflowCreated`.

docs/designs/080-phase-transition-events.md
kaynetu: 58ff951 corrects Context and References to state that `a331c85` wired the Job recorder while Workflow already had one.

docs/designs/080-phase-transition-events.md
kaynetu: 58ff951 identifies these as Warnings and documents `InsufficientGPUCapacity` followed by `Failed / PartitionError`.

docs/designs/080-phase-transition-events.md
kaynetu: 58ff951 corrects the inventory: Job has only `warnf`, while Certification has both `warnf` and `normalf`.

docs/designs/080-phase-transition-events.md
kaynetu: 58ff951 specifies `TestCaseParser` goldens for structured transition results, retaining tables for nil-safety and single-value assertions.

docs/designs/080-phase-transition-events.md
kaynetu: 58ff951 evaluates the retry-helper hook explicitly and specifies a shared pure flip detector while keeping event policy and post-write emission with callers.

docs/designs/080-phase-transition-events.md
kaynetu: 58ff951 adds the kube-apiserver reference for `--event-ttl` and keeps the Events API link separate.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,836 @@
...
+   The existing action-failure Warnings are kept **unless they share a
+   reason with the Failed transition on the same object.** The `events/v1`
+   recorder correlates on `(type, action, reason, reportingController,
+   reportingInstance, regarding, related)`, and the message is not part of
+   the key. Every emit helper in this repository passes the reason as the
+   `action` argument (`Eventf(obj, nil, eventType, reason, reason, ...)`),
+   and **this record makes that an explicit constraint**: the transition hook
+   and every fallback pass `action == reason` too. Under that constraint a
+   hand-placed Warning immediately followed by a Failed transition with the
+   same reason is not two rows but one `Event` with `count: 2`, which the
```
ndipebot: This pair cannot produce `count: 2`, and the mechanism matters because three other passages rest on it.

`eventKey.regarding` is a full `corev1.ObjectReference` compared by struct equality (`client-go@v0.36.4`, `tools/events/event_broadcaster.go:58-66`, `:177-178`, `:297-309`), and `tools/reference/ref.go:92-98` fills its `ResourceVersion` from the live object. At all three sites the status write refreshes that resourceVersion in place before the second emission: `pkg/controller/status.go:42` calls `c.Status().Update(ctx, obj)` on the caller's own pointer, and controller-runtime `pkg/client/typed_client.go:268,287` sets `body := obj` and finishes `.Do(ctx).Into(body)`. Decision 3 emits the transition only after that write succeeds, so the hand-placed Warning carries RV n and the transition carries RV n+1. Different key, no isomorphic hit at `:177-178`, two rows with `count: 1` each.

The hedge on 258-262 names `action` as the only thing that could split the pair, and resourceVersion appears nowhere in the document.

Decision 5's conclusion survives unchanged: two identical rows for one failure violate the same single-row expectation the cases assert, and the collector serializes every row. What needs restating is the mechanism sentence here and in the three places that repeat it.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,836 @@
...
+   projection for deterministic test fixtures: `type`, `reason`,
+   `message`, `involvedObject.kind`, `involvedObject.name`, and `count`,
+   sorted by the first five. Event names, UIDs, timestamps, source, and
+   reporting instance are omitted; they are not stable across runs.
+
+   `count` is load-bearing but coarse. The `events/v1` recorder correlates
+   on `(type, action, reason, reportingController, reportingInstance,
+   regarding, related)`; the message is **not** part of that key, and
+   `action == reason` by the decision 5 constraint. Repeated emissions for
+   the same object and reason therefore collapse into one `Event` rather
```
ndipebot: Same library point, and here it is a missing qualifier rather than a wrong claim. Repeated emissions collapse into one `Event` only while the involved object's resourceVersion is unchanged, because `eventKey.regarding` is the whole `ObjectReference` (`event_broadcaster.go:58-66`, `:297-309`; `tools/reference/ref.go:92-98`).

That qualifier does hold for the defect this paragraph is actually about. A naive per-poll emitter writes nothing, so `pkg/controller/status.go:37-41` skips the write and the resourceVersion holds still, which is exactly the growing count described on 787-795. It does not hold for the decision 5 pairs, where a status write sits between the two emissions. Worth one clause so the two cases do not read as one rule.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,836 @@
...
+  - a Certification rejected-Create case asserting the persisted reason stays
+    `WorkflowFailed` (the catch-all no longer overwrites it) and Events and
+    Conditions agree; and a genuine build error asserting
+    `WorkflowValidationFailed` on both;
+  - a Certification `WorkflowCreationError` case asserting both the action
+    Warning and the outcome Warning appear with their distinct reasons;
+  - the three same-reason sites from decision 5 (WorkloadRun `BuildFailed`,
+    Workflow `HeterogeneousPlatform`, Workflow `OverrideError`), each
+    asserting a single transition row with `count: 1` and no fallback after
+    successful status persistence. A `count: 2` here is the regression these
```
ndipebot: Following from the resourceVersion point on 257: the duplicate these three cases exist to catch will show up as an unexpected second row with `count: 1`, not as `count: 2` on the row that already satisfied readiness.

The test design is fine either way, since the projection serializes every row for the object and an extra row fails the golden just as a `2` would. It is the stated signature that needs changing, because it is the sentence someone reviewing a failed golden will reach for.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,836 @@
...
+about validation. Suppressing the pass event would make a Job that passed
+indistinguishable, in its Events section, from one whose validation has not
+run. The pass event is the only row keyed on a condition being written
+`False`, and that asymmetry is stated in decision 1 rather than hidden.
+
+### Keep every existing action Warning alongside the new transition events
+
+Rejected. At three sites the action Warning and the Failed transition share
+reason and object (and at two of them the message as well). The recorder
+correlates on reason, not message, so each pair collapses into one `Event`
```
ndipebot: Third instance of the same premise. "each pair collapses into one `Event` with `count: 2`" and "an inflated count for a single failure" should both be two duplicate rows for a single failure, for the reason on 257.

The rejection itself stands on the duplication alone, so this is a rewrite of the justification rather than a change of verdict.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,836 @@
...
+   completion. The tier that fails on a verdict is the Workflow, whose
+   `Failed` reason becomes `JobValidationFailed` or `JobHardwareFailed`.
+   Hardware failure likewise leaves the phase alone; execution continues.
+
+   Because `Succeeded` cannot stand in for the passing verdict, the pass gets
+   its own event too: a `Normal / ThresholdsMet` when `ValidationFailed` is
+   first written `False`, so the Events section of a Job that ran and passed
+   reads `Normal / WorkloadCompleted` then `Normal / ThresholdsMet`, and one
+   that ran and failed validation reads `Normal / WorkloadCompleted` then
+   `Warning / ThresholdViolation`. Without the pass event, a `Succeeded` Job
```
ndipebot: The Job's `ValidationFailed` reason on a threshold miss is `ThresholdViolated`, not `ThresholdViolation`. `pkg/threshold/evaluator.go:147` sets it, `pkg/controller/job_controller.go:787` writes it through, and it is persisted that way in `cmd/integration/testdata/reconcile/job-goodput-threshold-fail/expected.json:70`, with the same string in `job-bandwidth-threshold-fail` and `job-bandwidth-threshold-range-fail`.

`ThresholdViolation` (`pkg/controller/helpers.go:60`, `api/v1alpha1/job_types.go:82`) has two uses, neither of them a Job condition reason: a `NodeFailureReason` inside `status.failedNodes` (`job_controller.go:827`), and the WorkloadRun mirror condition (`workloadrun_controller.go:162`). Since decision 4 makes the event reason the condition's reason, the event a violating Job emits is `Warning / ThresholdViolated`.

Same substitution needed on 76, 95, 125 and 553. Separately, 206 offers `ReasonThresholdViolation` as an example of an emitted transition reason, but its only condition use is the WorkloadRun mirror, which 122-128 explicitly gives no event, so that example cannot occur.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,836 @@
...
+    asserts a single InProgress event;
+  - a Job hardware-failure case: the first detection yields one
+    `Warning / HardwareFailureDetected` row with `count: 1`; a second pass that adds another
+    failed node while the condition is already `True` emits nothing more; the
+    Job's execution phase is unchanged by the verdict, and if the workload
+    itself later fails, that `Failed` transition is a separate row with its
+    own reason;
+  - a Job threshold case in each direction, both asserting the Job stays
+    `Succeeded`: a violation shows `Normal / WorkloadCompleted` then
+    `Warning / ThresholdViolation` and no Job `Failed` row, with the owning
```
ndipebot: This occurrence is not just wording. Expected rows are declared verbatim as `{type, reason}` pairs and readiness waits for every `expect` row (349-368), now under a cumulative three-minute budget. An implementer transcribing `ThresholdViolation` from here writes a reason that is never emitted, so the case hangs to the deadline rather than failing on a diff, which is a much worse signal to debug. Should be `ThresholdViolated`, per the note on 113.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,836 @@
...
+   `ValidationFailed`.** Both are deliberately outside the Job's exclusive
+   set. `HardwareFailed` is not terminal; execution continues after it is
+   set. `ValidationFailed` records the threshold verdict and is written
+   `True` on a violation, a measurement timeout, or an unknown threshold key,
+   and `False` with `ThresholdsMet` when thresholds pass. Neither is a phase
+   transition, so the rule above would never emit for them, yet they are the
+   two verdicts a burn-in exists to produce. Each gets its own event: a
+   `Warning` emitted once, when the condition flips from absent or `False` to
+   `True`, with the condition's reason (`HardwareFailureDetected`,
+   `ThresholdViolation`, `MeasurementTimeout`, `UnknownThresholdKey`) and
```
ndipebot: Minor gap in the same list: a malformed CEL threshold expression also reaches `ValidationFailed=True`, with reason `InvalidThresholdExpression` (`pkg/threshold/evaluator.go:137`, written through `job_controller.go:787`).

Nothing catches it earlier. `threshold.ValidateKeysError` (`evaluator.go:78-85`) only checks registry membership, there is no webhook, and `api/v1alpha1/job_types.go:234` declares `Thresholds` as a plain `map[string]string`. The enumeration on 88-91 has the same gap.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,836 @@
...
+  - At each call site that runs `setWorkloadRunCondition` then
+    `Status().Update`, capture the previously-true execution type before the
+    mutation and emit after a successful update when it differs. Prefer one
+    small helper that wraps the pair so the rule lives in one place.
+- `cmd/integration/integration_test.go`
+  - Wire recorders for all six reconcilers.
+  - Enforce decision 6's cumulative 3-minute deadline for opted-in cases from
+    fresh recorder startup through collection, sharing the remaining budget
+    across waits. Leave cases without event collection unchanged.
+  - Add an `events` list to `waitConfig` with `involvedKind`, `involvedName`,
```
ndipebot, line 495: This filter is name-scoped, and the suite does not isolate Events between cases, so an opted-in case can collect a previous case's rows.

`cmd/integration/integration_test.go:55-67` starts one envtest control plane for the whole suite and `:74` runs every case directory against it. `startManager` at `:91` is per-case, so the fresh-broadcaster requirement on 340 is already satisfied by the harness, but the API server is never reset. `deleteTestObjects` (`:334-356`) deletes only the objects the case declared, and nothing deletes Events. Almost every fixture lives in `default`. Event names carry a `UnixNano` suffix (`event_recorder.go:99`) so they never collide, and involved-object names already repeat across cases today.

The projection drops UIDs (319) and this list takes `involvedKind`, `involvedName` and `namespace`, so as specified the outcome depends on case ordering. Worth stating whether the list is UID-scoped, or whether an opted-in case deletes Events in the namespace before the manager starts.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,836 @@
...
+    itself later fails, that `Failed` transition is a separate row with its
+    own reason;
+  - a Job threshold case in each direction, both asserting the Job stays
+    `Succeeded`: a violation shows `Normal / WorkloadCompleted` then
+    `Warning / ThresholdViolation` and no Job `Failed` row, with the owning
+    Workflow's `Failed / JobValidationFailed` row in the same case; a pass
+    shows `Normal / WorkloadCompleted` then `Normal / ThresholdsMet`;
+  - a checkpoint-restart Job case asserting the restart does not emit a second
+    InProgress event for the same Job when the phase does not flip;
+  - the second `OverrideError` site (the early `applyOverrides` guard): the
```
ndipebot, line 596: These two cannot be integration cases as written. There is no seam in the harness to make `Status().Update` fail against envtest, and `waitConfig` has no error-injection field. The same applies to the failed-write halves implied on 567-570.

The recorder-level matrix on 509 is where the failed-write assertions belong. The integration case can still carry the success half, which is the part envtest can actually drive.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,836 @@
...
+   timed-out Job carries `InProgress=True` and `Failed=True` together; the
+   "one previously-true type" assumption does not hold there. The rule for
+   this site is therefore stated directly: emit `Warning / JobTimedOut`
+   regarding the **Job**, from the Workflow reconciler's recorder, when
+   `JobFailed` was not `True` before the write and the write succeeds. The
+   `!ts.terminal` guard already ensures the write happens once; the pod-drain
+   re-entry pass sees a terminal Job and does not rewrite. Fixing the
+   exclusivity violation itself (routing this write through the shared helper
+   so `InProgress` flips `False`) is out of scope: it would change every
+   timeout golden's conditions and `ObservedGeneration`, and belongs in its own
```
ndipebot: This cost does not exist. No `expected.json` under `cmd/integration/testdata/reconcile/` contains `JobTimedOut`, and no case drives a Job timeout: `workflow-suspended-job-no-timeout` asserts the absence, and `job-bandwidth-threshold-timeout` is a `MeasurementTimeout`. The only repo hits for `JobTimedOut` are a comment in an `input_config.yaml` and an input file under `pkg/controller/testdata/`.

The deferral is still right on the other ground you give, that it is a status-correctness fix deserving its own record. I would drop the golden-churn half here and on 726-729 rather than leave a reader sizing the deferral against an empty set.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,836 @@
...
+  and `setJobValidationStatus` through their own `updateStatusWithRetry`
+  calls, not through the exclusive-set wrapper. That is why decision 1 gives
+  them an explicit rule rather than assuming the phase hook covers them.
+  Neither verdict changes the Job's execution phase: a threshold violation
+  lands on a Job that is already `Succeeded`, and a hardware failure lands on
+  one that is still running. The lifecycle consequence appears one tier up,
+  as the Workflow's `Failed / JobValidationFailed` or `Failed /
+  JobHardwareFailed`. Reading a Job's events top to bottom therefore gives
+  "what happened" (`WorkloadCompleted`) and then "what was found"
+  (`ThresholdsMet` or a verdict Warning), as two rows.
```
ndipebot: Three rows, not two, under decision 1. The Job's InProgress transition emits `Normal / WorkloadCreated` (`pkg/controller/job_controller.go:399`), so the section reads `WorkloadCreated`, then `WorkloadCompleted`, then the verdict. Same count on 110-113.

ndipebot (commented):
Re-reviewed at `58ff951`. All fifteen of my inline findings and CodeRabbit's are addressed in the text, and I checked each one against the code rather than against the reply. Fourteen are cleanly closed. The correlation-key fix is closed as stated but carried a second-order error into three places, which is the first comment below.

Three things I would want resolved before this lands:

1. The `count: 2` premise under decision 5 (comments on 257, 325, 570, 712). `eventKey.regarding` is the full `ObjectReference` including `ResourceVersion`, and a status write sits between the two emissions, so the pairs land as two rows rather than one row with an inflated count. The decision survives, the mechanism and the stated regression signature do not. This is the third round in which the events library has been described a different wrong way, so I would rather we pin it once against the pinned source than iterate again.
2. `ThresholdViolation` is not the Job's condition reason (comments on 113, 553, 95). As written, the integration fixture on 553 declares a reason that is never emitted, so that case would hang to the new three-minute budget instead of failing on a diff.
3. Decision 6's collector is name-scoped against a suite that shares one API server and never deletes Events (comment on 466). As specified the result is order-dependent.

The rest are small: two fallback assertions that cannot be integration cases, a deferral rationale citing golden churn against an empty set, a row count, and one reconciler missing from the requeue list.

Everything else in the revision holds up. I re-derived the `refreshTime` and `finishTime` behaviour, the second-emission-only patch, and the fresh-broadcaster deadline argument against `client-go@v0.36.4`, and they are all correct now.

docs/designs/080-phase-transition-events.md
ndipebot: Confirmed against the commit itself. Context on 9 and the References entry on 829 both read correctly now.

docs/designs/080-phase-transition-events.md
ndipebot: Closed on both numbers. One gap left in the same sentence: it says every reconciler requeues and then names five of six. BandwidthMeasurement is missing (`pkg/controller/bandwidthmeasurement_controller.go:34`, `:106`). The five you do name check out.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. `WaitingForNodes` to `WorkflowCreated` is a pair that exists.

docs/designs/080-phase-transition-events.md
ndipebot: Closed, and better than I asked for. The new Alternatives entry on 761-770 evaluates the hook and rejects it on `updateStatusWithRetry` being generic over `client.Object`, and decision 3 now routes through a shared flip detector on 183 and 406, which collapses the per-seam prose I was objecting to.

docs/designs/080-phase-transition-events.md
ndipebot: The key is right now: all seven fields, matching `getKey` at `event_broadcaster.go:297-309`, and making `action == reason` an explicit constraint is the correct fix.

One consequence of the corrected key did not make it through. `regarding` is the full `ObjectReference`, and it carries `ResourceVersion`, which changes what decision 5 predicts for the same-reason pairs. New comment on 257.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. Both guards are in the inventory on 270-274 and the fourth fallback site is specified, which is the even treatment I was after. Separate point about where two of those assertions can run, on 558.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. 304 states the non-fatal continue.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. Warning type and the `InsufficientGPUCapacity` to `Failed / PartitionError` follow-on are both on 300.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. 329-335 separates the 30 minute refresh from the 6 minute idle cleanup and says closure lands on an eligible cleanup pass rather than exactly 6 minutes after the last emission. I re-read `finishSeries` at `event_broadcaster.go:145-161` and the description matches.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. One versus at-least-two is the right characterisation and matches `recordToSink` at `event_broadcaster.go:179-192`.

The cumulative three-minute deadline works as an argument too, and I checked why: `wait.UntilWithContext` at `event_broadcaster.go:393-394` fires both loops immediately at t=0 against an empty cache and not again inside a three-minute window, so with a fresh broadcaster per case neither flush can land mid-case.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. The typed-error prerequisite on 226-237 does it, and using the existing collision handler as precedent is the right framing.

One item for the implementation checklist. When the inner `setCertificationFailed` fails, nothing writes terminal status any more, so the comment at `certification_controller.go:602-606` that justifies one event per failed Create attempt on the grounds that subsequent requeues short-circuit stops being true on that branch. 232-235 already says the error-driven retry is intended, so it is only the comment that needs to follow the code.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. 431 has it right.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. 482 and 532 put the structured transition results in `testutil.TestCaseParser` goldens and keep tables for the nil pins and the recorder counts, which is inside the exception as written.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. 787-795 now describes the series staying open under a 15s requeue and the count growing for the life of the phase, which is the mechanism. The fresh-row-after-idle-cleanup half is there too.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. 833 cites the kube-apiserver page and 832 keeps the Events API link separate.

docs/designs/080-phase-transition-events.md
kaynetu: Addressed in 6497851. Decision 5 now states that regarding includes resourceVersion: the pre-write Warning carries RV n and the post-write transition carries RV n+1, producing two rows with count 1. The single-notification conclusion remains unchanged.

docs/designs/080-phase-transition-events.md
kaynetu: Addressed in 6497851. Decision 6 now says emissions aggregate only while every key field, including resourceVersion, is unchanged. It also explains that a no-op poll preserves the resource version, while an intervening status write can produce a fresh row.

docs/designs/080-phase-transition-events.md
kaynetu: Addressed in 6497851. The test plan now identifies an unexpected second row with count 1 as the regression signature because the pre-write and post-write emissions carry different resource versions.

docs/designs/080-phase-transition-events.md
kaynetu: Addressed in 6497851. The rejected alternative now describes two duplicate rows with count 1, caused by the resourceVersion change across the status write, instead of one row with an inflated count.

docs/designs/080-phase-transition-events.md
kaynetu: Addressed in 6497851. Every Job condition and event occurrence now uses ThresholdViolated, and the transition-reason example uses ThresholdViolated rather than the non-emitting WorkloadRun mirror constant.

docs/designs/080-phase-transition-events.md
kaynetu: Addressed in 6497851. The threshold-failure integration expectation is now Warning / ThresholdViolated, matching the exact Job condition reason used by readiness matching.

docs/designs/080-phase-transition-events.md
kaynetu: Addressed in 6497851. Both ValidationFailed outcome enumerations now include InvalidThresholdExpression, including the emitted Warning reason list.

docs/designs/080-phase-transition-events.md
kaynetu: Addressed in 6497851. The design now requires resolving the current involved object UID and filtering both readiness and serialization by that UID. The unstable UID is omitted only from the golden projection.

docs/designs/080-phase-transition-events.md
kaynetu: Addressed in 6497851. Envtest coverage now keeps only the successful status-persistence case. Status-update failure assertions, including the failed-write fallback cases, are assigned to the recorder-level matrix because the integration harness has no failure-injection seam.

docs/designs/080-phase-transition-events.md
kaynetu: Addressed in 6497851. The claim about changing existing JobTimedOut integration goldens was removed. The deferral now rests on the separate status-correctness change and calls for dedicated tests and golden review.

docs/designs/080-phase-transition-events.md
kaynetu: Addressed in 6497851. Decision 1 now lists WorkloadCreated, WorkloadCompleted, and the verdict for both pass and fail paths. The Notes section likewise describes the sequence as three rows.

kaynetu:
@coderabbitai Agreed on the Workflow path. The ADR overstated “execution continues” by describing only the Job condition model.

`HardwareFailed` stays additive on the Job: `isTerminalState` ignores it, so the Job exclusive set does not flip. The Workflow path is already terminal: `getJobTerminalState` treats `hwFailed` as terminal, and `completeTerminalGroup` deletes the TrainJob, then fails or retries the group. That is existing orchestration, not something this events record will change.

We will document that existing split rather than keep the workload running so a later Job `Failed` can fire. On a Workflow-owned Job the expected extra rows are Job `Warning / HardwareFailureDetected` and Workflow `Failed / JobHardwareFailed`. A later Job `Failed / WorkloadFailed` is not the designed next step; if it appears it is from the Job reconciler observing the TrainJob deletion, not from the workload failing on its own. A direct Job with no Workflow parent can remain `InProgress` and later emit `Failed` if the workload fails independently.

ndipebot (commented):
Re-reviewed at `b195d76`. All eleven findings from my last pass are genuinely closed, and I checked each against the source rather than the reply.

The three substantive ones:

1. The `count: 2` premise is now correct in all four places (265-276, 345-350, 606-610, 748-754). 271-274 states that `regarding` carries `resourceVersion`, that the pre-write Warning is RV n and the transition RV n+1, and that each row lands with `count: 1`, while keeping the counterfactual for an unchanged key. The decision-6 qualifier on 345-349 is right to point out that a no-op poll performs no write, which is what `pkg/controller/status.go:37-41` does, so the naive per-poll case still aggregates as described. The regression signature on 606-610 now names the extra row.
2. `ThresholdViolated` is used everywhere. `ThresholdViolation` has zero occurrences left in the document, the integration expectation on 590 is correct, and `InvalidThresholdExpression` is in both enumerations. The transition-reason example on 217 correctly dropped the WorkloadRun mirror constant.
3. The collector is UID-scoped now (331-337, 496-501, 388-390), with the reason stated. That closes the cross-case bleed, and omitting the UID from the projection while filtering on it is the right split.

The smaller items are all closed too: the envtest seam (598-601), the timeout deferral now resting on the runtime change rather than golden churn (186-189, 765-769), the three-row sequence (112-118, 845-856), the BandwidthMeasurement requeue, and the `certification_controller.go` comment added to the checklist on 467-469.

I also checked the new hardware-failure lifecycle text from `277959e`, since it makes fresh claims about orchestration. It holds: `getJobTerminalState` sets `terminal = hwFailed || failed || succeeded` (`pkg/controller/workflow_controller.go:1416`), `completeTerminalGroup` folds `hwFailed` into `jobFailed` and deletes the workload before the retry decision (`:1339-1348`), and the retry-or-fail branch is `:1360-1385`. Hedging the later `Failed / WorkloadFailed` as an ordering-dependent teardown artifact rather than a promised row is the right call.

Two trivia below, neither worth holding for.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,879 @@
...
+against the steady-state requeue because the action either does not recur or
+recurs only when it fails again.
+
+Issue #252 is the deliberately split-off other half: events on the phase
+transitions themselves (InProgress, Succeeded, Failed). #250 excluded them
+because they need dedup logic against the poll. Every reconciler requeues:
+Job, Workflow, and Certification on a configurable interval (15s in
+production), WorkloadRun on a fixed 15s package constant, and
+GoodputMeasurement on its 60s default sample interval. BandwidthMeasurement
+uses a configurable requeue interval with a 15s production default. A naive
```
ndipebot, line 25: Accurate as far as it goes: `defaultBandwidthMeasurementRequeueInterval = 15 * time.Second` (`pkg/controller/bandwidthmeasurement_controller.go:34`) behind the `RequeueInterval` field and `getRequeueInterval()` (`:68`, `:110-114`).

Worth one clause though, because it is the same distinction GoodputMeasurement got called out for last round. During active measurement BandwidthMeasurement requeues on `getSampleInterval`, not on the requeue interval, and that resolves to the shared `defaultSampleInterval` of 60s (`bandwidthmeasurement_controller.go:102-107` returning the constant at `goodputmeasurement_controller.go:35`). It is the path taken at `:210`, `:219`, `:257`, `:296`, `:317` and `:360`; the 15s interval covers the surrounding states at `:167`, `:196`, `:235`, `:242` and `:247`. So the sibling tiers actually sample at the same cadence.

Low stakes, since decision 1 gives neither measurement tier a transition event, so nothing downstream is sized off this number.

docs/designs/080-phase-transition-events.md
```diff
@@ -0,0 +1,879 @@
...
+it has been idle for more than 6 minutes), so with a 15s requeue a
+naive per-pass emission produces one row whose count grows for the life of
+the phase, plus a fresh row if an idle gap lets cleanup remove the series.
+That is inflation, not suppression, and it is only visible to the API server
+on the second emission
+and at flush time. Dedup must be a property of the emit decision.
+
+## Notes
+
+- **ADR-078 is held by [PR #336](https://github.com/NVIDIA/cluster-readiness-engine/pull/336)
```
ndipebot, line 839: Stale as of the merge in `b195d76`. PR #336 is merged, and the merge brought `078-jobset-ownership.md` onto this branch along with its index row, so the README now reads `077 / 078 / 080`. Only ADR-079 (PR #337) is still open.

The note can shrink to just the 079 case, or come out entirely once #337 lands.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. 271-274 now states that `regarding` carries `resourceVersion`, that the pre-write Warning is RV n and the transition RV n+1, and that each row lands with `count: 1`. Keeping the unchanged-key counterfactual is a good call, since that is the case decision 6 relies on.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. The qualifier on 345-349 is right, and so is the reason given for it: a no-op poll performs no write (`pkg/controller/status.go:37-41`), so the resource version holds still and the naive per-poll defect still aggregates the way the paragraph describes.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. 606-610 names the extra row as the signature and says why the two emissions carry different resource versions.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. 748-754 now reads as two duplicate rows for one failure, and the rejection is unchanged.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. Zero occurrences of `ThresholdViolation` left in the document, and 217 correctly dropped the WorkloadRun mirror constant from the transition-reason example.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. 590 reads `Warning / ThresholdViolated`, which is what `pkg/threshold/evaluator.go:147` actually writes, so the fixture can satisfy readiness.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. `InvalidThresholdExpression` is in both enumerations now, on 77 and 100.

docs/designs/080-phase-transition-events.md
ndipebot: Closed, and the fix is the right shape. 331-337 states the shared-API-server reason, 496-501 puts UID resolution in the harness checklist, and 388-390 applies it to readiness as well as serialization. Filtering on the UID while keeping it out of the projection is exactly the split I was after.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. 598-601 moves the failed-write half to the recorder-level matrix and names the missing injection seam as the reason, keeping the success half in envtest.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. 186-189 and 765-769 now rest the deferral on the runtime status change rather than on golden churn, which is the part that was checkable and wrong.

docs/designs/080-phase-transition-events.md
ndipebot: Closed. Three rows on both 112-118 and 845-856. The rewritten hardware-failure path is a genuine improvement over what I flagged: I checked `getJobTerminalState` at `pkg/controller/workflow_controller.go:1416` and `completeTerminalGroup` at `:1339-1348`, and `hwFailed` does fold into `jobFailed` and drive workload deletion before the retry-or-fail branch, so the text matches the code.

ndipebot (approved):
/lgtm
/approve

docs/designs/080-phase-transition-events.md
kaynetu: noted

docs/designs/080-phase-transition-events.md
kaynetu: noted

asivanadi0 (approved):
APPROVE — re-review at HEAD `b195d764fd04dfb7b2b0d2da460c19b516dc91e8` (tip after `58ff951` / `6497851` / `277959e` / main merge).

Prior COMMENT was on `63aee68`. Re-read ADR-080 + README Notes at this tip and verified each previously flagged gap against the text and the cited controller surfaces.

### Closed since `63aee68`

1. **`Failed / WorkflowFailed` vs catch-all overwrite (ndipebot #1)** — Decision 5 now requires a typed Create-rejected error and catch-alls that skip the second Failed write; persisted reason is `WorkflowFailed`, so Events and Conditions can agree under emit-on-flip. Integration cases assert the catch-all no longer clobbers it.
2. **events/v1 correlation + `series.count` (ndipebot #2/#3)** — Decision 5 pins `action == reason` as an explicit constraint, keys on the full `regarding` ObjectReference (including resourceVersion), and explains why pre-write / post-write pairs land as two `count: 1` rows rather than `series.count: 2`. Decision 6 describes second-emission-only API patch, 30m refresh / 6m idle finalize, and the coarse `1` vs `≥2` collector signal with a fresh-broadcaster 3-minute budget.
3. **README index 078/079** — Notes explicitly reserve ADR-078 (#336) and ADR-079 (#337); branch gap is merge ordering. Index on this tip includes 078 after the #336 merge; 079 noted as held by #337.
4. **Issue #252 six-reconciler scope** — Decision 1 states aloud that this amends the issue title; revisit trigger is a future measurement `Failed` condition.
5. **WorkloadRun `ValidationFailed`** — Decision 1: status-only, no event; Job Warning + Workflow Failed already carry the operator-facing facts.

### Accuracy nits also closed

- Context uses `WaitingForNodes` → `WorkflowCreated` (not `Running`).
- Implementation notes Job has only `warnf`; Certification has both.
- Early `OverrideError` `applyOverrides` guard is transition-only with the same `OverrideErrorStatusUpdateFailed` fallback — three dual-emit sites, four fallback sites.
- `CordonedNodesExcluded` / `HeterogeneousGPU` called Warning-typed exclusion notices.
- Alternatives Considered now evaluates `updateStatusWithRetry` honestly (partly adopted via shared `conditionFlip`).

No residual gaps from the prior pass. Approving the record at this tip.

PR #346, opened 2026-09-14
fix(setup): respect JobSet ownership across install, reset, and recovery

### LINKED ISSUE #335: [Bug]: setup init [deps] fails on clusters that already have JobSet installed; setup reset deletes the pre-existing JobSet CRD

kaynetu:
## Summary

Implement ADR-078 for issue #335 so `nvcrectl setup` can coexist safely with independently managed JobSet installations.

- detect JobSet ownership with a CRD-first fast path and a complete, paginated fingerprint scan; distinguish bundled, external, CRD-only, absent, and unknown states, and fail closed when ownership is ambiguous
- select the matching Trainer install mode without refreshing the retained JobSet CRD schema; preserve pinned-version skips and refuse existing Trainer installations whose JobSet CRD is missing
- retain the JobSet CRD unconditionally during reset and recovery, protect unrelated namespace resources with verified ownership chains and repeated inventory checks, and enforce namespace UID preconditions before deletion
- make phase skipping effective, preserve Helm-state errors, propagate Trainer uninstall failures, and report retained resources and partial failures accurately
- add ownership, recovery, phase-isolation, and failure-path regressions plus a dedicated disposable Kind/Helm lifecycle UAT covering bundled upgrade, reset/init, and an external renamed JobSet controller
- update setup, installation, deployment, and ADR documentation for the new ownership and recovery boundaries

## Related Issue

Closes #335

Implements the design approved in #336.

## Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [x] 🔧 Refactoring
- [x] 🔨 Build/CI

## Component(s) Affected
- [ ] API / CRDs
- [ ] Catalog / Workloads
- [x] CLI (nvcrectl)
- [x] Helm / Deployment
- [x] Documentation / CI
- [x] Other: setup lifecycle UAT

## Testing
- [x] Tests pass locally

Verified on the final signed commit with:

- `make lint-fix`
- `make build`
- `make test`
- `make verify-doc-links` (279 relative links)
- `make test-uat-setup` using Helm 3 and a disposable Kind cluster

The setup lifecycle UAT passed all three scenarios: bundled upgrade retains the JobSet CRD, bundled reset/init retains and reuses it, and an external renamed JobSet controller survives init/reset with the CRD present.

## Risk Assessment

- Ambiguous or incomplete ownership evidence now stops setup and prints the manual Trainer install plus `--skip-phases=deps` path. Restricted kubeconfigs can therefore require operator action instead of allowing destructive recovery.
- Reset and recovery never delete the shared JobSet CRD. JobSet schema refresh remains operator-managed.
- Automatic namespace recovery remains intentionally narrow: unexpected protected resources, incomplete discovery, changed UIDs, or failed repeated checks abort before destructive cleanup. Events and EndpointSlices do not block recovery, as specified by ADR-078.
- Trainer uninstall failures now stop reset before CRD deletion and return a non-zero status; rerunning remains convergent because Helm uninstall uses `--ignore-not-found`.

## Checklist
- [x] Self-review completed
- [x] Ready for review

No API type or CRD schema was changed; generated manifests remain unchanged.


hack/test-uat-setup.sh
```diff
@@ -0,0 +1,53 @@
...
+    kind delete cluster --name "${cluster}" || true
+  fi
+  exit "${status}"
+}
+if kind get clusters | grep -Fxq "${cluster}"; then
+  echo "Refusing to reuse existing Kind cluster ${cluster}"
+  exit 1
+fi
+
+owned=1
```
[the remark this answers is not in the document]
kaynetu: Fixed in 95d54c0. `owned=1` now runs only after `kind create cluster` succeeds, so the cleanup trap cannot delete a cluster this process did not prove it created. The isolated setup lifecycle UAT passed all three scenarios.

pkg/setup/helm.go
```diff
@@ -455,9 +500,40 @@ func classifyHelmInstallFailure(output, namespace string) failureClass {
...
+		{"clusterrolebinding", jobSetControllerName},
+		{"validatingwebhookconfiguration", "jobset-validating-webhook-configuration"},
+		{"mutatingwebhookconfiguration", "jobset-mutating-webhook-configuration"},
+	}
+	for _, object := range known {
+		// Bind the object to Helm's ownership-error clause on the same line.
+		// A debug mention elsewhere must not reclassify an unrelated collision.
+		pattern := `(?im)^Error: (?:INSTALLATION|UPGRADE) FAILED: (?:Unable to continue with (?:install|update): )?` +
+			regexp.QuoteMeta(object.kind) + ` "` + regexp.QuoteMeta(object.name) +
+			`" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata;[^\r\n]*meta\.helm\.sh/release-`
```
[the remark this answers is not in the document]
kaynetu: Addressed in 95d54c0 with a captured Helm 3.19.2 upgrade fixture. Reproducing JobSet 0.10.1 plus Trainer 2.2.1 with an explicit `jobset.install=true` upgrade produced `Error: UPGRADE FAILED: Unable to continue with update: ClusterRole ...`; the existing regex classifies it as JobSet ownership. Helm 3.19.2 does not emit the proposed `rendered manifests contain...` clause on this path, so the regex remains constrained and the real transcript is pinned in `helm3-upgrade-collision`.

pkg/setup/setup.go
```diff
@@ -432,39 +540,67 @@ func handleTrainerInstallFailure(sp setupPhaseParams, output string, installErr
...
-// exists, or when a TrainingRuntime/ClusterTrainingRuntime exists that is
-// not Helm-owned, because deleting the CRDs would destroy them. Any listing
-// failure blocks recovery too: safety must be proven, not assumed.
-func trainerRecoveryGate(ctx context.Context, c client.Client) (bool, []string) {
+// trainerWorkloadGate enforces the cluster-wide custom-resource blockers.
+func trainerWorkloadGate(sp setupPhaseParams) (bool, []string) {
 	var blockers []string
 	for _, group := range []string{trainerAPIGroup, jobsetAPIGroup} {
-		resources, err := discoverResourcesByGroup(ctx, c, group)
+		resources, err := discoverResourcesByGroup(sp.ctx, sp.c, group)
```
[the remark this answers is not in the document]
kaynetu: Fixed in 95d54c0. `discoverResourcesByGroup` now returns an error when a matching CRD has no served version. The phase-level `unserved-trainer-crd-refuses` golden proves recovery stops before mutation: `uninstallCalls: 0` and `crdDeletes: 0`.

pkg/setup/helm.go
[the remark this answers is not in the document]
kaynetu: Clarified in a33dfa1: NVCRE sets `jobset.install=false` only for external mode; bundled mode omits the override and relies on the chart default. The reproduction explicitly restores `true` because the seed release persisted `false`. The fixture now records all three exact Helm commands and uses that wording.

pkg/setup/jobset_ownership.go
```diff
@@ -0,0 +1,649 @@
...
+			}
+		}
+		return jobSetObservation{mode: jobSetModeBundled,
+			evidence:        []string{"exact Trainer release manifest and live JobSet resources agree"},
+			manifestObjects: manifestObjects}
+	}
+	if manifestContainsJobSet(manifestObjects) {
+		return unknownJobSet("external live ownership conflicts with bundled JobSet resources in the exact Trainer release manifest")
+	}
+	supportToken, err := verifyExternalJobSetController(ctx, c, findings, owner)
```
ndipebot, line 222: This is the one I want fixed before merge.

`verifyExternalJobSetController` confirms that a webhook `clientConfig.service` ref resolves to a same-release `Service` (`:462`) backed by a same-release `Deployment` whose labels the selector matches (`:477`). It never checks a name, and it never establishes that the release reconciles the JobSet API. Compare the non-Helm path, which pins `jobset-webhook-service` at `:238` and the `jobset-controller` Deployment at `:249`.

Upstream of that, `objectHasJobSetRule` (`:400-426`) matches on `apiGroups` alone, checking neither `resources` nor `verbs`, and the only consumer exclusion in the file is the hardcoded Trainer role at `:284`. So any Helm-installed JobSet *consumer* that has an admission webhook on jobsets arrives here and is classified `external`.

Kueue is the case that worries me, and we support it as a gang scheduler (`api/v1alpha1/job_types.go:296`). Its chart renders:

1. `kueue-manager-role` with unconditional `jobset.x-k8s.io` rules on `jobsets`, `jobsets/finalizers` and `jobsets/status` (`charts/kueue/templates/rbac/role.yaml:197-222`).
2. `mjobset.kb.io` and `vjobset.kb.io` with `rules.apiGroups: [jobset.x-k8s.io]` and `clientConfig.service` pointing at `kueue-webhook-service`, neither behind an `{{ if }}` (`templates/webhook/manifests.yaml:182-210` and `:847-876`).
3. `app.kubernetes.io/managed-by: Helm` on all of them via `kueue.labels`, plus Helm's own `meta.helm.sh/*` annotations from `setMetadataVisitor`.
4. A webhook `Service` selector and a `kueue-controller-manager` pod template that both use `kueue.selectorLabels`, so `selectorMatches` succeeds.

The state that triggers it is one this PR creates. Decision 3 makes `setup reset` retain `jobsets.jobset.x-k8s.io` while Helm uninstall still removes the bundled controller. Re-run `init` on that cluster with Kueue installed: the manifest read is skipped at `:97`, `manifestContainsJobSet({})` is false, and we select `external`. `newTrainerHelm`'s install closure at `setup.go:339-341` passes `mode != jobSetModeExternal` as false, and `trainerHelmUpgradeArgs` appends `--set jobset.install=false` (`helm.go:278-280`). Trainer comes up with no JobSet controller anywhere on the cluster, every TrainJob's JobSet is created and never reconciled, and certifications hang with no diagnostic. It persists, because the next `init` sees a release deployed at the pinned version and skips.

ADR-078 asks for `unknown` here. Line 41: "A live webhook backend alone does not establish external JobSet controller ownership", "A rule match alone does not establish a controller", "Correlate all matches with release and live-controller evidence". The Rationale names this outcome as "the silent post-reset failure the original proposal produced".

To keep the scope honest, the neighbouring states do fail closed. A JobSet installed by its own Helm release gives two owners and `unknown` at `:181`, and a `kubectl apply` install gives a mixed owned/unowned set and `unknown` at `:176`. The unsafe case is precisely the one where no JobSet controller exists.

No fixture covers it. Of the 19 directories under `pkg/setup/testdata/jobset-ownership/`, `consumer-only-role/` holds only the Trainer role excluded at `:284`, and `external-controller/` uses a `verbs: [get, list, watch]` ClusterRole with a Deployment whose name is never checked, which is shape-identical to a consumer.

What I would want: positive evidence that the candidate release actually serves the JobSet API before selecting `external`, and a fixture modelled on Kueue's rendered output asserting `unknown`.

pkg/setup/jobset_ownership.go
```diff
@@ -0,0 +1,649 @@
...
+		required[key] = true
+		refs = append(refs, finding.serviceRefs...)
+		tokenParts = append(tokenParts, key+"/"+string(finding.object.GetUID()))
+	}
+	for key, present := range required {
+		if !present {
+			return "", fmt.Errorf("missing %s", key)
+		}
+	}
+	if len(refs) != 2 {
```
ndipebot: This count is the first of two reasons `verifySupportedNonHelmJobSetController` cannot match a real JobSet install.

`refs` accumulates `finding.serviceRefs` for every fingerprinted configuration at `:226`, and `webhookServiceRefs` (`:428-443`) appends one key per `webhooks[]` entry with no dedup. So this counts webhook entries, not distinct Services, and every install channel renders two per configuration:

1. Upstream `v0.12.0` `manifests.yaml`: `mjobset.kb.io` plus `mpod.kb.io`, and `vjobset.kb.io` plus `vpod.kb.io`.
2. The jobset 0.11.0 chart that Trainer 2.2.1 pins: `mutate-jobset-x-k8s-io-v1alpha2-jobset.x-k8s.io` and `mutate--v1-pod.x-k8s.io` against the same service.

Four refs either way, so `!= 2` rejects it whatever the names are.

The second reason is the identities. `jobSetControllerName` is `jobset-controller` (`setup.go:63`), a chart identity, but `jobset.labels` in the chart's `_helpers.tpl` unconditionally emits `app.kubernetes.io/managed-by: {{ .Release.Service }}`. So a `helm template | kubectl apply` install, which does reach this function since it leaves no `meta.helm.sh/*` annotations, is rejected by `hasAnyHelmOwnershipMetadata` at `:222`. The genuinely non-Helm channel, `kubectl apply -f .../manifests.yaml`, ships `jobset-manager-role` and `jobset-controller-manager`, which are not in the required set at `:214-216` and fail the `Get` at `:249`.

Net effect for an operator on the standard kustomize install: the first loop iteration returns `unsupported fingerprint ClusterRole/jobset-manager-role`, the verify error is discarded, and `:176` selects `unknown`, so the dependency phase fails with `[deps] JobSet ownership is unknown: unverified non-Helm or incomplete ownership evidence: ...`.

That fails closed, and it is not a regression: `main` had no detection at all and that operator collided at the Helm level instead. The Helm-owned external path works and `test/uat/setup` covers it. The problem is narrower, that this branch can only ever execute against its own fixture, and the fixture cannot reveal it: `testdata/jobset-ownership/supported-nonhelm-controller/input_objects.yaml` declares one webhook per configuration and carries no labels, which neither channel produces. ADR-078 asks for "realistic chart-derived fixtures" (line 135) and for testing "the detector against the published chart's actual resource identities" (line 53). The names here are chart-accurate but the webhook shape and label set are not.

Suggestion: accept both the kustomize and the chart identities, replace the count with a check that every collected ref resolves to the same verified webhook Service, and add the doc comment line 39 asks for saying which install this pattern targets.

pkg/setup/helm.go
```diff
@@ -455,9 +500,40 @@ func classifyHelmInstallFailure(output, namespace string) failureClass {
...
+		{"customresourcedefinition", jobSetCRDName},
+		{"clusterrole", jobSetControllerName},
+		{"clusterrolebinding", jobSetControllerName},
+		{"validatingwebhookconfiguration", "jobset-validating-webhook-configuration"},
+		{"mutatingwebhookconfiguration", "jobset-mutating-webhook-configuration"},
+	}
+	for _, object := range known {
+		// Bind the object to Helm's ownership-error clause on the same line.
+		// A debug mention elsewhere must not reclassify an unrelated collision.
+		pattern := `(?im)^Error: (?:INSTALLATION|UPGRADE) FAILED: (?:Unable to continue with (?:install|update): )?` +
```
ndipebot: `FAILED: ` is mandatory in this pattern, and our fresh-install path never produces it.

We always invoke `helm upgrade --install` (`:269`). In Helm 3 that command's install branch does `rel, err := runInstall(...); if err != nil { return err }` with no wrapping (`cmd/helm/upgrade.go:160-162`). The `INSTALLATION FAILED` wrap lives in `newInstallCmd`'s `RunE` at `cmd/helm/install.go:158`, not inside `runInstall`, which is declared at `install.go:224`. `pkg/action/install.go:357` wraps the ownership conflict with `Unable to continue with install`. So the transcript we actually capture on a fresh install is:

```
Error: Unable to continue with install: ClusterRole "jobset-controller" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"
```

I compiled this exact pattern and ran it against that string. No match. The bare `helm install` form matches.

So `trainerActionInstall` at `setup.go:462` and the recovery reinstall at `setup.go:706` both fall through to `failureClassOther`, `printTrainerFailureGuidance` prints nothing, and the operator gets the raw Helm output with no `jobset.install=false` remedy. Only the retry arm still classifies, which happens to be the arm the goldens were captured from: the reproduction headers in `standalone-jobset-trainer-collision/input.yaml` and `real-clusterrole-collision/input.yaml` record bare `helm install`, and `helm3-upgrade-collision` records bare `helm upgrade`. None was captured with `upgrade --install`.

ADR-078's Diagnostics row wants guidance "on fresh install, retry, and recovery reinstall". Making the `FAILED: ` segment optional covers it, and a fourth fixture captured from `helm upgrade --install` would pin it so this cannot drift back.

pkg/setup/setup.go
```diff
@@ -432,39 +540,67 @@ func handleTrainerInstallFailure(sp setupPhaseParams, output string, installErr
 		_, _ = fmt.Fprintln(sp.out)
 	}
 
+	preUninstall, blockers := recoveryGate(sp, &baseline, false)
+	blockers = append(blockers, compareRecoveryEvidence(baseline, preUninstall, false)...)
+	if len(blockers) != 0 {
```
ndipebot, line 548: This is the only refusal exit in the recovery arm that returns without calling `printManualTrainerRecovery`. The other five (`:505`, `:515`, `:527`, `:537`, `:556`) all print it, and the helper's own doc comment at `:735` still says "It is printed on every fail-fast path".

So an operator who has already confirmed the destructive plan and is then refused by the post-confirmation gate gets the blocker list and the raw Helm error, no procedure, and a still-broken Trainer release. Three committed goldens pin the gap: `protected-created-during-confirmation`, `namespace-replaced-during-confirmation` and `protected-api-disappears` all end at the blocker list.

The gate itself fails closed, no uninstall and no CRD deletes, so this is internal consistency and UX rather than an ADR conformance issue.

pkg/setup/setup.go
```diff
@@ -495,47 +639,78 @@ func trainerRecoveryGate(ctx context.Context, c client.Client) (bool, []string)
...
-	ns.SetKind("Namespace")
-	ns.SetName(trainerNamespace)
-	switch err := sp.c.Delete(sp.ctx, ns); {
-	case apierrors.IsNotFound(err):
-		_, _ = fmt.Fprintf(out, "  Namespace %s already deleted.\n", trainerNamespace)
-	case err != nil:
-		return fmt.Errorf("delete namespace %s: %w", trainerNamespace, err)
-	default:
-		WaitForNamespaceDeletion(sp.ctx, sp.c, trainerNamespace, out)
+	if err := deleteNamespaceWithUID(sp, baseline.namespaceUID); err != nil {
```
ndipebot, line 726: `deleteNamespaceWithUID` returns `nil` without issuing a Delete when the inspected UID is empty (`recovery_inventory.go:464-470`), which is the right fail-closed call given there is no precondition to attach. But the messages around it are unconditional, here at `:700` and again at `:711-713`, so when `kubeflow-system` was already absent at the baseline gate we print a deletion that never happened. `collectRecoveryEvidence` returns early on NotFound with `namespaceUID` unset (`recovery_inventory.go:47-49`).

`testdata/init-recovery/pending-upgrade-conflict-recovers/` pins the contradiction: no `seedRecovery`, no Namespace in `input_objects.yaml`, and `expected.txt:17` and `:20` both record the deletion anyway. The fixture's own comment at `input.yaml:3-4` says "the recovery reports it as already deleted", which is output we do not emit.

Gating both messages on a non-empty UID, something like "namespace kubeflow-system is already absent", would make the transcript and that comment true. ADR-078:130 wants recovery plans, manual guidance and retained-resource output to agree with actual actions.

The precondition itself is fine and is exercised: `failed-conflict-safe-recovers` ships a real Namespace with a UID, `verified-release-inventory-recovers` seeds one, and `namespace-replaced-during-confirmation` asserts the UID-change blocker.

pkg/setup/setup.go
```diff
@@ -1158,7 +1376,7 @@ func discoverResourcesByGroup(ctx context.Context, c client.Client, apiGroup str
 			}
 		}
 		if apiVersion == "" {
-			continue
+			return nil, fmt.Errorf("CRD %s in group %s has no served version", item.GetName(), apiGroup)
```
ndipebot: This hardening is what makes the recovery gate at `:597` fail closed, which is the decision-4 goal, and `init-recovery/unserved-trainer-crd-refuses` covers it. My concern is the other caller.

`discoverResourcesByGroup` is shared with `discoverNVCREResources` (`:1320`), which feeds `deleteNVCRECRs` (`:1203-1206`), and `RunReset` runs `[cr]` first and returns on its error (`:881-886`). So an NVCRE CRD whose every `spec.versions` entry has `served: false` now aborts `setup reset` with `[cr] discover NVCRE resources: CRD <name> in group nvcre.nvidia.com has no served version`, before the `[helm]` phase that would have deleted that CRD. Kubernetes permits that state, since apiextensions requires exactly one storage version but no served one, so it takes a half-applied or deliberately disabled CRD to reach.

Small blast radius: reset stops before deleting anything, the message names the CRD, and `--skip-phases=cr` is a working escape. ADR-078 does not ask for this either way, and the reset contract at line 91 points the other direction for the retained-CRD lookup.

A strict variant used only by the recovery caller would let the `[cr]` phase keep its previous skip. If the stricter reset behavior is intended, worth saying so and adding a `[cr]` fixture for it.

.github/workflows/uat.yml
```diff
@@ -29,6 +29,50 @@ permissions:
...
+
+      - name: Run isolated setup lifecycle UAT
+        run: make test-uat-setup
+
+      - name: Upload setup lifecycle diagnostics
+        if: failure()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: setup-uat-diagnostics-${{ github.run_id }}
+          path: ${{ env.SETUP_UAT_DIAGNOSTICS }}
```
ndipebot, line 77: `SETUP_UAT_DIAGNOSTICS` is written only inside `cleanup()` in `hack/test-uat-setup.sh:34`, and the trap that arms it is not installed until `:47`. Any failure before that leaves it unset, including failures in the earlier steps that never reach the script at all: checkout, setup-go, setup-helm, Install UAT tools, Prep system for Kind.

`upload-artifact` reads `path` via `core.getInput(Inputs.Path, {required: true})` before it looks at `if-no-files-found`, so this step then fails with `Input required and not supplied: path` rather than warning, and we get a second red step that tells us nothing.

Setting a default in the script before the trap, or adding `env.SETUP_UAT_DIAGNOSTICS != ''` to this step's `if`, covers it.

ndipebot (changes requested):
I went through this against ADR-078 decision by decision, including the validation table, and I checked the detector against the real Kueue chart, the upstream JobSet manifests and Helm's own source rather than reasoning about what they probably do.

The shape of the implementation is right. The five-state model, the CRD-first probe, unconditional JobSet CRD retention, the UID-based recovery inventory, the rechecks at every destructive boundary and all four required UAT scenarios are there, and the fail-closed instinct is good in most places. I found one fail-open that inverts the point of the PR, and two places where implemented code cannot match the input it was written for. Those three are inline, along with four smaller ones.

Two things I checked and want to record as fine so nobody re-reviews them:

1. Pinning `listJobSets` to `v1alpha2` is correct. A CRD serving something else fails the list and lands on `unknown`, which is what the bounded-detection contract asks for.
2. The `pending[:0]` in-place filter in `collectRecoveryEvidence` is correct in both places. The write index never passes the read index.

## Validation-table coverage

Everything below is an ADR-078 validation row with no fixture behind it. The code is fail-closed in each case, so none of this is a live defect, but each one would let a real regression land green. I checked these by mutation rather than by grep: I deleted or inverted the production code and re-ran `go test ./pkg/setup/`, and it stayed green.

1. **The ownership guidance text is unasserted.** The eight `jobset-ownership-failure` goldens pin only the classifier boolean. The three ownership-specific lines in `printTrainerFailureGuidance` (`jobset_ownership.go:645`, `:646`, `:648`), including the `jobset.install=false` remedy and the older-CLI `setup reset --skip-phases=deps` note, appear in no golden on any of the three call sites. Deleting all three `Fprintln` calls leaves the package green. The negative direction of the `:642` guard is pinned by `recovery-reinstall-fails-once`, so it is the positive direction and the emitted text that are missing. Partly blocked on the `helm.go:527` comment below: until the classifier can fire on `helm upgrade --install`, only the retry site at `setup.go:488` can be pinned. The shared fallback lines are covered by `recovery-required-list-fails` and `unknown-state-refuses`, so the gap is these three lines and not guidance in general.

2. **Recovery with a JobSet present is untested in both directions.** Nothing puts a JobSet instance in the cluster during recovery. `createRecoveryMutation` (`init_recovery_test.go:360`) offers only foreign-configmap, trainingruntime and replace-namespace arms, so "a new bundled JobSet blocks further cleanup" and "Bundled JobSets block controller removal" are unverified: dropping `sp.jobSetMode == jobSetModeExternal &&` from the blocker at `setup.go:626` leaves the suite green. Related, the external-mode arm of ownership revalidation is executed by nothing. `ownershipToken` is compared only under `sp.jobSetMode == jobSetModeExternal` at `setup.go:572` and `:669`, `grep ownershipToken pkg/setup/*_test.go test/uat/setup/*.go` comes back empty, and no `init-recovery` fixture selects external mode. The token is correct today, since all three builders sort and every part embeds a UID, but both regression directions pass: an unstable token aborts every legitimate external recovery, and a token without UIDs lets a replaced external controller through immediately before `helm uninstall`. "Selected mode survives uninstall/reinstall" is therefore proven for bundled and crd-only only.

3. **The protected-resource list failure never reaches the inventory.** The one `listFailureKind` fixture, `recovery-required-list-fails`, is consumed by the ownership probe at `jobset_ownership.go:314` and aborts with `installCalls: 0`, so `collectRecoveryEvidence`'s own fail-closed branch at `recovery_inventory.go:66-68` never executes. Replacing it with `continue` leaves the package green. The adjacent no-UID blocker at `:79` is unreached for the same reason, so both fail-closed branches on the path that decides whether to uninstall a release and delete a namespace are unexercised. The discovery half of the row is covered, by `recovery-discovery-fails`. A fixture with `extendedDiscovery: true` and `listFailureKind` on a kind the ownership probe never lists closes it: PVC, Pod, Lease and ReplicaSet all qualify, since `knownJobSetNamespacedLeftovers` covers only ServiceAccount, Service, ConfigMap, Secret, Deployment, Role and RoleBinding.

4. **`asymmetricChartRefsWarning`'s new arm is never taken.** None of the five fixtures under `testdata/asymmetric-chart-refs-warning/` sets `skipHelm`, even though the harness has the field (`helm_test.go:125`, `:134`). Deleting `case nvcreOnGHCR && helmSkipped` at `helm.go:620`, or rewriting it to the unsatisfiable `nvcreOnGHCR && helmSkipped && trainerOnGHCR`, leaves all five passing. The UAT does not reach it either, since its `init --skip-phases=helm` run uses default refs and exits at the first `nvcreOnGHCR == trainerOnGHCR` arm. The deps analogue has a dedicated `trainer-ghcr-but-deps-skipped` fixture; this one wants `nvcre-ghcr-but-helm-skipped`.

5. **`trainerAPIsRemoved` is unreachable under test.** `recoveryDiscoveryStub` never lists a `trainer.kubeflow.org` resource and `compareRecoveryEvidence` is only ever called directly with `false`, so both arms (`recovery_inventory.go:61` and `:396`) are dead in tests, even though production discovery via `ServerPreferredNamespacedResources` makes that arm load-bearing on every real recovery.

6. **No case covers a descendant surviving its owner's deletion.** No fixture removes a verified owner. In `verified-release-inventory-recovers` the controller Deployment survives the uninstall stub, because `init_recovery_test.go:187-207` deletes only bundled-Helm-owned Secrets and the fixture's Secret has no `app.kubernetes.io/managed-by: Helm` label. So a regression in the prior-evidence arm would be masked by the still-live fixpoint loop and the golden would not move. After a real `helm uninstall` the Deployment is gone and the final gate at `setup.go:695` depends entirely on `matchesPriorEvidence`; a regression there aborts recovery with "not a verified Trainer release resource or UID-linked descendant" after the CRDs are already deleted.

7. **CRD-first probe minimality is only half pinned.** The fresh-absent fixture asserts mode, evidence and `manifestCalls: 0`, but installs no List interceptor, so nothing proves the namespaced leftover lists and the cluster-wide JobSet list are skipped. Injecting both above the `!crdPresent` gate leaves the suite green.

8. **The phase confirmation summary has no test.** `printPhaseList`'s `(skipped)` output at `setup.go:1060` is unreachable from tests, since `RunInit` and `RunReset` are never invoked and the UAT always passes `--auto-approve`. So "confirmation summary marks exactly the phases that will run" is unverified. It cannot disagree with execution today, because both read the same skip map.

## Docs

9. The cobra `Long` strings went stale. `init`'s help at `setup.go:122-125` still says a wedged release "is recovered automatically when no Trainer workloads exist, or the manual procedure is printed". Recovery now also requires a clean `kubeflow-system` protected-resource inventory and unambiguous JobSet ownership, and the post-confirmation refusal prints no procedure (see the `setup.go:545` comment). `reset`'s help at `:813` still says the retained list carries "the cleanup command for each", which the new JobSet CRD entry deliberately does not.

10. One of the three documentation clauses in ADR-078:131 did not land. `docs/cli-reference/setup.md:55`, `docs/getting-started/install.md:50` and `docs/operations/deployment.md:47` all cover "the schema is not refreshed" and the operator's responsibility, but none of them names automatic shared-CRD reconciliation as deferred work needing its own design. That language exists only inside the ADR.

kaynetu:
Addressed in `264b5f6`. Evidence for each validation-table item:

1. Ownership guidance is asserted at all three call sites: [fresh install](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/init-recovery/fresh-install-jobset-ownership-guidance/expected.txt), [retry](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/init-recovery/retry-jobset-ownership-guidance/expected.txt), and [recovery reinstall](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/init-recovery/recovery-reinstall-jobset-ownership-guidance/expected.txt).
2. Recovery now covers a bundled JobSet blocker, a JobSet created during cleanup, stable external-mode recovery, and external-controller UID replacement: [fixtures](https://github.com/NVIDIA/cluster-readiness-engine/tree/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/init-recovery).
3. Inventory failures are pinned after ownership inspection: [protected list denial](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/init-recovery/inventory-protected-list-fails/expected.txt) and [allowed object without UID](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/init-recovery/protected-object-without-uid-blocks/expected.txt).
4. The skipped-Helm warning arm is pinned by [nvcre-ghcr-but-helm-skipped](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/asymmetric-chart-refs-warning/nvcre-ghcr-but-helm-skipped/expected.json).
5. The recovery discovery stub now exposes a Trainer resource and stops serving it after CRD deletion. This takes both `trainerAPIsRemoved` branches in [init_recovery_test.go](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/init_recovery_test.go).
6. The uninstall stub removes the verified Deployment. Its surviving ReplicaSet and Pod must pass through prior UID evidence in [verified-release-inventory-recovers](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/init-recovery/verified-release-inventory-recovers/expected.txt).
7. [TestAbsentProbeSkipsNamespacedAndJobSetLists](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/jobset_ownership_test.go) fails on any forbidden namespaced or JobSet list and requires exactly the three fingerprint lists.
8. [TestPhaseSummaryMatchesGatedPhases and TestPrintPhaseListMarksExactlySkippedPhases](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/setup_test.go) pin the accepted phases and `(skipped)` output.
9. The [init and reset help](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/setup.go) now states all recovery gates, late partial-cleanup behavior, no rollback, and the warning-only JobSet CRD entry.
10. Automatic shared-CRD reconciliation is explicitly deferred in the [CLI reference](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/docs/cli-reference/setup.md), [installation guide](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/docs/getting-started/install.md), and [deployment guide](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/docs/operations/deployment.md).

`make manifests generate`, `make lint-fix`, `make build`, and `make test` pass at `264b5f6`. The full integration suite passed in 190 seconds.

pkg/setup/jobset_ownership.go
kaynetu: Fixed in `264b5f6`.

External mode now requires JobSet chart identity on both the admission configuration and the webhook-backing Deployment. A same-release Service and selector match are no longer sufficient.

The Kueue-derived [external-consumer-release](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/jobset-ownership/external-consumer-release/expected.json) and [nameOverride=jobset](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/jobset-ownership/external-consumer-name-override/expected.json) fixtures both resolve to `unknown`. [external-lookalike-deployment](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/jobset-ownership/external-lookalike-deployment/expected.json) pins the backing-Deployment check.

`go test ./pkg/setup/ -count=1` and the full `make test` pass.

pkg/setup/jobset_ownership.go
kaynetu: Fixed in `264b5f6`.

The non-Helm verifier now deduplicates webhook references and requires every webhook to resolve to one verified Service. It accepts the upstream kustomize identities `jobset-manager-role` and `jobset-controller-manager`, plus the chart identities.

[supported-nonhelm-controller](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/jobset-ownership/supported-nonhelm-controller/input_objects.yaml) mirrors the published v0.12.0 manifest, including two webhooks per configuration and kustomize labels. [supported-nonhelm-chart-identities](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/jobset-ownership/supported-nonhelm-chart-identities/expected.json) covers the other supported identity set. [nonhelm-split-webhook-services](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/jobset-ownership/nonhelm-split-webhook-services/expected.json) fails closed.

The verifier’s doc comment names both supported channels and explains why `helm template | kubectl apply` remains ambiguous.

pkg/setup/helm.go
kaynetu: Fixed in `264b5f6`.

The `<VERB> FAILED: ` segment is now optional. The classifier also accepts either Helm release annotations or the managed-by label while retaining the concrete JobSet kind/name binding.

[helm3-upgrade-install-fresh-collision](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/jobset-ownership-failure/helm3-upgrade-install-fresh-collision/input.yaml) is the captured `helm upgrade --install` fresh-install transcript. [helm3-managed-by-label-collision](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/jobset-ownership-failure/helm3-managed-by-label-collision/input.yaml) pins the label-only form. [unrelated-managed-by-label](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/jobset-ownership-failure/unrelated-managed-by-label/expected.json) remains false.

Fresh install, retry, and recovery reinstall now each have a guidance golden.

pkg/setup/setup.go
kaynetu: Fixed in `264b5f6`.

The post-confirmation refusal now calls `printManualTrainerRecovery` after printing its blockers.

The updated [protected-created-during-confirmation](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/init-recovery/protected-created-during-confirmation/expected.txt), [namespace-replaced-during-confirmation](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/init-recovery/namespace-replaced-during-confirmation/expected.txt), and [protected-api-disappears](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/init-recovery/protected-api-disappears/expected.txt) goldens all include the operator-managed procedure. They still assert no uninstall or CRD deletion.

pkg/setup/setup.go
kaynetu: Fixed in `264b5f6`.

Recovery now prints “Namespace kubeflow-system is already absent; no deletion was required” when the baseline UID is empty. The completion summary uses the same outcome. A real UID still takes the preconditioned delete path.

[pending-upgrade-conflict-recovers](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/testdata/init-recovery/pending-upgrade-conflict-recovers/expected.txt) pins the absent case. Existing seeded-namespace and UID-replacement cases continue to pin deletion and refusal.

pkg/setup/setup.go
kaynetu: Fixed in `264b5f6`.

Discovery is now split by caller. `discoverServedResourcesByGroup` fails closed for recovery. `discoverResourcesByGroup` skips unserved CRDs for reset’s explicit `[cr]` cleanup.

[TestUnservedCRDBlocksRecoveryButNotResetCR](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/pkg/setup/reset_deps_test.go) asserts both contracts. The existing `unserved-trainer-crd-refuses` golden still proves recovery stops before uninstall or CRD deletion.

.github/workflows/uat.yml
kaynetu: Fixed in `264b5f6`.

[hack/test-uat-setup.sh](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/hack/test-uat-setup.sh) publishes `SETUP_UAT_DIAGNOSTICS` before the preflight checks and before the cleanup trap.

[uat.yml](https://github.com/NVIDIA/cluster-readiness-engine/blob/264b5f6d97f37b4b3c17ee4350c80494a02727f6/.github/workflows/uat.yml) also requires `env.SETUP_UAT_DIAGNOSTICS != ''` before invoking `upload-artifact`. Failures before the script therefore do not create a second missing-`path` failure.

ndipebot (approved):
All seven inline findings and the ten validation-table items are addressed. I re-checked each against the code and mutation-tested the new fixtures: inverting or deleting each fix now fails a test.

PR #348, opened 2026-09-15
feat: add ADR for cordoned node selection

natherz97:
## Summary

This PR adds an ADR which covers adding support for running NVCRE against cordoned nodes. This ADR is being proposed to allow automatically running validation from NVCRE against new nodes or nodes returning from remediation from NVSentinel. For both of these use-cases, an operator may choose to leverage a node cordon rather than an explicit taint. As a result, it would be beneficial to add support to NVCRE for this use-case rather than require operators to swap any existing cordon with a taint prior to running validation.

The end goal of this effort is to allow fault-quarantine in NVSentinel to request a Certification from NVCRE prior to uncordoning any node which needs post-remediation validation. Additional details are available here: https://github.com/NVIDIA/NVSentinel/blob/main/docs/designs/049-node-validation.md

## Type of Change
- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] API / CRDs
- [X] Controller / Reconcilers
- [ ] Catalog / Workloads
- [ ] CLI (nvcrectl)
- [ ] Helm / Deployment
- [ ] Documentation / CI
- [ ] Other: ____________

## Testing
- [X] Tests pass locally
- [X] Manual testing completed
- [X] No breaking changes (or documented)

## Checklist
- [X] Self-review completed
- [X] Commits are signed off for the DCO (`git commit -s`)
- [X] `make manifests generate` run (if `*_types.go` was modified)
- [X] Golden files updated (if integration test output changed)
- [X] Documentation updated (if needed)
- [X] Ready for review


lalitadithya (approved):

docs/designs/081-cordoned-node-selection.md
```diff
@@ -0,0 +1,93 @@
...
+path between NVCRE and NVSentinel requires support for selecting cordoned nodes.
+
+## Decision
+
+Allow an operator to target a cordoned node by adding the node.kubernetes.io/unschedulable
+taint to a Certification's taintSelectors, without any changes to the Certification or
+Workflow API.
+
+1. Node discovery: discoverTargetNodes will not exclude cordoned nodes that matched a
+taintSelectors entry for the node.kubernetes.io/unschedulable taint. If this taint is not
```
lalitadithya, line 47: this would also have to extend to if the user specifies just `effect: NoSchedule` right?

docs/designs/081-cordoned-node-selection.md
```diff
@@ -0,0 +1,93 @@
...
+present in taintSelectors, the existing behavior will apply which filters out cordoned
+nodes from the set of selected nodes.
```
ndipebot, line 49: `nodeMatchesTaints` requires every selector to match (`pkg/controller/workflow_controller.go:702-709`), so once `node.kubernetes.io/unschedulable` is in `taintSelectors` the run narrows to cordoned nodes only. That's fine for the NVSentinel revalidation case this is written for, but it means one Certification can't span cordoned and uncordoned nodes, and the field ends up doing double duty as a filter and a mode switch. Worth stating outright.

On ordering: the `taintSelectors` filter runs at `:643`, before the cordon branch at `:662-673`, and it's the only one of the four filters that keeps no record of what it dropped. `kubectl cordon` sets `spec.unschedulable` and the node lifecycle controller writes the taint afterwards, so a node caught in that window fails the taint match and disappears with no event, no `excludedNodes` entry and no `INCOMPLETE` downgrade.

`docs/getting-started/first-certification.md:23` and `docs/concepts/health-monitoring-remediation.md:64` both tell operators cordoned nodes are skipped, so they'll want updating when this lands.

docs/designs/081-cordoned-node-selection.md
```diff
@@ -0,0 +1,93 @@
...
+2. Pod tolerations: no changes are required for ensuring that NVCRE test pods can execute
+against a cordoned node. Specifically, any taint specified in taintSelectors will automatically
+be tolerated by test pods.
```
ndipebot, line 52: Tolerations only go onto the workload pod templates, via `adapter.SetTolerations` at `pkg/controller/workflow_controller.go:1048`. The example sets `enableMNNVL: true` on GB200, which pulls in `_lib/deps/gb200-compute-domain-and-dra-comm.yaml` and creates a `ComputeDomain`; its IMEX pods are applied as-is and get nothing from us. They're DaemonSet-managed so the cordon taint isn't actually a problem, but "no changes are required" reads broader than what the injection covers.

Separately, setting `taintSelectors` drops the blanket `Operator: Exists` toleration (`:1040-1053`), and selection only requires the listed taints to be present, not that no others are. A cordoned node carrying an unrelated taint gets selected and then can't schedule the test pods. Already documented on the field, but it belongs in Consequences here.

docs/designs/081-cordoned-node-selection.md
```diff
@@ -0,0 +1,93 @@
...
+3. Health monitor default: if taintSelectors specifies the node.kubernetes.io/unschedulable
+taint, we will skip applying the default NodeHealthMonitor which results in HardwareFailure 
+for any cordoned node. If this taint is not present in taintSelectors, the default
+monitor CEL expression will continue to be applied.
```
ndipebot, line 56:
```suggestion
3. Health monitor: if taintSelectors specifies the node.kubernetes.io/unschedulable
taint, the cordon health check will be dropped for the run. This means clearing
nodeHealthMonitor on the created Job even when the catalog jobTemplate sets it, not
only skipping the nil default, since every catalog entry sets the
node.spec.unschedulable == true expression explicitly. If this taint is not present in
taintSelectors, the default monitor CEL expression will continue to be applied.
```

docs/designs/081-cordoned-node-selection.md
```diff
@@ -0,0 +1,93 @@
...
+apiVersion: nvcre.nvidia.com/v1alpha1
+kind: Certification
+metadata:
+  name: gpu-node-1-certification
+spec:
+  target:
+    nodeSelector:
+      nvidia.com/gpu.product: NVIDIA-GB200
+    nodeNames:
+      - gpu-node-1
```
ndipebot, line 69: `api/v1alpha1/workflow_types.go:125-126` says `nodeNames` makes `nodeSelector` ignored, but the code ANDs them: `workflow_controller.go:606` lists by label and `:628` then filters that result by name. The example depends on the AND behavior, so it's correct as written and the doc comment is the thing that's wrong. This is the first place we lean on it though, so probably worth fixing the comment at the same time.

docs/designs/081-cordoned-node-selection.md
```diff
@@ -0,0 +1,93 @@
...
+
+**Status:** Proposed
+
+**Date:** 2026-09-15
+
+## Context
+
+Currently, there is no mechanism which supports running a Certification or Workflow which
+targets a cordoned GPU node. Specifically, the workflow-controller will ignore any cordoned
+node in discoverTargetNodes even if the node satisfies all other selection requirements from
```
ndipebot, line 11: Trailing whitespace on 11, 20, 23, 27, 30, 54 and 82, and no newline at end of file. Every other file in `docs/designs/` is clean on both.

ndipebot (changes requested):
Item 3 doesn't change anything on the Certification path. `job.Spec` is a deep copy of the catalog `jobTemplate.spec` (`pkg/controller/workflow_controller.go:981`) and the default monitor only gets applied when that copy comes back nil (`:1061`). Every catalog entry that defines a `jobTemplate` sets `nodeHealthMonitor` explicitly, all eight of them, including `pkg/catalog/entries/communication/nccl-loopback.yaml:48-50`, which is the variant in the example here. So skipping the default leaves the catalog's `node.spec.unschedulable == true` in place and `gpu-node-1` still lands on `HardwareFailureDetected`. The example doesn't do what it says it does. This needs to say whether we clear an explicitly set monitor or change the catalog entries.

Missing Implementation, Rationale, Consequences, Notes and References. `077-workload-image-override.md` is the shape I'd follow. Consequences is the one I actually want: with no monitor at all, `checkNodeHealth` returns at `pkg/controller/job_controller.go:1155` before it discovers any node, so ADR-061's detection path 1 is off for the whole run, not just for the cordon we already knew about. Not much is lost in practice, since every selected node is cordoned by the time `taintSelectors` has matched and the default expression is a single clause that's true on all of them, but say it rather than leave it implied.

The approach itself looks right to me. Reading the well-known taint out of `taintSelectors` is close to what kube-scheduler does with `pod.spec.tolerations`, and ADR-063 already made `taintSelectors` a behavior trigger rather than a pure filter, so this isn't new surface. Rest is inline.

asivanadi0 (commented):
Checked HEAD `51469e81c954af95abf66a0c55e6b07ea8233b6a` against ndipebot’s CHANGES_REQUESTED review (`5224627136`, 2026-09-16 15:17 UTC). Tip is still the original ADR plus a merge-from-main; nothing since that review addresses the CR. Not approving.

## Still open (blocking)

1. **Decision item 3 is incorrect on the Certification path.** `job.Spec` is a deep copy of the catalog `jobTemplate.spec` (`workflow_controller.go:981`) and the default monitor is applied only when that copy is nil (`:1061`). All eight catalog `jobTemplate`s set `nodeHealthMonitor` explicitly (including `nccl-loopback.yaml:48-50`, the example variant). Skipping the nil default leaves `node.spec.unschedulable == true` in place, so the example still fails cordoned nodes as `HardwareFailureDetected`. ADR needs to say whether we clear an explicitly set monitor or change the catalog entries.

2. **ADR shape incomplete vs `077-workload-image-override.md`.** Missing Implementation, Rationale, Consequences, Notes, and References. Consequences is the load-bearing one: with no monitor at all, `checkNodeHealth` returns at `job_controller.go:1155` before discovering nodes, so ADR-061 detection path 1 is off for the whole run — say that explicitly rather than imply it.

## Inline gaps still unaddressed

- **Filter vs mode-switch** (`discussion_r4027664482`): once `node.kubernetes.io/unschedulable` is in `taintSelectors`, `nodeMatchesTaints` AND-matches (`:702-709`) so the run narrows to cordoned nodes only. State that. Also the taint filter (`:643`) runs before the cordon branch (`:662-673`) with no drop record — a node in the `kubectl cordon` → lifecycle-taint window disappears with no event / `excludedNodes` / `INCOMPLETE`. Operator docs that say cordoned nodes are skipped (`first-certification.md`, `health-monitoring-remediation.md`) need a follow-up callout.
- **Toleration scope** (`discussion_r4027664493`): injection is only via `adapter.SetTolerations` on workload pod templates (`:1048`). `enableMNNVL: true` pulls IMEX/`ComputeDomain` pods that get nothing from us (DaemonSet-managed, so cordon is fine, but “no changes are required” overclaims). Setting `taintSelectors` also drops the blanket `Operator: Exists` toleration (`:1040-1053`); a cordoned node with an extra unrelated taint is selected then unschedulable for test pods — belongs in Consequences.
- **`nodeNames` + `nodeSelector`** (`discussion_r4027664535`): example correctly relies on AND behavior (`:606` then `:628`), but `workflow_types.go:125-126` still says `nodeNames` makes `nodeSelector` ignored. Fix the godoc alongside this ADR.
- **Hygiene** (`discussion_r4027664540`): trailing whitespace on lines 11, 20, 23, 27, 30, 54, 82 and missing final newline — every other file under `docs/designs/` is clean.

Approach (reuse `taintSelectors` as the cordon opt-in, consistent with ADR-063) still looks right. Happy to re-review once the health-monitor semantics and Consequences section land on tip.

docs/designs/081-cordoned-node-selection.md
natherz97: My understanding is that any entry in taintSelectors requires an explicit key being set. So the only way to get a blanket toleration would be to not specify taintSelectors at all. Specific to the unschedulable taint, the 2 valid taintSelectors a user could specify would be either of these since there's no value ever applied:

No effect:
```
taintSelectors:
  - key: node.kubernetes.io/unschedulable
```
With effect:
```
taintSelectors:
  - key: node.kubernetes.io/unschedulable
    effect: NoSchedule
```
Does that answer your question?

docs/designs/081-cordoned-node-selection.md
natherz97: Thanks for catching this! So regardless of whether the NodeHealthMonitor is specified by the catalog, Workflow, or the default, we'll skip applying it if taintSelectors specifies the node.kubernetes.io/unschedulable taint. I applied this change locally in latest revision.

docs/designs/081-cordoned-node-selection.md
natherz97: For 1, I clarified that the pod toleration are applied specifically to workload pods. For 2, I added an item under the consequences section that explicitly states that taints other than the unschedulable taint also need to be listed under taintSelectors for it to be tolerated.

docs/designs/081-cordoned-node-selection.md
natherz97: For 1, I called out that we cannot target cordoned and uncordoned nodes within the same Certification under the consequences section. For 3, both of those separate docs have been updated for accuracy.

For 2, this race condition would only occur if a node was cordoned concurrently with a Certification being created so from an operator's point of view, it seems reasonable for that node to be excluded from the Certification. In other words, it would be the client's responsibility to ensure that any taint is present before starting a Certification regardless of whether that's the unschedulable taint or not. I called this out under consequences section as well.

ndipebot (approved):
/lgtm
/approve

PR #349, opened 2026-09-16
feat: implement ADR-080 phase-transition events

### LINKED ISSUE #252: [Feature]: Emit Normal events on phase transitions across all six reconcilers

kaynetu:
## Summary

Implement [ADR-080](https://github.com/NVIDIA/cluster-readiness-engine/blob/main/docs/designs/080-phase-transition-events.md) so `kubectl describe` shows lifecycle transitions for Certification, Workflow, Job, and WorkloadRun. InProgress and Succeeded transitions emit Normal Events; Failed transitions emit Warnings. Events use the persisted condition's reason and message, emit only after successful status writes, and suppress repeated phases and reason/message-only updates.

Add separate Job hardware and validation verdict Events, including `ThresholdsMet`, and the Workflow-driven `JobTimedOut` Event on the Job. Preserve concurrent terminal Job decisions during status retries. Keep action-failure diagnostics when status persistence fails, and preserve Certification's specific `WorkflowFailed` reason after rejected Workflow creation.

Add UID-scoped Event integration goldens, a bounded collection deadline, and recorder-level coverage for retries, failed writes, fallback warnings, polling-message changes, and checkpoint restart. Mark the ADR accepted and document operator-facing Event semantics. Measurement reconcilers retain their action warnings; their outcomes are reported through Job verdicts, as specified by ADR-080.

## Related Issue

Closes #252

Implements the design merged in #338.

## Type of Change

- [x] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected

- [ ] API / CRDs
- [ ] Catalog / Workloads
- [ ] CLI (nvcrectl)
- [ ] Helm / Deployment
- [x] Documentation / CI
- [x] Other: integration harness and controller regression tests

## Testing

- [x] Tests pass locally

Validation: `make manifests generate`, `make lint-fix`, `make ci`, and `git diff --check`.

The automated suite covers successful and failed lifecycle transitions, hardware and threshold verdicts, timeout/pod-drain re-entry, rejected Workflow creation, status-write failures and conflict retries, fallback recovery, checkpoint restart, and changing polling messages. Event goldens assert API persistence; recorder assertions count emissions before API aggregation.

Manual Kind + KWOK validation on this implementation:

- An H100 NCCL Certification and the H100 WorkloadRun fixture completed successfully, with one start and one completion Event per participating lifecycle tier.
- An intentionally invalid override produced one Workflow `OverrideError` and one WorkloadRun `WorkflowFailed` Warning.
- A short Job timeout produced one Job `JobTimedOut`, one Workflow `IterationsFailed`, and one WorkloadRun `WorkflowFailed` Warning.
- After five annotation-update rounds, each observed transition still had one Event row with count 1.

KWOK simulates workload completion; these checks validate lifecycle propagation and Events, not GPU execution or performance measurements. The full catalog UAT suite was not run for this controller change.

## Risk Assessment

- Events remain best-effort diagnostics. A crash after a status write can lose its Event; status remains authoritative.
- Terminal failures now emit Warning Events. Consumers must distinguish transition reasons from retained action warnings.
- The Job retry guard preserves a concurrent terminal decision instead of replacing it with a stale workload observation. The existing timeout condition shape, including simultaneous InProgress and Failed conditions, is unchanged.
- Existing `OverrideNoOp` and transient `WorkloadCreationError` warnings can still appear during successful runs; they are separate from transition deduplication.
- No CRD schema, RBAC, chart, or dependency changes.

## Checklist

- [x] Self-review completed
- [x] `make manifests generate` run (no API type changes)
- [x] Golden files updated (opted-in Event cases and dedicated regression fixtures)
- [x] Documentation updated
- [x] Ready for review


asivanadi0 (approved):
## Review (ADR-080 phase-transition events)

Tip reviewed: `01a04aaf8d6dbfd2f5fa54362b6561b7780cc58c`. CI green on this SHA (Test, Lint, UAT/Kind+KWOK, Build, Analyze Go, DCO, generated-files verify, etc.). No prior human reviews.

This is a careful implementation of Accepted ADR-080, not a drive-by Event sprinkle. Approving.

### What lines up well

- **Flip ≠ `changed`.** `conditionFlip` / `conditionTransition` ignore reason/message/`ObservedGeneration`/`extra`-only edits; exclusive wrappers and `setJobFailed` emit only when the true type changes, and only after a successful write. That is the whole point of splitting #252 from #250.
- **Conflict-retry correctness.** Transition is cleared and recomputed inside each mutate attempt (`setExclusiveStatusConditionUnless`), so a stale pre-retry flip cannot leak into an emit. WorkloadRun’s `setWorkloadRunConditionAndUpdate` mirrors the same before/after rule at its direct-update seam without forcing it onto the shared helper.
- **Decision 5 fallbacks.** Same-reason pre-write Warnings at WorkloadRun `BuildFailed`, Workflow `HeterogeneousPlatform`, and both `OverrideError` sites are moved to distinct `*StatusUpdateFailed` reasons on failed persistence; successful paths notify only via the Failed transition. Certification’s typed `workflowCreateRejectedError` stops the catch-all from rewriting `WorkflowFailed` → `WorkflowValidationFailed`, so Events and Conditions stay aligned.
- **Job verdicts + timeout hook.** `HardwareFailed` / `ValidationFailed` (including absent→False `ThresholdsMet`) and the Workflow-driven `JobTimedOut` write are hooked as specified, including emit-against-Job from the Workflow recorder and no duplicate on pod-drain re-entry.
- **Concurrent terminal preservation.** Routing Job exclusive writes through `setExclusiveStatusConditionUnless(..., r.isTerminalState)` so a conflict retry cannot replace a Workflow timeout / Succeeded winner with a stale workload observation is a real correctness win; `TestJobPhaseWritePreservesConcurrentTerminalDecision` pins it (including no Event on the discarded attempt).
- **Harness.** UID-filtered `events/v1` projections, opt-in `events:` blocks, cumulative deadline, and recorder-level counters for node-poll / checkpoint-restart match the ADR’s testing-method amendments. Goldens I spot-checked (rejected Create, threshold fail/pass, timeout, early override) match the promised shapes.

### Residual (non-blocking)

1. **Branch state:** tip is diverged from `main` (`behind_by: 1`). Rebase before merge so CI re-runs on a current base.
2. **Deferred status shape (already in ADR):** timed-out Jobs still carry `InProgress=True` beside `Failed=True`, and that write still uses a bare `Status().Update` without the shared retry helper. Fine to leave as a follow-up record; just do not treat this PR as having closed that exclusivity gap.
3. **Coverage preference (optional):** `workflow-job-timeout` waits for Workflow `Failed` but only goldens Job Events. A sibling expect row for the Workflow terminal Warning would lock the manual Kind/KWOK sequence end-to-end; other Workflow Failed cases already cover the generic path.
4. **Docs nit:** ADR narrative still illustrates Job start as `WorkloadCreated` in several places; fixtures that enter mid-flight correctly show `WorkloadRunning` as the first InProgress Event. Harmless if readers treat those snippets as the create-path happy path, not as every golden’s literal first row.

No blocker findings on design fidelity, emit-after-write safety, or dedup. Thanks for the thorough recorder + envtest matrix.

docs/operations/troubleshooting.md
```diff
@@ -8,6 +8,34 @@ description: Diagnose and resolve common issues — stuck jobs, hardware detecti
...
+```bash
+kubectl describe certifications.nvcre.nvidia.com <name> -n <namespace>
+kubectl describe workflows.nvcre.nvidia.com <name> -n <namespace>
+kubectl describe jobs.nvcre.nvidia.com <name> -n <namespace>
+kubectl describe workloadruns.nvcre.nvidia.com <name> -n <namespace>
+```
+
+Job Events also report hardware and validation verdicts. A completed workload
+first emits `WorkloadCompleted`, then emits `ThresholdsMet` or the validation
+failure reason. A hardware verdict leaves the Job phase unchanged, but its
```
ndipebot: This is unconditional but only holds when the Job has thresholds configured.

`checkPerformanceThresholds` returns at `job_controller.go:738-740` on `len(job.Spec.Thresholds) == 0`, before any call to `setJobValidationStatus`, and `reasonThresholdsMet` is written from exactly one non-test site (`job_controller.go:793`) inside that function. Both callers (`:302` and `:474`) go through the same early return.

Only the `communication/nccl-*` catalog entries set `thresholds:`. `training/nemotron5-*` and `diagnostics/dcgm-level4` set none, so for those an operator following this doc waits for a verdict Event that never arrives. The PR's own `job-trainjob-succeed` golden shows the real shape: two Events, `WorkloadRunning` and `WorkloadCompleted`, with no verdict row.

Suggest something like: "A completed workload emits `WorkloadCompleted`. If the Job configures `thresholds`, it then emits `ThresholdsMet` or the validation failure reason."

pkg/controller/status.go
```diff
@@ -74,9 +152,34 @@ func setExclusiveStatusCondition[T client.Object](
...
+// stop is checked against each retry's object before any mutations. A caller
+// can preserve a concurrent terminal decision without changing other tiers.
+func setExclusiveStatusConditionUnless[T client.Object](
+	ctx context.Context,
+	c client.Client,
+	obj T,
+	conditions func(T) *[]metav1.Condition,
+	allTypes []string,
+	conditionType, reason, message string,
+	stop func(T) bool,
```
ndipebot, line 169: This parameter and the guard below it are the right change, but ADR-080 has no record of them.

I checked this both ways before raising it. Reverting `setJobFailed` (`job_controller.go:956`) and `JobReconciler.setExclusiveCondition` (`:1380`) to the plain `setExclusiveStatusCondition` emits an Event in 4 of the 6 combinations covered by `TestJobPhaseWritePreservesConcurrentTerminalDecision`, including a `Warning WorkloadFailed` landing on a Job the Workflow had already marked `Warning JobTimedOut`. So the guard is load-bearing, not decoration: the ADR's flip-detection premise at `:149-152` does not hold here, because the Workflow's timeout write is additive and leaves `JobInProgress=True`, which makes `trueConditionType` resolve the before-type to `InProgress` so the retry does see a flip.

The issue is purely that the record contradicts the code. ADR-080:435 says "Keep `changed` and its semantics" and `:445-447` says `setJobFailed` keeps its `extra` closure "unchanged". Since this commit also flips the ADR to `Accepted`, I would like the paragraph added rather than deferred.

Two consequences worth writing down explicitly:

1. `changed` no longer implies a landed write. `wrote = false` on line 179 is what keeps `recordJobStatus` honest, and nothing pins it: deleting that line leaves the whole package and the integration suite green.
2. On the timeout path the `extra` closure never runs, so `status.failedNodes` stays empty and `nvcre_job_status` stays at `in_progress=1`. On `ccc8991` the 409 refetch re-ran the closure and produced `failedNodes` populated with the reason clobbered to `WorkloadFailed`, so the new behavior is better, but the gauge is now permanently stale instead of intermittently corrected.

docs/designs/080-phase-transition-events.md
```diff
@@ -1,6 +1,6 @@
 # ADR-080: Phase Transition Events Across the Lifecycle Tiers
 
-> **Status:** Proposed
+> **Status:** Accepted
```
ndipebot, line 3: Flipping to `Accepted` in the same commit that implements the design is the part I would push back on slightly. Most of the edits here are individually defensible, and the amendment at `:559-563` is exactly the right way to handle a testing-method substitution. But the record should describe what shipped, and right now it is missing the `stop` / `isTerminalState` guard entirely while stating the opposite at `:435` and `:445-447`. See my note on `status.go:169`.

Not asking for a rewrite, just the one paragraph before this goes to `Accepted`.

docs/designs/080-phase-transition-events.md
```diff
@@ -591,8 +609,6 @@ Both facts drive the scope decision below.
     row, with the owning Workflow's `Failed / JobValidationFailed` row in the
     same case; a pass shows `Normal / WorkloadCompleted` then `Normal /
     ThresholdsMet`;
-  - a checkpoint-restart Job case asserting the restart does not emit a second
-    InProgress event for the same Job when the phase does not flip;
   - the second `OverrideError` site (the early `applyOverrides` guard): the
```
ndipebot, line 657: Amending the checkpoint-restart bullet here is the right call and the amendment at `:559-563` explains it well. The same substitution happened to the dedup bullet at `:592-594` and was not amended.

That bullet sits under "Integration cases with `Event` collection opted in, using `testutil.TestCaseParser` goldens" (`:581-582`), but what shipped is `certification-event-progress`, which sets `verifyNodePollEvents: true`, has no `events:` key, and asserts through the in-harness `phaseCountingRecorder`. The substitution is reasonable, since `waitingForNodesMessage` embeds a `time.Since`-derived countdown and is timing-sensitive by construction. Worth a one-line amendment matching the one you already wrote here.

I should be fair about one thing: I could not actually reproduce the nondeterminism. The golden regenerated byte-identical 6 of 6 times. The countdown does move with wall clock, so drift under load is plausible, but I am inferring that rather than reporting it.

pkg/controller/job_controller.go
```diff
@@ -1106,12 +1135,18 @@ func (r *JobReconciler) setJobHardwareFailed(ctx context.Context, job *nvcrev1al
...
+			flip = &flips[0]
+		}
 		changed = changed || c
 		return c
 	})
 	if err != nil {
 		return fmt.Errorf("failed to update status: %w", err)
 	}
+	if flip != nil && flip.After.Present && flip.After.Status == metav1.ConditionTrue {
+		r.warnf(job, flip.Condition.Reason, "%s", flip.Condition.Message)
```
ndipebot, line 1148: This passes the persisted condition message straight into the Event note, and the message built at `:1304` is `fmt.Sprintf("Hardware failure detected on node(s): %v", noderesults.FailedNodeNames(allFailedNodes))`, so its length scales with the number of failed nodes.

`events.k8s.io/v1` caps `note` at 1024 bytes. Past that the API server returns 422 and client-go gives up: `event_broadcaster.go:270` logs `Server rejected event (will not retry!)` and drops it at `:272`. No Event object is ever created. The condition and the metric still carry the detail, so the data is not lost, but the `Warning HardwareFailureDetected` row an operator is looking for simply is not there.

Nothing bounds it: `status.failedNodes` has no `maxItems` and CRD `conditions[].message` allows 32768 characters.

To be fair about scope, this site is not the first to break. Three pre-existing Warnings comma-join node names and cross the cap one node earlier at 43-character EKS names: `workflow_controller.go:390-392`, `:404-406` and `:447-449`, all byte-identical on `ccc8991`. So this is a pre-existing class, not a regression. I am flagging it here because this PR generalizes the pattern to every tier and adds a new site with the lowest threshold in the repo.

A single `eventNote()` truncation helper in the shared `eventf` wrappers would cover all of it. It has to do its own `fmt.Sprintf`, since the wrappers forward `messageFmt` and `args` through to `Recorder.Eventf`, and it needs to reach all six reconcilers that call `Recorder.Eventf`, not just the four transition wrappers.

pkg/controller/workflow_controller.go
```diff
@@ -2468,6 +2479,11 @@ func (r *WorkflowReconciler) setExclusiveCondition(ctx context.Context, workflow
 	if err != nil {
 		return err
 	}
+	if transition != nil {
+		r.eventf(workflow,
+			transitionEventType(transition.NewTrueType, nvcrev1alpha1.WorkflowFailed),
+			transition.Condition.Reason, "%s", transition.Condition.Message)
```
ndipebot, line 2485: Same 1024-byte concern as `job_controller.go:1148`, and this one is the tightest in the repo. `transition.Condition.Message` here can carry `notEnoughNodesMessage` (built at `:507`), which enumerates architecture and capacity exclusions by node name. At realistic node-name lengths that crosses 1024 bytes at around 19 excluded nodes.

Which is a bit unfortunate, because a large exclusion list is exactly the case where the operator most needs the Event.

cmd/integration/testdata/reconcile/workloadrun-event-success/input_config.yaml
```diff
@@ -0,0 +1,39 @@
...
+              - type: Complete
+                status: "True"
+                reason: JobCompleted
+                message: All workers completed
+                lastTransitionTime: "2026-01-22T10:00:00Z"
+waitFor:
+  kind: WorkloadRun
+  name: event-success-run
+  namespace: default
+  condition: Succeeded
```
ndipebot, line 7: This case is timing-flaky. It failed 2 of 8 runs under `-race` (at 60.85s and 62.89s) on `timed out waiting for condition Succeeded on WorkloadRun/event-success-run`, and passed 16 of 16 without `-race`.

The root cause is more interesting than the flake. From the run log: right after the `WorkflowCreated` Event, a concurrent reconcile enters `mirrorWorkflowStatus` (`workloadrun_controller.go:137-140`), hits a `NotFound` caused by informer-cache lag on the Workflow it had just created, and persists terminal `WorkloadRunFailed / ReasonWorkflowDeleted` with the message "Workflow was deleted". Being terminal, the WorkloadRun then never mirrors the Workflow's later `Succeeded`, so the expected `Normal/WorkflowSucceeded` never arrives.

That condition write is pre-existing on `ccc8991`. What is new here is that the PR turns it into a user-visible `Warning WorkflowDeleted` that an operator will reasonably read as someone having deleted their Workflow.

Worth treating as two separate items: harden the case, and separately consider requeueing rather than declaring `WorkflowDeleted` when the Workflow was created in the same generation.

One caveat on my evidence: I have only observed this under `-race`, and no target in the repo passes `-race`. The race is timing-dependent rather than detector-dependent, so I believe it can surface without it, but I have not seen it do so.

ndipebot (commented):
Read through this closely and ran the suite locally against `01a04aa`. The design is sound and most of the implementation is exactly right: transitions are computed inside the retry callback and reset per attempt, emission sits strictly after a successful write, `transitionEventType` takes each tier's own `Failed` constant so a terminal failure cannot render as `Normal`, no user string is ever used as a format string, and Event projections are filtered by fresh UID so a name-reusing case cannot contaminate a golden. `workflowCreateRejectedError` also fixes a real status bug: on the base branch the generic catch-all rewrote the persisted reason from `WorkflowFailed` to `WorkflowValidationFailed`.

Two things I would like resolved before this merges, one I would like soon after, and a few follow-ups.

## Blocking

1. `docs/operations/troubleshooting.md:25-27` promises a verdict Event that most runs never emit. `checkPerformanceThresholds` returns at `job_controller.go:738-740` when `len(job.Spec.Thresholds) == 0`, and `reasonThresholdsMet` has exactly one non-test write site, `job_controller.go:793`, inside that function. Both callers go through the same early return, so no other path can emit it. Only the `communication/nccl-*` catalog entries set `thresholds:`; `training/nemotron5-*` and `diagnostics/dcgm-level4` set none, as does any hand-written Job or WorkloadRun. That makes this the common case rather than an edge case. The sentence needs a no-thresholds clause.

2. ADR-080 does not describe the `stop` / `isTerminalState` guard, and this commit flips the record to `Accepted`. To be clear about which way this cuts: the guard is correct and it is load-bearing. I reverted both Job call sites to plain `setExclusiveStatusCondition` and it emits an Event in 4 of the 6 combinations that `TestJobPhaseWritePreservesConcurrentTerminalDecision` covers, including a `Warning WorkloadFailed` on a Job the Workflow had already given `Warning JobTimedOut`. The ADR's flip-detection premise at `:149-152` does not hold at this seam, because the timeout write is additive and leaves `JobInProgress=True`, so `trueConditionType` resolves the before-type to `InProgress` and the retry does see a flip. The problem is only that the record says the opposite: `:435` is "Keep `changed` and its semantics" and `:445-447` says `setJobFailed` keeps its `extra` closure "unchanged; only the post-write emission is added". Please add a paragraph covering the guard and its two consequences: `changed` no longer implies a landed write, and `recordJobStatus` therefore leaves `nvcre_job_status` at `in_progress=1` for a timed-out Job permanently, since the series is only removed on Job deletion and timed-out Jobs are kept for the report.

## Strong should-fix, fine as an immediate follow-up

3. Event notes over 1024 bytes are rejected outright by the API server, so no Event object is created at all. I reproduced it end to end through the real controller path: patching `job-trainjob-fail`'s TrainJob failure message to 1116 characters produces `Server rejected event (will not retry!) ... message: Invalid value: "": can have at most 1024 characters`, and the case then fails on missing Events. client-go performs no truncation (`tools/events/event_recorder.go:66` and `:110` pass the note through verbatim) and does not retry. The class predates this PR and its worst instance is already on `main`, which is why I am not making it a gate. But this PR now routes persisted condition messages into notes at all four tiers, while CRD `conditions[].message` allows 32768 characters, and it adds a new length-scaling site at `workflow_controller.go:2482-2486`. One truncation helper in the shared `eventf` wrappers covers it. Worth noting that six reconcilers call `Recorder.Eventf`, not four: `goodputmeasurement_controller.go:1523` and `bandwidthmeasurement_controller.go:566` bypass the transition wrappers.

## Follow-ups, not gates

4. The new `workloadrun-event-success` case is timing-flaky. It failed 2 of 8 runs under `-race` on `timed out waiting for condition Succeeded on WorkloadRun/event-success-run`, and passed 16 of 16 without. See the inline note for the root cause, which is a pre-existing race this PR makes operator-visible.

5. `job-bandwidth-threshold-pass` swapping `bandwidthMeasurement.minBusBandwidthGBps` for `thresholds.busBandwidthGBps` drops the only integration coverage of the legacy `minBusBandwidthGBps` field. Worth restoring somewhere.

6. The transient "existing Workflow is being deleted; retrying" path at `certification_controller.go:585-602` produces 0 Events on `ccc8991` and 1 `Warning WorkflowValidationFailed` here. The code is byte-identical to the base and the underlying misclassification (a retry path that writes `Failed=True`) is pre-existing, so this PR only makes it audible. Worth a separate issue rather than a change here.

7. Optional test additions: a `stop: true` case in the transition golden, since the input schema already has a `conflictWinner` knob; and asserting message and count in the event readiness gate at `integration_test.go:1171-1172`, which currently keys only on `{type, reason}`.

One small correction to my own earlier read, in case it came up elsewhere: moving the `HeterogeneousPlatform` and `OverrideError` Warnings post-write does not remove a per-requeue re-emission. On the base branch that Warning already fired exactly once, because the following `setWorkflowFailed` succeeds and `reconcileJob` short-circuits on `isTerminal`. What the move actually buys is that the Warning no longer claims a `Failed` phase before the write that would persist it. Still worth doing, just for a different reason.

kaynetu:
Follow-up [#352](https://github.com/NVIDIA/cluster-readiness-engine/issues/352) tracks the pre-existing informer-ordering race behind the `workloadrun-event-success` flake: a transient Workflow cache miss can be misclassified as `WorkflowDeleted`. This PR hardens the fixture without changing production behavior; the controller fix remains scoped to that issue.

docs/operations/troubleshooting.md
kaynetu: Fixed in `2a09b58`. `docs/operations/troubleshooting.md` now says `WorkloadCompleted` is unconditional and the verdict Event exists only when thresholds are configured.

pkg/controller/status.go
kaynetu: Addressed in `2a09b58`. ADR-080 now documents `setExclusiveStatusConditionUnless`, the terminal guard, skipped `extra` callbacks, and the stale timeout gauge consequence. `TestJobPhaseWritePreservesConcurrentTerminalDecision` now pins that discarded mutations do not update the requested phase metric.

docs/designs/080-phase-transition-events.md
kaynetu: Addressed in `2a09b58`. The ADR now describes the shipped `stop` / `isTerminalState` behavior and its consequences, so it remains `Accepted`.

docs/designs/080-phase-transition-events.md
kaynetu: Addressed in `2a09b58`. The testing plan now records `verifyNodePollEvents` as an explicit recorder-based exception to API Event goldens because the countdown message is time-dependent.

pkg/controller/job_controller.go
kaynetu: Fixed in `2a09b58`. `formatEventNote` caps notes at 1,024 bytes, preserves UTF-8 boundaries, and is used by all six recorder wrappers. Unit boundary tests and a real envtest Event persistence test cover it.

pkg/controller/workflow_controller.go
kaynetu: Covered by the same fix in `2a09b58`. Workflow transition notes pass through `formatEventNote`, while the full condition message remains unchanged in status.

cmd/integration/testdata/reconcile/workloadrun-event-success/input_config.yaml
kaynetu: Hardened in `2a09b58`. The success fixture starts from persisted `InProgress` plus an existing succeeded Workflow and passed 3/3 normally and 8/8 under `-race`. The production cache-miss behavior remains separate in #352.

kaynetu:
Addressed both CodeRabbit outside-diff findings in `57305ed`: Event projections now use `count` as the final deterministic sort key, and all API operations inside ADR-080’s cumulative case budget use the deadline-derived context (including spec immutability verification and final collection). Full verification passed.

kaynetu:
CodeRabbit’s latest outside-diff notes on `57305ed`:

**Recover the existing Workflow after the status write fails** (`workloadrun_controller.go:122-125`) — valid stuck path, not introduced here. Create can succeed and the status write can still fail, so `workflowRef` never lands. The next reconcile hits `AlreadyExists` at `:106-108` and only requeues; that branch is unchanged since `908cb14c`. Blindly adopting the existing Workflow is not the fix: Certification already refuses foreign name collisions (`certification_controller.go:582-604`). Tracked in #354.

**Guard hardware status updates after refetch** (`job_controller.go:1115-1164`) — not applying. The exclusive-phase `isTerminalState` guard is for InProgress/Succeeded/Failed writes. ADR-080 says hardware and validation verdict writes stay independent of that guard, and `Succeeded` + `HardwareFailed` is the documented outcome when hardware fails during a run. Reconcile already skips *starting* a health check once the Job is terminal (`:201-203`). `setJobValidationStatus` uses the same retry shape because thresholds run after `Succeeded`; putting the phase guard on that family of setters would drop those verdicts.

asivanadi0 (approved):
## Re-review (tip `606059f`)

Prior APPROVE was on `01a04aa`. Tip moved through `2a09b58` (ndipebot findings) → `57305ed` (Event-collection determinism) → `606059f` (merge `main`). CI green on this SHA. Re-checked the prior gaps; approving again.

### Prior gaps — closed

1. **Troubleshooting Event overclaim.** `docs/operations/troubleshooting.md` now states `WorkloadCompleted` is unconditional, and that `ThresholdsMet` / validation-failure Events exist only when thresholds are configured. Matches the no-threshold goldens.

2. **ADR `stop` / `isTerminalState` record.** ADR-080 Implementation now documents `setExclusiveStatusConditionUnless`: refreshed-object guard before mutation, `changed=false` / no transition when Succeeded or Failed is already True, skipped `extra` on the discarded attempt, and the consequence that `changed` is not proof a write landed. Status stays `Accepted` with the shipped behavior described rather than contradicted.

3. **Testing-plan amendment for `verifyNodePollEvents`.** Dedup bullet now records the recorder-count exception for the time-dependent countdown message, parallel to the checkpoint-restart amendment.

4. **Event-note 1024-byte cap.** `formatEventNote` truncates with UTF-8 safety and `... [truncated]`; all six recorder wrappers (`Job`, `Workflow`, `Certification`, `WorkloadRun`, `Goodput`, `Bandwidth`) route through it with a literal `"%s"`. Unit boundary + envtest persistence coverage present. Troubleshooting documents the limit and points operators at status conditions for the full message.

5. **`workloadrun-event-success` flake.** Fixture now starts from persisted `InProgress` + an existing Succeeded Workflow and goldens only `Normal/WorkflowSucceeded`. Production informer-ordering race remains #352 — correct split.

6. **Harness determinism (`57305ed`).** Event projection sort key includes `Count`; API ops inside the cumulative Event budget use the deadline-derived context (waits, deletion, frozen-goodput, spec-immutability, final collect).

### What still looks right

Emit-after-successful-write, flip ≠ reason/message churn, conflict-retry recomputation, Job terminal preservation via `Unless`, and Decision 5 fallbacks are unchanged in substance from `01a04aa` and still hold on this tip.

### Residual (non-blocking, unchanged)

- **#352** still owns the production `WorkflowDeleted` misclassification on informer lag; this PR correctly hardens the fixture without pretending to fix the controller path.
- Timed-out Jobs can still carry `InProgress=True` beside `Failed=True` (additive Workflow timeout write). Exclusivity repair remains a separate follow-up, as previously noted.
- CodeRabbit’s “recover Workflow after status write fails” path is pre-existing / tracked (#354 per author); not introduced here.

No remaining blocker from the prior COMMENT set. Thanks for the precise follow-ups.

cmd/integration/integration_test.go
```diff
@@ -663,21 +852,23 @@ func deleteAfterWait(t *testing.T, c client.Client, specs []collectSpec) {
...
+	defer cancel()
 	interval := 500 * time.Millisecond
 
 	for _, spec := range cfg.WaitForDeletion {
+		timeout := boundedWaitTimeout(t, time.Duration(cfg.TimeoutSeconds)*time.Second, deadline)
 		require.Eventually(t, func() bool {
-			obj := getObject(context.Background(), t, c, spec)
+			obj := getObject(ctx, t, c, spec)
 			return obj == nil
 		}, timeout, interval, "timed out waiting for deletion of %s/%s", spec.Kind, spec.Name)
```
ndipebot, line 885: This wait can now pass on an object that was never deleted.

The predicate is `getObject(ctx, t, c, spec) == nil`, and `getObject` swallows every `Get` error into `nil` (see my comment further down at the `case kindJob:` arm). Before this change the call used `context.Background()`, so the only way to get `nil` was a real `NotFound`. Now that `ctx` is derived from the shared deadline, `context.DeadlineExceeded` is indistinguishable from deletion: once the budget expires, the very next poll returns `nil` and `require.Eventually` reports success.

I confirmed this against envtest. With an expired deadline-derived context the predicate returns `true` for a ConfigMap that was never deleted; the same call on `context.Background()` returns `false`.

There is no live exposure today, because no case combines `events:` with `waitForDeletion` (27 versus 16, zero overlap). The first case that does will go green without the resource ever going away. Worth noting too that reverting all six `contextForDeadline` sites leaves the entire suite green, so this wiring currently has no coverage of its own.

One line closes it:

```suggestion
		require.Eventually(t, func() bool {
			obj := getObject(ctx, t, c, spec)
			return obj == nil
		}, timeout, interval, "timed out waiting for deletion of %s/%s", spec.Kind, spec.Name)
		require.NoError(t, ctx.Err(), "deadline expired during waitForDeletion of %s/%s", spec.Kind, spec.Name)
```

ndipebot (commented):
`formatEventNote` is correct on the byte boundary: I confirmed against an envtest apiserver that a 1024-byte note is accepted and 1025 is rejected, and that reverting the six call sites reproduces the original rejection on the node-shortfall path. The `wrote = false` pin is real too: deleting that line now fails four subcases of `TestJobPhaseWritePreservesConcurrentTerminalDecision`, and restoring the pre-review test file makes the same mutation green again. On the dedup bullet, removing the guard at `pkg/controller/status.go:221` fails `certification-event-progress` with expected 1, actual 5, so the ADR text now describes something that exists.

Build and merge hygiene are clean at `606059f`: `go build`, `go vet`, the full `./pkg/...`, `./test/...` and `./cmd/...` suites with envtest, `-race`, `-shuffle=on`, `make manifests generate` a no-op, and the `main` merge byte-identical to `ccc8991..d90b5ff`.

Three of the inline comments below I would like resolved before this goes in: the `waitForDeletion` deadline hole, the WorkloadRun coverage gap, and the `workloadrun-mpi` behavior under `-race`. The remaining two are low priority.

cmd/integration/integration_test.go
```diff
@@ -862,7 +1059,7 @@ func getObject(ctx context.Context, t *testing.T, c client.Client, spec collectS
 	key := types.NamespacedName{Name: spec.Name, Namespace: spec.Namespace}
 
 	switch spec.Kind {
-	case "Job":
+	case kindJob:
 		obj := &nvcrev1alpha1.Job{}
 		if err := c.Get(ctx, key, obj); err != nil {
 			return nil
```
ndipebot: This is the half that makes the `waitForDeletion` change above unsound: every `Get` error collapses to `nil`, including `context.DeadlineExceeded`, and every arm of this `switch` does the same. That is fine for a caller that only ever asks "does this exist", but `waitForDeletion` now reads the answer as "it is gone" under an expired context.

cmd/integration/testdata/reconcile/workloadrun-event-success/expected.json
```diff
@@ -0,0 +1,75 @@
...
+      "type": "Normal",
+      "reason": "WorkflowSucceeded",
+      "message": "Workflow completed successfully",
+      "involvedObject": {
+        "kind": "WorkloadRun",
+        "name": "event-success-run"
+      },
+      "count": 1
+    }
+  ]
```
ndipebot, line 74: After the rewrite this case emits exactly one Event in total. I ran it verbose: `Normal / WorkflowSucceeded` is the only row, the `WorkflowCreated` count is zero, and `event-success-run-job` creations are zero, so the WorkloadRun to Workflow to Job to TrainJob chain no longer runs here at all.

That leaves WorkloadRun as the only tier with no integration case asserting `InProgress` and `Succeeded` together, which `docs/designs/080-phase-transition-events.md:617-619` requires per tier. The other three still have it (`certification-succeed`, `workflow-succeed`, `job-trainjob-succeed`). `workloadrun-mpi` is an `InProgress` then `Failed` path, so it does not cover the pair.

The hardening itself is real and I reproduced it: the old fixture failed 4 of 10 runs under `-race`, the new one is 12/12 clean without `-race` and 10/10 with. My objection is only to how the gap is recorded. I would rather the ADR carry this as a known gap pending `#352` than amend the required testing method, since the underlying bug is production-visible: `mirrorWorkflowStatus` treats a transient cache `NotFound` as a real deletion, and this PR promotes that into a user-facing `Warning / WorkflowDeleted`.

docs/designs/080-phase-transition-events.md
```diff
@@ -556,6 +582,33 @@ Both facts drive the scope decision below.
...
+  guard reachable in an Event integration fixture.
+- **WorkloadRun success testing-method amendment:** start the success fixture
+  from a persisted InProgress status and an existing Succeeded Workflow, then
+  assert the WorkloadRun's `Normal / WorkflowSucceeded` transition. Creating
+  both resources under the running manager exposes a pre-existing informer
+  ordering race in which the WorkloadRun can observe its new `workflowRef`
+  before the Workflow enters the cache and incorrectly persist
+  `Failed / WorkflowDeleted`. The `workloadrun-mpi` integration golden retains
+  API-level coverage of `Normal / WorkflowCreated`, while the WorkloadRun
+  recorder test covers the full InProgress-to-Succeeded sequence.
```
ndipebot: The last clause overstates what the recorder test does. `pkg/controller/status_transition_test.go:231` uses a fake client and synthetic `Started` and `Completed` reasons, never runs `Reconcile`, and would keep passing if the WorkloadRun reconciler stopped emitting `WorkflowCreated` entirely. It is not a substitute for the API-level assertion this amendment removes.

Process point as much as a content one: this paragraph was added in `2a09b58..57305ed`, the same push that weakened the fixture it legalizes. I would prefer the ADR record the gap and point at `#352` rather than redefine the requirement to match what now ships.

cmd/integration/testdata/reconcile/workloadrun-mpi/expected.json
```diff
@@ -1354,5 +1354,27 @@
...
+    {
+      "type": "Warning",
+      "reason": "WorkflowFailed",
+      "message": "Failed to partition nodes: node \"gpu-node-01\" missing topology label \"topology.k8s.aws/network-node-layer-1\"",
+      "involvedObject": {
+        "kind": "WorkloadRun",
+        "name": "test-mpi-run"
+      },
+      "count": 1
+    }
```
ndipebot, line 1368: Giving this case a golden event list makes it sensitive to the race that was moved off `workloadrun-event-success` rather than fixed.

On a full-suite run under `-race` on this branch, `workloadrun-mpi` failed with "timed out waiting for expected Events regarding `WorkloadRun/test-mpi-run`", with `Warning / WorkflowDeleted` in the log for three separate WorkloadRuns. `grep -c WorkflowDeleted` is 4 on this branch against 1 on `main`, and on `main` it appears only as a persisted condition, never as an Event. Most of the `-race` churn on my machine is pre-existing (I baselined `main` twice), but this case is newly exposed because of the block above.

If `#352` is not landing in the same release, I would either make this case's event assertion tolerant of the race or hold the `WorkflowDeleted` Warning emission behind `#352`, so operators do not see a misleading Warning in the interim.

docs/designs/080-phase-transition-events.md
```diff
@@ -591,8 +646,6 @@ Both facts drive the scope decision below.
     row, with the owning Workflow's `Failed / JobValidationFailed` row in the
```
ndipebot, line 654: Low priority, and just above this line rather than on it: the head of this bullet (lines 643-645) still says the threshold sequence runs "after `Normal / WorkloadCreated`". No golden in the repo contains a Job `Normal / WorkloadCreated` row. Both threshold fixtures pre-create the TrainJob, so `pkg/controller/job_controller.go:399` never fires. Same class of unamended substitution as the dedup bullet that was fixed, one bullet further down.

pkg/controller/event_note.go
```diff
@@ -0,0 +1,29 @@
...
+// Callers must pass the result through a literal "%s" recorder format.
+func formatEventNote(format string, args ...any) string {
+	note := fmt.Sprintf(format, args...)
+	if len(note) <= maxEventNoteBytes {
+		return note
+	}
+	end := maxEventNoteBytes - len(eventNoteTruncationSuffix)
+	for end > 0 && !utf8.RuneStart(note[end]) {
+		end--
+	}
```
ndipebot, line 36: Low priority. This walk-back crosses an entire run of continuation bytes, so invalid UTF-8 over-truncates badly: 500 ASCII bytes followed by binary collapses to 514 bytes instead of 1024, and an all-continuation-byte note collapses to the bare `... [truncated]` marker. Container and NCCL output can carry binary.

My envtest probe shows the apiserver accepts invalid UTF-8 at 1024 bytes, so the walk-back is defensive rather than required. A bounded step (stop after at most 3 bytes, or check `utf8.DecodeLastRune` on the prefix) keeps the intent without the pathological case.

The byte-boundary logic is otherwise correct, and I verified the 1024 accept / 1025 reject boundary directly against an envtest apiserver.

kaynetu:
Release scope clarification following the latest review:

- We will work on #352 immediately after finishing this PR. Keeping it in a separate PR is a scope boundary, not a decision to ship the known false-deletion behavior.
- **Do not cut a release containing #349 until both #349 and #352 have landed**, and the affected WorkloadRun lifecycle fixtures pass repeated race-instrumented runs.
- Full cache-backed WorkloadRun creation-to-success integration coverage remains a known gap pending #352. Restore that coverage as part of #352; the hardened terminal-event fixtures and fake-recorder tests are not equivalent substitutes.
- The fixture hardening in this PR does not fix the production race. We are not treating false `WorkflowDeleted` as a valid test outcome or suppressing the production Warning to hide it.

cmd/integration/integration_test.go
kaynetu: Fixed in [bbddc08](https://github.com/NVIDIA/cluster-readiness-engine/commit/bbddc083159e935d7c71383ee82483f76cd8f490). [waitForDeletion](https://github.com/NVIDIA/cluster-readiness-engine/blob/bbddc083159e935d7c71383ee82483f76cd8f490/cmd/integration/integration_test.go#L880) now succeeds only on NotFound with a live context, and checks the context after the wait. Tests for read failures and deadline propagation pass, including under -race: TestWaitForDeletionRejectsReadFailures and TestWaitForDeletionPropagatesDeadline.

cmd/integration/integration_test.go
kaynetu: Fixed in [bbddc08](https://github.com/NVIDIA/cluster-readiness-engine/commit/bbddc083159e935d7c71383ee82483f76cd8f490). [readObject](https://github.com/NVIDIA/cluster-readiness-engine/blob/bbddc083159e935d7c71383ee82483f76cd8f490/cmd/integration/integration_test.go#L1080) preserves read errors; getObject returns nil only for NotFound and fails the test on other errors. TestReadObjectPreservesErrors covers a live object, absence, cancellation, deadline expiry, Forbidden, and transport failure; it passes under -race.

cmd/integration/testdata/reconcile/workloadrun-event-success/expected.json
kaynetu: Agreed. [bbddc08](https://github.com/NVIDIA/cluster-readiness-engine/commit/bbddc083159e935d7c71383ee82483f76cd8f490) records this as a [known lifecycle coverage gap](https://github.com/NVIDIA/cluster-readiness-engine/blob/bbddc083159e935d7c71383ee82483f76cd8f490/docs/designs/080-phase-transition-events.md#L589) pending #352, not equivalent replacement coverage. We will address #352 immediately after this PR and restore the full creation-to-success assertion; the [release prerequisite](https://github.com/NVIDIA/cluster-readiness-engine/blob/bbddc083159e935d7c71383ee82483f76cd8f490/docs/designs/080-phase-transition-events.md#L922) requires both changes before release.

docs/designs/080-phase-transition-events.md
kaynetu: Corrected in [bbddc08](https://github.com/NVIDIA/cluster-readiness-engine/commit/bbddc083159e935d7c71383ee82483f76cd8f490). The [ADR](https://github.com/NVIDIA/cluster-readiness-engine/blob/bbddc083159e935d7c71383ee82483f76cd8f490/docs/designs/080-phase-transition-events.md#L596) now explicitly says the fake-client test pins helper deduplication, not Reconcile or API-level creation coverage. The full per-tier lifecycle requirement remains; WorkloadRun's gap is tracked in #352.

cmd/integration/testdata/reconcile/workloadrun-mpi/expected.json
kaynetu: Addressed the fixture exposure in [bbddc08](https://github.com/NVIDIA/cluster-readiness-engine/commit/bbddc083159e935d7c71383ee82483f76cd8f490). [workloadrun-mpi](https://github.com/NVIDIA/cluster-readiness-engine/blob/bbddc083159e935d7c71383ee82483f76cd8f490/cmd/integration/testdata/reconcile/workloadrun-mpi/input_config.yaml#L4) now runs initial reconciliation with the direct API client before manager startup; MPI construction snapshots remain unchanged, and only the WorkflowCreated Event was removed from the golden. Both affected WorkloadRun fixtures passed five fresh-process -race runs. We do not accept WorkflowDeleted as a valid outcome or suppress its Warning; #352 is the immediate follow-up and both PRs must land before release.

docs/designs/080-phase-transition-events.md
kaynetu: Corrected in [bbddc08](https://github.com/NVIDIA/cluster-readiness-engine/commit/bbddc083159e935d7c71383ee82483f76cd8f490). The [threshold testing bullet](https://github.com/NVIDIA/cluster-readiness-engine/blob/bbddc083159e935d7c71383ee82483f76cd8f490/docs/designs/080-phase-transition-events.md#L650) now states that the fixtures pre-create the TrainJob and do not assert WorkloadCreated. It describes the actual WorkloadCompleted plus validation-verdict coverage. The full integration suite passed in 215.936s.

pkg/controller/event_note.go
kaynetu: Fixed in [bbddc08](https://github.com/NVIDIA/cluster-readiness-engine/commit/bbddc083159e935d7c71383ee82483f76cd8f490). [formatEventNote](https://github.com/NVIDIA/cluster-readiness-engine/blob/bbddc083159e935d7c71383ee82483f76cd8f490/pkg/controller/event_note.go#L25) examines at most three preceding bytes and trims further only when a valid rune crosses the cutoff. [TestFormatEventNoteBinaryInput](https://github.com/NVIDIA/cluster-readiness-engine/blob/bbddc083159e935d7c71383ee82483f76cd8f490/pkg/controller/event_note_test.go#L76) covers all-continuation bytes, ASCII followed by binary, and an invalid lead byte; these retain the 1,024-byte output rather than collapsing to the marker. The binary and existing valid-UTF-8 boundary tests pass under -race.

asivanadi0 (approved):
## Re-review (tip `b8ae291`)

Prior APPROVE on `606059f` is stale. Tip moved through `bbddc08` (ndipebot follow-ups) → `b8ae291` (merge `main`). **Test SUCCESS** on this SHA ([CI Test](https://github.com/NVIDIA/cluster-readiness-engine/actions/runs/35400887449/job/105780271616)). Branch is current with `main` (`behind_by: 0`). Approving again.

### Prior gaps since `606059f` — closed in `bbddc08`

1. **`waitForDeletion` false positive.** Predicate is now `ctx.Err() == nil && IsNotFound(err)` via `readObject`, with an explicit post-wait `ctx.Err()` check. `TestWaitForDeletionRejectsReadFailures` (subprocess) pins that deadline / Forbidden reads do not satisfy deletion; `TestWaitForDeletionPropagatesDeadline` pins the shared case deadline.

2. **`getObject` error collapse.** `readObject` preserves Get / context errors; `getObject` returns nil only on NotFound and fails the test otherwise. `TestReadObjectPreservesErrors` covers Canceled, DeadlineExceeded, Forbidden, and transport errors.

3. **WorkloadRun lifecycle coverage honesty.** ADR testing plan now records the known gap pending #352: success fixture asserts only terminal `Normal / WorkflowSucceeded` from a pre-persisted InProgress + existing Succeeded Workflow; fake-client recorder test is explicitly *not* a substitute for Reconcile / API-level creation coverage. Release prerequisite text matches the author note (do not ship ADR-080 without #352).

4. **`workloadrun-mpi` race exposure.** `initializeWorkloadRun: true` runs initial reconcile on the direct API client before the manager cache starts; golden asserts only terminal `Warning / WorkflowFailed` and does not accept `WorkflowDeleted`.

5. **Threshold testing-plan wording.** Bullet now states fixtures pre-create the TrainJob and do not assert `WorkloadCreated`.

6. **`formatEventNote` binary / over-truncation.** Walk-back is bounded to `utf8.UTFMax` and only shortens when a valid rune crosses the cutoff; `TestFormatEventNoteBinaryInput` covers continuation-byte runs.

### Still holds from earlier reviews

Emit-after-successful-write, flip ≠ reason/message churn, conflict-retry recomputation inside `Unless`, Job terminal preservation (`wrote = false` / skipped `extra`), Decision 5 fallbacks, and the 1024-byte note cap across all six recorder wrappers remain intact on this tip.

### Residual (non-blocking)

- **#352** still owns the production `WorkflowDeleted` misclassification; this PR correctly scopes the fixture hardenings and documents the release gate.
- Timed-out Jobs can still carry `InProgress=True` beside `Failed=True` (additive Workflow timeout write). Exclusivity repair remains a separate follow-up.
- CodeRabbit’s “recover Workflow after status write fails” path remains pre-existing / #354.

No remaining blocker from the `606059f` COMMENT set. Thanks for the precise follow-ups.

pkg/controller/event_note.go
```diff
@@ -0,0 +1,38 @@
...
+
+// formatEventNote bounds the formatted note to the events/v1 API limit.
+// Status messages remain intact; only the recorder's copy is shortened.
+// Callers must pass the result through a literal "%s" recorder format.
+func formatEventNote(format string, args ...any) string {
+	note := fmt.Sprintf(format, args...)
+	if len(note) <= maxEventNoteBytes {
+		return note
+	}
+	end := maxEventNoteBytes - len(eventNoteTruncationSuffix)
```
ndipebot: This is the one I would like fixed before it goes in. The fix for my over-truncation comment traded it for the opposite failure.

Keeping the invalid bytes instead of walking back over them is right as far as my comment went, and the Go string is now a clean 1024 bytes. But `encoding/json` replaces each invalid byte with U+FFFD on the way out, three bytes for one, so the note the API server actually receives is larger than the string you budgeted. Measured through the real client-go JSON codec, all three inputs in your new `TestFormatEventNoteBinaryInput` go over the `events/v1` limit on the wire:

| input | wire bytes before | wire bytes now |
|---|---|---|
| all-continuation-1200 | 15 | 3042 |
| 500 ASCII + 700 binary | 514 | 2042 |
| 1008a + 0xff + continuation run | 1023 | 1026 |

The limit is 1024 (`Note` in `k8s.io/api/events/v1/types.go`). The old code stayed under it only because it threw those bytes away. Valid UTF-8 is unaffected, which is why the property tests see nothing. The awkward part is that `TestFormatEventNoteBinaryInput` now pins the over-limit output as the expected result, so this would not be caught later either.

Budgeting against the encoded form fixes it, and it keeps everything your test asserts about not collapsing to the marker:

```suggestion
	// Budget against the encoded form. encoding/json expands each invalid byte
	// to U+FFFD, so invalid input would otherwise grow up to threefold in transit.
	note := strings.ToValidUTF8(fmt.Sprintf(format, args...), "\uFFFD")
	if len(note) <= maxEventNoteBytes {
		return note
	}
	end := maxEventNoteBytes - len(eventNoteTruncationSuffix)
```

That needs `strings` added to the import block. `cmd/integration/event_note_test.go` is already the right place to prove it, since it is the only test that drives a note through a real broadcaster to a real API server, and its own comment says a fake recorder cannot detect rejection of oversized notes. It was not touched in this round and still only exercises `strings.Repeat("界", 600)`.

One caveat on my own evidence: I measured the codec round-trip, not a live rejection, because envtest could not bind in my sandbox. A narrow slice of this is also pre-existing, since `1008a + 0xC3 + ascii` is 1026 wire bytes under both the old and new code.

cmd/integration/integration_test.go
```diff
@@ -857,106 +1067,185 @@ func hasConditionWithReason(conditions []metav1.Condition, condType, reason stri
...
 
-func getObject(ctx context.Context, t *testing.T, c client.Client, spec collectSpec) client.Object { //nolint:gocyclo
+func getObject(ctx context.Context, t *testing.T, c client.Client, spec collectSpec) client.Object {
 	t.Helper()
-	key := types.NamespacedName{Name: spec.Name, Namespace: spec.Namespace}
+	obj, err := readObject(ctx, c, spec)
+	if apierrors.IsNotFound(err) {
+		return nil
+	}
+	require.NoError(t, err, "reading %s/%s", spec.Kind, spec.Name)
```
ndipebot, line 1092: Non-blocking, but worth fixing while you are here. This `require.NoError` can take down the whole test binary rather than failing one case.

Two callers reach it from inside `require.Eventually` condition closures, at `:806` and `:855`. testify runs those closures as `go checkCond()`, so `FailNow` fires off the test goroutine. That is `runtime.Goexit`, the result channel is never fed, and if the case has already finished the follow-up `t.Errorf` panics. I reproduced it with a full stack trace: `panic: Fail in goroutine after ... has completed`, frames `integration_test.go:1076` to `:806` to `runtime/asm_arm64.s:1039`, exit status 2, which loses every remaining integration case. Mutating `getObject` back to swallowing errors gives no panic and a clean exit 1, so it is specific to this change.

On reachability I want to be fair, because it cuts both ways. Typed `Get` calls through `mgr.GetClient()` route to `CacheReader.Get`, which discards the context and can only return the object or `NotFound`, so the transient-apiserver-error path is not reachable. The deadline path is: `readObject` checks `ctx.Err()` before the `Get`, and `boundedWaitTimeout` at `:772-782` makes the `Eventually` timer and the context deadline expire together by construction whenever the case deadline is the binding constraint. I did not observe this in any full-suite run, so treat it as a hazard rather than a sighting.

`waitForDeletion` already has the right shape: it calls `readObject` directly and returns `false`. The same treatment at the two other closure sites would close it.

Separately, `require.NoError(t, ctx.Err(), ...)` after the `require.Eventually` in `waitForDeletion` is dead code. `require.Eventually` calls `FailNow`, so control only reaches that line when the predicate already established `ctx.Err() == nil`. Deleting it leaves the full envtest suite green at 256.770s. It is harmless, just never going to fire, and I would rather it not read as a live guard. My fault for suggesting that exact line without checking the control flow.

cmd/integration/testdata/reconcile/workloadrun-mpi/input_config.yaml
```diff
@@ -13,3 +16,9 @@ collect:
   - kind: Workflow
     name: test-mpi-run
     namespace: default
+events:
+  - involvedKind: WorkloadRun
+    involvedName: test-mpi-run
+    namespace: default
+    expect:
+      - {type: Warning, reason: WorkflowFailed}
```
ndipebot, line 24: I am fine with the `#352` deferral and I appreciate that the ADR spells the gap out rather than burying it. One thing to record explicitly while you are updating it.

With this `expect` list reduced to `WorkflowFailed`, `Normal / WorkflowCreated` is now asserted by nothing anywhere in the repo, unit or integration. I guarded the emission at `pkg/controller/workloadrun_controller.go:238` on `condition.Reason != ReasonWorkflowCreated` and `./pkg/...` and `./test/...` both exit 0, with all 27 event-collecting fixtures passing 3/3. The coverage was real before this commit: the same mutant against the `606059f` fixture gives 0/10 against 3/10 unmutated, and three of those failures had zero `WorkflowDeleted` lines, which isolates the mutation from the race. Ten recorded envtest runs at `606059f` show the Event 10/10; at `bbddc08`, 0/10.

Your reply said only the `WorkflowCreated` row was removed, which is literally true. What it does not say is that this row was the last API-level assertion of that Event, and that the ADR at `606059f` had cited exactly this fixture as the fallback that made the round-1 removal acceptable. Adding `Normal / WorkflowCreated` to the `#352` restore list alongside the creation-to-success assertion would close the loop.

While I am here, `TestWaitForDeletionPropagatesDeadline` and the `deadline` subcase of `TestWaitForDeletionRejectsReadFailures` are vacuous. `boundedWaitTimeout(t, 1s, now+200ms)` returns about 200ms against a 500ms poll interval, so testify's timer fires before the predicate is evaluated even once. I replaced the entire predicate with `return true`, making `waitForDeletion` a no-op that declares everything deleted, and both still passed. Only the `forbidden` subcase and `TestReadObjectPreservesErrors` carry signal, and those two are solid.

cmd/integration/integration_test.go
```diff
@@ -857,106 +1067,185 @@ func hasConditionWithReason(conditions []metav1.Condition, condType, reason stri
...
-		obj := &batchv1.Job{}
-		if err := c.Get(ctx, key, obj); err != nil {
-			return nil
-		}
-		return obj
+		obj = &batchv1.Job{}
 	default:
-		t.Fatalf("unknown kind: %s", spec.Kind)
-		return nil
+		return nil, fmt.Errorf("unknown kind: %s", spec.Kind)
```
ndipebot: Minor, fold in or skip. This used to be `t.Fatalf("unknown kind: %s", ...)`, so a typo'd `kind:` in a fixture failed immediately with the reason. Now it is an error value, `IsNotFound` is false for it, and a `waitForDeletion` spec with a bad kind polls for the whole timeout instead. Probe with `Kind: "Bogus"` and `TimeoutSeconds: 4`: elapsed 4.0013095s, output `timed out waiting for deletion of Bogus/x`, and `unknown kind` appears nowhere. A fixture typo now looks identical to a controller that failed to delete.

docs/designs/080-phase-transition-events.md
```diff
@@ -364,8 +366,10 @@ Both facts drive the scope decision below.
...
    per case, so periodic refresh and idle cleanup cannot change that signal.
    Each wait uses the smaller of its configured timeout and the remaining
-   budget; fail the case if collection exceeds the deadline. Existing
-   configurable per-wait timeouts do not guarantee this total duration.
+   budget, and API calls in waits, verification, deletion, and final collection
+   carry the same deadline so a blocked call cannot outlive it. Fail the case
+   if collection exceeds the deadline. Existing configurable per-wait timeouts
+   do not guarantee this total duration.
    This rule does not apply to retained
    action or informational events, which may legitimately show `2`.
```
ndipebot, line 374: Minor, and a few lines below this rather than on it. The example block at `:381-389` still carries the same `WorkloadCreated` claim you corrected at `:650`: it shows `- { type: Normal, reason: WorkloadCreated }` paired with `JobTimedOut` for a fixture `test-timeout` that does not exist. The real fixture is `cmd/integration/testdata/reconcile/workflow-job-timeout/input_config.yaml` and it expects `WorkloadRunning`.

Related, and genuinely pre-existing rather than yours: driving the shipped fixtures through `JobReconciler.Reconcile` in four variants shows the actual suppressor of `WorkloadCreated` is the pre-set `status.workloadRef` and the guard at `pkg/controller/job_controller.go:309`, not the pre-created TrainJob that `:651-652` now names. With only the TrainJob pre-created and `workloadRef` nilled, the fixture does not run quietly, it emits `Warning WorkloadCreationError` twice and never reaches Succeeded, because `createWorkloadFromSpec` does a bare `Create` at `:375` and hits AlreadyExists. Worth a wording tweak since the ADR is meant to be the record of why.

ndipebot (commented):
Went through the round-2 fixes the same way, by reverting each one and checking whether anything actually fails. All seven are addressed and five are properly pinned. `readObject` and `getObject` in particular are clean: swallowing the error again fails `TestReadObjectPreservesErrors`, removing the `ctx.Err()` pre-check fails it too, and deleting the `NotFound` arm takes down 30 integration cases. The `workloadrun-mpi` flake is genuinely fixed rather than relocated, 15/15 clean against 2/10 at `606059f`, and running the first reconcile before `startManager` is the right shape. I also checked that the per-tier requirement at `docs/designs/080-phase-transition-events.md:625-626` is byte-identical to the old `:618-619`, so nothing was weakened to fit, and the `#352` gap and release gate are recorded honestly.

Worth saying plainly: measured against `main`, this PR is net-additive on coverage. Executed cases go 767 to 773 across `cmd/integration` and `pkg/controller`, six pure additions and zero removals.

What I am raising below is not leftover review comments. It is three things the fix commit itself introduced, one of which is in the same function my last comment was about. Only the first is blocking for me.

Verified at `b8ae291`: `go build ./...`, `go vet`, and `./pkg/controller/` all clean after the `main` merge, which touched none of the implementation files.

pkg/controller/event_note.go
kaynetu: Fixed in [258af6f](https://github.com/NVIDIA/cluster-readiness-engine/commit/258af6fdd84772e8a6f2058e49815ff99195032c). Invalid UTF-8 is normalized before the 1,024-byte budget. Binary-input tests now check JSON round trips, and TestBinaryEventNotePersistsWithinAPILimit verifies acceptance through a real broadcaster and envtest API server. Both passed with race detection; the full make test gate also passed.

cmd/integration/integration_test.go
kaynetu: Fixed in [258af6f](https://github.com/NVIDIA/cluster-readiness-engine/commit/258af6fdd84772e8a6f2058e49815ff99195032c). Both polling closures now call readObject directly and return false on read errors, avoiding getObject's fatal assertion in the polling goroutine. The post-Eventually context assertion was also removed; deadline rejection remains in the deletion predicate. Full integration tests passed (214s).

cmd/integration/testdata/reconcile/workloadrun-mpi/input_config.yaml
kaynetu: Recorded explicitly in [258af6f](https://github.com/NVIDIA/cluster-readiness-engine/commit/258af6fdd84772e8a6f2058e49815ff99195032c): the last API-level WorkloadRun Normal / WorkflowCreated assertion was removed. #352 must restore that Event assertion and the full creation-to-success lifecycle. Coverage remains deferred, not restored; both PRs must land before release.

The deadline claim does not match this checkout. Pinned Testify v1.12.1 starts checkCond immediately, before its select loop, rather than waiting for the 500ms tick. TestWaitForDeletionPropagatesDeadline uses a five-second deadline and checks the deadline captured inside Get; replacing the predicate with return true leaves that value zero. The read-failure subtests also require a nonzero child exit, so unconditional success is rejected. The deadline subtest now additionally checks a stderr marker proving Get ran, retaining the original 200ms deadline. Ten race-enabled repetitions passed.

cmd/integration/integration_test.go
kaynetu: Fixed in [258af6f](https://github.com/NVIDIA/cluster-readiness-engine/commit/258af6fdd84772e8a6f2058e49815ff99195032c). Fixture kinds are validated before deletion polling. TestWaitForDeletionRejectsUnknownKind verifies the immediate "unknown kind: Bogus" diagnostic and rejects a generic deletion timeout. The regression test passed with race detection.

docs/designs/080-phase-transition-events.md
kaynetu: Corrected in [258af6f](https://github.com/NVIDIA/cluster-readiness-engine/commit/258af6fdd84772e8a6f2058e49815ff99195032c). The ADR example now matches workflow-job-timeout: test-workflow-job-timeout-job, WorkloadRunning, then JobTimedOut. The threshold explanation now identifies status.workloadRef as the creation-path guard and distinguishes a pre-created TrainJob with no reference, which hits AlreadyExists. Documentation-link validation passed.

asivanadi0 (approved):
## Re-review (tip `893880c`)

Prior APPROVE on `b8ae291` is stale. Tip moved through `258af6f` (ndipebot round-3 hardenings) → `8419035` (merge `main` / #360–#362) → `5d94e5a` (merge #363) → `893880c` (merge #364). **Actual tip is `893880c5dc4477dd6ff3ad0e25c4c414c4a1948a`**, not the previously expected `8419035`. Branch is current with `main` (`behind_by: 0`). Full CI was green on `8419035`; checks on this tip are re-running after the #364 merge.

### Gaps since `b8ae291` — closed in `258af6f`

1. **Event-note JSON expansion.** `formatEventNote` now runs `strings.ToValidUTF8(..., "\uFFFD")` *before* the 1,024-byte budget, so isolated invalid bytes that expand threefold on the wire cannot push the note past the events/v1 limit. Binary-input unit cases assert JSON round-trips; `TestBinaryEventNotePersistsWithinAPILimit` drives a real broadcaster through envtest.

2. **Polling `FailNow`.** `waitForCondition` / finalizer-strip `Eventually` closures call `readObject` and return false on error instead of `getObject`'s fatal assertion. Unknown kinds are rejected up front via `rejectUnknownKind` / `objectForKind` (with `errUnknownKind`), and `TestWaitForDeletionRejectsUnknownKind` pins the fast-fail path.

3. **ADR honesty.** Examples now show `WorkloadRunning` for the mid-flight timeout fixture; the #352 release gate explicitly requires restoring API-level `Normal / WorkflowCreated` coverage (none remains after the MPI expect reduction).

### Still holds

Emit-after-successful-write, flip ≠ reason/message churn, conflict-retry recomputation inside `Unless`, Job terminal preservation (`wrote = false` / skipped `extra`), Decision 5 fallbacks, and the 1,024-byte note path across all six recorder wrappers are unchanged in substance from `b8ae291`.

Main merges since then (#360–#364) are deps / release / Fern register and do not touch the ADR-080 controller seams.

### Residual (non-blocking, unchanged)

- **#352** still owns production `WorkflowDeleted` misclassification and WorkloadRun creation-to-success Event coverage restoration.
- Timed-out Jobs can still carry `InProgress=True` beside `Failed=True`.
- Pre-existing Workflow-recover-after-status-write-fails path remains #354.

No blocker from the `b8ae291` COMMENT set or the `258af6f` follow-ups. Approving on tip `893880c`.

ndipebot (approved):

Issue #347, opened 2026-09-15
[Bug]: Enforce presence immutability for optional JobSpec fields


kaynetu:
### Prerequisites

- [x] I searched existing issues
- [x] I can reproduce this issue

### Code of Conduct

- [x] I agree to follow the project's Code of Conduct

### Bug Description

`JobSpec.nodeHealthMonitor`, `goodputMeasurement`, and `bandwidthMeasurement` are documented and marked as immutable, but each optional field has only a field-level CEL rule:

```go
// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="... is immutable"
```

CEL transition rules scoped to an optional field are not evaluated when that field changes between absent and present. A direct Job, or a Workflow carrying the same `JobSpec` through `spec.jobTemplate.spec`, can therefore add or remove these fields after creation even though changing an existing value is rejected.

This leaves two meanings of "immutable" in the same CRD. ADR-079 identifies the complete pattern for new optional immutable fields: a parent-level presence rule plus field-level equality.

### Expected behavior

For all three fields, admission rejects:

- absent to present;
- present to absent; and
- changes to a present value.

The rule must apply to direct Jobs and propagate through `JobTemplateSpec.Spec` to Workflows.

### Acceptance Criteria

- Add parent-level presence-invariance rules for `nodeHealthMonitor`, `goodputMeasurement`, and `bandwidthMeasurement` on `JobSpec`.
- Keep the existing field-level equality rules.
- Add envtest admission coverage for add, remove, and value-change transitions on direct Jobs and Workflow Job templates.
- Cover existing objects whose stored specs omit the optional fields.
- Regenerate and review the affected CRDs and generated code.
- Document the compatibility effect: objects created without one of these fields cannot add it later.

### Component

API / CRDs

### Steps to Reproduce

1. Create a Job without one of the three optional fields.
2. Update the Job to add the field.
3. Observe that API admission accepts the update despite the field's immutable marker.
4. Repeat by removing a present field and through `Workflow.spec.jobTemplate.spec`.

### Environment

- NVCRE version: current `main`
- Kubernetes version: envtest version pinned by the repository
- GPU architecture: not applicable
- Cloud provider / platform: not applicable
- Deployment method: any

### Logs/Output

No controller log is required; this is an API admission behavior gap.

### References

- ADR-079, Decision 6
- `api/v1alpha1/job_types.go`
[...]

PR #353, opened 2026-09-17
feat: declarative labels for generated workload objects

### LINKED ISSUE #212: [Feature]: Allow labels on the TrainJob   created by the workload adapter
### Prerequisites

### Code of Conduct

### Feature Summary

TrainJobAdapter.Build()  sets only  Name  and  Namespace  on the TrainJob's ObjectMeta . There is no way for a user or the controller to place arbitrary labels on the TrainJob object CRE creates. Every external queueing  and scheduling system in the ecosystem identifies workloads by exactly such a label, so CRE currently cannot participate in any of them.

### Problem/Use Case

 Both of the major Layer-2 systems a GPU customer is likely to be running key off a label on the workload object:

   KAI Scheduler  │  kai.scheduler/queue        │ TrainJob metadata (docs[2])

  [1]: docs https://kueue.sigs.k8s.io/docs/tasks/run/trainjobs/
  [2]: docs [https://trainer.kubeflow.org/en/latest/operator-guides/job-schedu…](https://trainer.kubeflow.org/en/latest/operator-guides/job-scheduling/kai.html)

  CRE has neither seam:

    // pkg/workload/trainjob.go:44
    trainJob := &trainerv1alpha1.TrainJob{
        TypeMeta: metav1.TypeMeta{ ... },
        ObjectMeta: metav1.ObjectMeta{
            Name:      name,
            Namespace: namespace,     // ← no Labels, and no way to supply any
        },
        Spec: *spec.TrainJob,
    }

   InjectPodLabel()  ( pkg/workload/trainjob.go:63 ) exists but writes to the pod template metadata — one level too low for both systems.

  Concrete consequences today:

  • Kueue clusters: CRE's TrainJobs carry no queue name. Either Kueue ignores  them and they bypass quota entirely (a governance problem — a burn-in run consuming every GPU outside the quota system is what a platform team will veto), or, if the deployment is configured to capture unlabeled jobs, Kueue suspends them and they never run.
  • KAI clusters:  pkg/platform/runtime.go:121  writes  kai.scheduler/queue into the pod template labels rather than the TrainJob. Whether KAI's podgrouper still resolves the queue from pod level needs validating on a live cluster, but it is not the documented placement.
  • Anyone integrating Volcano, Yunikorn, or a custom admission controller hits the same wall.

### Proposed Solution

  Option A — extend the  Adapter  interface (preferred). Mirrors the existing  InjectPodLabel  /  SetNodeSelector  /  SetTolerations  mutator convention in  pkg/workload/adapter.go:32 :

    // SetWorkloadLabels sets labels on the workload object itself (not the
  pod
    // templates). External queueing systems — Kueue, KAI Scheduler, Volcano —
    // identify workloads by a label at this level.
    // Mutates spec in-place — call on a DeepCopy before Build.
    SetWorkloadLabels(spec *crev1alpha1.WorkloadSpec, labels
  map[string]string)

  This needs a field on  WorkloadSpec  (or on  TrainJobSpec  wrapper) to carry  them through to  Build() .

  Option B — a  metadata.labels  passthrough on the Job tier, surfaced up  through  Workflow.spec.jobTemplate  and  Certification , applied to the created workload object. Slightly more plumbing but gives users a  declarative path without a new adapter method.

  Either way the labels must be settable from:

  •  WorkloadRun  (already has  spec.gangScheduler , so this completes that story)
  •  Certification  /  Workflow  catalog path (currently has no route at all)

  ### Acceptance Criteria

  [ ] A user can set arbitrary labels on the TrainJob object created by a  WorkloadRun .
  [ ] A user can do the same for a  Certification  /  Workflow -driven run.
  [ ]  kueue.x-k8s.io/queue-name  and  kai.scheduler/queue  both land on the  TrainJob's  metadata.labels , verified in a unit test asserting against the built object.
  [ ] Existing pod-level label injection ( InjectPodLabel ) is unchanged —  this is additive.
  [ ] Integration test case covering the rendered TrainJob metadata.
  [ ]  docs/api-reference/  updated.

  ### References

  •  pkg/workload/trainjob.go:44  —  Build()
  •  pkg/workload/trainjob.go:63  —  InjectPodLabel() , pod-level
  •  pkg/workload/adapter.go:32  —  Adapter  interface and mutator convention
  •  pkg/platform/runtime.go:116  —  applyGangScheduler() , current KAI label placement

### Component

API / CRDs

kaynetu:
## Summary

NVCRE's Job controller creates a framework workload object from `spec.workload`, but nothing could label that object. The gap mattered because those labels are read *before* any workload pod exists: Kueue selects a local queue from `kueue.x-k8s.io/queue-name` on the TrainJob's `metadata.labels`, and Kubeflow's KAI Scheduler guide documents `kai.scheduler/queue` in the same place. ADR-076 wrote the queue label to the runtime Job and pod templates but never to the TrainJob itself, so NVCRE could not produce KAI's documented placement, Kueue's required placement, or arbitrary labels for any other integration keyed on the submitted object.

This implements ADR-079 (#337) and flips it Proposed → Accepted.

- add `workloadMetadata.labels` with `JobSpec` as the canonical boundary, surfaced on `WorkloadRun` and on Certification's `CategoryOptions` — which, because `CertificationSpec` inlines it, is read both globally at `spec.workloadMetadata` and per category at `spec.categories[].options.workloadMetadata`, with no second top-level field
- compose additively and deterministically: resolved catalog JobSpec, then the owner's labels replacing by key, then the gang-scheduler queue under an insert/same/conflict rule. Maps are cloned before mutation, so composing never writes back into a catalog template or an API object. `ResolveOptions` merges workload labels per key rather than replacing wholesale, unlike `thresholds`/`resources`/`imagePullSecrets` beside it
- apply labels generically after `Adapter.Build` in a shared `workload.BuildObject`, used by both Job reconciliation and dry-run rendering. Every workload kind exposes metadata identically through `client.Object`, so the Adapter interface gains no `SetWorkloadLabels` method and future adapters need no label-specific code. Adapter-produced labels with unrelated keys survive; an identical value is idempotent, a conflicting one fails
- derive the TrainJob queue label from `gangScheduler`, and persist the resolved intent on the new immutable `WorkflowSpec.gangScheduler`. The Workflow controller checks the effective manifests against it after overrides resolve and before dependencies or Jobs are created, and again before each later group or iteration. The two metadata levels are treated differently on purpose: the workload object's queue label is NVCRE's to write, so a removed one is restored, while a runtime label or pod scheduler name that disagrees fails and is **never** silently repaired
- keep `workloadMetadata` and runtime-patch label maps out of per-job dependency renaming, which rewrites every quoted occurrence of a dependency name and would otherwise rewrite a queue value that happens to equal one. Scheduling fields are *preserved* rather than reapplied, so an operator's conflicting override survives to be rejected instead of being overwritten
- replace Certification's three duplicated post-resolve call-site sequences with the named `ApplyResolvedWorkflowTransforms` stage ADR-077 anticipated, with its mutation order pinned
- enforce reserved keys (`app.kubernetes.io/managed-by`, `nvcre.nvidia.com/`), Kubernetes label syntax, and a 32-entry cap at CRD admission, mirrored in Go for offline `nvcrectl render` and for composed maps that no single admission rule ever sees whole
- make workload metadata immutable in presence *and* value by pairing a parent-level `has(self.f) == has(oldSelf.f)` transition rule with field-level equality — a rule scoped to an optional field does not run when that field is added or removed, so equality alone would let a Job drop the field and re-add it under a different queue

One correction to the ADR's own testing plan: it called for "runtime patches changing launcher/worker labels **or scheduler names**". Kubeflow's `PodSpecPatch` cannot reach `schedulerName`, so a runtime patch cannot change it. The plan is corrected in this PR and the scheduler-name cases assert against the dependency payload instead, which is the effective value.

Verified with `make ci` on the committed tree, plus the offline/controller parity paths by hand: a WorkloadRun with `gangScheduler` and no `workloadMetadata` renders the resolved queue on the TrainJob and the persisted intent on the Workflow; a same-key/different-value spec fails at construction naming the key, both values, and the field path; and a JSON-patch override removing the queue label comes back restored on the created Job.

## Related Issue

Closes #212

## Type of Change
- [ ] 🐛 Bug fix
- [x] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [x] API / CRDs
- [ ] Catalog / Workloads
- [x] CLI (nvcrectl)
- [ ] Helm / Deployment
- [x] Documentation / CI
- [ ] Other: ____________

## Testing
- [x] Tests pass locally

`make ci` green: `verify` (codegen, mod, license headers, 296 doc links), `lint` (0 issues), `build`, and the full unit + integration suites.

173 new golden test cases, following the repository's `testutil.TestCaseParser` convention:

| Area | Cases | Covers |
|---|---|---|
| `pkg/platform/validate-resolved-gang` | 25 | post-override checks: runtime Job/pod labels, launcher/worker runtime patches, scheduler name changed and missing, redirected `runtimeRef`, restore vs reject |
| `cmd/integration/immutability` | 24 | CRD transition rules against a real API server, for direct Job, Workflow Job template, Workflow `gangScheduler`, WorkloadRun, and Certification |
| `pkg/workloadrun/workload-metadata` | 21 | controller and CLI construction, plus the override matrix against a persisted queue-A intent |
| `cmd/integration/validation` | 21 | admission: 32/33 labels, 63/64-char values, 253/254-char key prefixes, 63/64-char key names, malformed keys and values, reserved keys |
| `cmd/integration/reconcile` | 17 | end-to-end: TrainJob labels, checkpoint restart, terminal conflicts on discovery and later groups, label restore on the created Job, dependency-rename survival |
| `pkg/workload` (`label-composition`, `build-object`) | 27 | map helpers and the shared builder, including clone semantics and adapter-label collisions |
| others | 38 | transform stage, dry-run parity, Certification render, scheduling-field preservation |

All four generated CRDs install in envtest, which is what proves the CEL rules fit the per-rule and whole-schema cost budgets — including Certification's nested up-to-64-category path.

Manual validation on a local Kind + KWOK cluster covered the five user-visible paths: labels reaching the TrainJob (and *not* the pods), the derived queue label, construction-time conflict, a terminal `GangSchedulingConflict` with no Job created, and both immutability rejections.

## Risk Assessment

Medium. All new fields are optional and manifests without `workloadMetadata` or `gangScheduler` render byte-identically. Two intended behavior changes affect existing users:

1. **A configured `gangScheduler` now also places its queue label on the TrainJob**, in addition to ADR-076's runtime Job- and pod-template placement. This aligns the submitted object with Kubeflow's documented KAI placement. Manifests setting `gangScheduler` and no `workloadMetadata` gain the label with the explicit `queue`, or `default-queue` when omitted.
2. **Overrides that previously redirected a runtime queue or pod scheduler name now fail** the consistency check rather than silently producing a gang the scheduler will never hold. To run against a different queue, configure it via `gangScheduler` on a new owning resource.

Both are documented in ADR-079's Consequences and in the api-reference pages, with accepted/conflicting override examples. Direct Jobs and Workflows without `gangScheduler` receive label validation and passthrough only — no inferred queue checks — and keep responsibility for their own scheduling configuration.

Retrofitting presence immutability onto the existing `nodeHealthMonitor`, `goodputMeasurement`, and `bandwidthMeasurement` fields is deliberately out of scope and tracked by #347.

## Checklist
- [x] Self-review completed
- [x] Documentation updated (if needed) — `docs/api-reference/{job,workflow,workloadrun,certification}.md`
- [x] Ready for review


asivanadi0 (approved):
## First skim (ADR-079 workload labels) — tip `18eac57`

CI green on this SHA. Skim focused on `JobSpec` as the canonical boundary plus Kueue / KAI (and Run:ai) queue-label behavior. Approving.

### JobSpec boundary

- `WorkloadMetadata.labels` on `JobSpec` is the right convergence point: WorkloadRun / Certification resolve into it; hand-authored Workflows write `jobTemplate.spec.workloadMetadata` directly; Job reconciliation + dry-run both go through `workload.BuildObject` after `Adapter.Build`.
- Presence + value immutability pairing (`has(self.workloadMetadata) == has(oldSelf.workloadMetadata)` plus field-level `self == oldSelf`) correctly closes the drop-and-readd hole that field equality alone would leave for an optional field.
- Reserved keys (`app.kubernetes.io/managed-by`, `nvcre.nvidia.com/`), Kubernetes key/value syntax, and the 32-entry cap are enforced at CRD admission and mirrored in `ValidateLabels` for offline render and post-compose maps no single admission rule ever sees whole. Clone-before-mutate in `MergeLabels` / `InsertConsistentLabel` / `LabelsOf` is load-bearing and tested.

### Kueue / KAI / Run:ai

- **Kueue is explicit-only:** `kueue.x-k8s.io/queue-name` via `workloadMetadata.labels`; never inferred from `gangScheduler`. Golden `kueue-queue-label-only` pins that, with `gangScheduler: null`.
- **KAI derived from `gangScheduler`:** default key `kai.scheduler/queue`, omitted queue → `default-queue`, custom key works for Run:ai (`runai/queue`). Intent persisted on immutable `WorkflowSpec.gangScheduler` with defaults filled in.
- **Compose is additive across keys:** `kueue-and-gang-scheduler-compose` places both labels on the same TrainJob. Same-key insert/same/conflict is correct (`explicit-queue-matching-gang-is-idempotent` vs `…-conflicting-…-fails-at-construction`).
- **Asymmetry is intentional and documented:** NVCRE-owned TrainJob queue label is restored when removed; runtime Job/pod queue labels and `schedulerName` fail without repair. `PreserveSchedulingFields` preserves rather than re-applies through dependency rename so a redirected queue survives to be rejected. Labels land on the workload object via `client.Object.SetLabels`, not as pod-label injection.

### Also checked

- `ResolveOptions` merges workload labels per key (unlike wholesale replace of thresholds/resources beside it) and detaches before write so the Certification global map is not aliased.
- `ApplyResolvedWorkflowTransforms` pins order with workload metadata last; `18eac57` closes the offline `ResolveWorkflowForPlatform` hole so render cannot emit a Workflow whose persisted intent and resolved runtime disagree.
- Testing-plan correction (PodSpecPatch cannot reach `schedulerName`) matches Kubeflow reality; dependency-payload cases cover that seam instead.

### Residual (non-blocking)

- Behavior change for existing `gangScheduler` users (TrainJob now also gets the queue label; conflicting runtime redirects fail instead of silently shipping) is real but called out in ADR Consequences / Risk Assessment with the right escape hatch (reconfigure via `gangScheduler` on a new owner).
- Production paths look covered; I did not re-run the full suite locally on this skim — relying on green CI + the golden matrix.

No blocker findings on JobSpec placement or Kueue/KAI label semantics. Nice ADR fidelity.

pkg/controller/workflow_deps.go
```diff
@@ -380,59 +382,149 @@ func buildReplacementMap(deps []nvcrev1alpha1.DependencySpec, suffix string) map
...
+
 	// Marshal job spec for classification
 	jobSpecJSON, err := json.Marshal(spec)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to marshal job spec: %w", err)
+		return preparedJob{}, fmt.Errorf("failed to marshal job spec: %w", err)
 	}
 
-	_, jobDeps := classifyDependencies(workflow.Spec.Dependencies, jobSpecJSON)
+	workflowDeps, jobDeps := classifyDependencies(workflow.Spec.Dependencies, jobSpecJSON)
```
ndipebot, line 580: `classifyDependencies` is fed a `JobSpec` that still contains `workloadMetadata`, but `suffixJobSpec` deliberately excludes that field from renaming (`:412-414`, restored at `:425`). Those two need to agree, and right now they do not.

The seed loop in `classifyDependencies` promotes any dependency whose `metadata.name` appears as a string anywhere in the marshalled `JobSpec`, with no key filter (the `isResourceName` gate applies to `sharedRefIndex`, not to that loop). So a user label value that happens to equal a dependency object name, for example `nccl-all-reduce-runtime`, promotes that dependency to job scope. `ensureWorkflowDependencies` then never creates it under its own name, a per-group suffixed copy is created instead, and the submitted `TrainJob` references an object that does not exist.

The comment at `:402` states the opposite of what the code now does: "Nothing in it ever references a dependency". That was true before `workloadMetadata` existed.

Same pattern at `pkg/controller/workflow_controller.go:3067` and `:3071`.

This is not gated by RBAC. The misclassified object can be any dependency, and the manager role covers `trainer.kubeflow.org`, `configmaps`, and `resource.nvidia.com`, which is what the catalog dependencies actually are.

Fix is small: marshal a copy with `WorkloadMetadata` set to nil for classification at both sites, matching what `suffixJobSpec` already does for renaming. Worth a `classifyDependencies` testdata case pinning a label value equal to a dependency name.

The trigger is narrow, it needs an exact string match against a dependency object name, so I am not calling it a merge blocker. But the asymmetry is real and the fix is two lines.

pkg/platform/transform.go
```diff
@@ -0,0 +1,485 @@
...
+// check behind gangScheduler, as ValidateResolvedGangScheduling alone does,
+// would let resolved offline output carry labels the cluster will reject.
+//
+// The gang-scheduling contract is then checked when an intent is persisted.
+func ValidateResolvedJobTemplate(
+	js *nvcrev1alpha1.JobSpec,
+	deps []nvcrev1alpha1.DependencySpec,
+	gs *nvcrev1alpha1.GangSchedulerSpec,
+) error {
+	if err := workload.ValidateLabels(
```
ndipebot, line 192: This label check runs first and unconditionally, so a pure label failure on a Workflow with no `gangScheduler` at all still returns through the `gangSchedulingConflictError` path. `pkg/controller/workflow_controller.go:1061` attributes every `ValidateResolvedJobTemplate` error to `ReasonGangSchedulingConflict`, and the doc comment on that constant at `pkg/controller/helpers.go:106-110` says it means an override redirected a runtime queue label or a pod scheduler name. For a label error neither of those happened.

It is reachable. `OverrideSpec.JobTemplate` is `Schemaless` with `PreserveUnknownFields`, so CRD CEL never validates its contents, and `mergeJobTemplate` at `pkg/controller/workflow_detect.go:616-629` zeroes the struct and unmarshals the merged result wholesale. An override that writes an invalid `workloadMetadata.labels` value lands here with no gang scheduler in play.

Either give the label failure its own reason, or widen the constant's doc comment to cover resolved workload metadata generally.

api/v1alpha1/workloadrun_types.go
```diff
@@ -302,8 +302,22 @@ type WorkloadRunSpec struct {
...
+	// creates (today a Kubeflow TrainJob). Use it to place the workload in a
+	// Kueue local queue with kueue.x-k8s.io/queue-name, or for any other
+	// integration keyed on the submitted object's labels. It is copied into
+	// the generated Workflow's spec.jobTemplate.spec.workloadMetadata.
+	//
+	// These labels are not pod labels. When gangScheduler is also set, its
+	// resolved queue label is added here too; supplying the same key with a
+	// different value is a conflict and fails.
+	// +optional
+	WorkloadMetadata *WorkloadMetadata `json:"workloadMetadata,omitempty"`
```
ndipebot, line 320: `workloadMetadata.labels` rejects `app.kubernetes.io/managed-by` and anything under `nvcre.nvidia.com/` at admission, which is right.

`gangScheduler.queueLabelKey` at `:354` has no matching rule. It validates `MaxLength=317`, the qualified-name pattern, and the 253-character prefix, but nothing stops `nvcre.nvidia.com/queue`.

A key like that is admitted, then fails on every reconcile at composition time, when the queue label is inserted into `workloadMetadata.labels` and `ValidateLabels` rejects it as reserved. Since the spec is immutable the user cannot edit their way out, they have to delete and recreate.

Adding the same reserved-key `XValidation` to `queueLabelKey` turns that into an admission error that names the problem at submit time.

pkg/platform/transform.go
```diff
@@ -0,0 +1,485 @@
...
+	trainerv1alpha1 "github.com/kubeflow/trainer/v2/pkg/apis/trainer/v1alpha1"
+
+	nvcrev1alpha1 "github.com/NVIDIA/cluster-readiness-engine/api/v1alpha1"
+	"github.com/NVIDIA/cluster-readiness-engine/pkg/workload"
+)
+
+// jobTemplateLabelsPath names the resolved workload metadata in errors. Every
+// caller of ApplyResolvedWorkflowTransforms is writing this same field on a
+// resolved WorkflowSpec, whatever higher-level API the labels came from.
+const jobTemplateLabelsPath = "spec.jobTemplate.spec.workloadMetadata.labels"
```
ndipebot: This is a package constant used unconditionally, so a WorkloadRun failure reports `spec.jobTemplate.spec.workloadMetadata.labels`. A WorkloadRun has no `spec.jobTemplate`. The user wrote `spec.workloadMetadata.labels` or `spec.gangScheduler.queue` and gets pointed at a path that does not exist in the API they used.

It is reached from `pkg/workloadrun/workloadrun.go:393`, and the wrong path is already pinned into a persisted condition message in `cmd/integration/testdata/reconcile/workloadrun-gangscheduler-queue-conflict-fails/expected.json:53`, so it is user-visible in status, not just in a log line.

Threading the caller's own path in, the way `workload.ValidateLabels` already takes `fieldPath`, would fix it. The golden file would need regenerating.

docs/api-reference/workloadrun.md
```diff
@@ -55,10 +55,83 @@ _Fields documented so far:_
...
 | `gangScheduler.queueLabelKey` | string | Optional. Label key the queue value is written under; defaults to `kai.scheduler/queue` when unset. Set it to `runai/queue` on a Run:ai cluster, which reads that label and ignores `kai.scheduler/queue`. When non-empty, must be a valid Kubernetes label key (qualified name): an optional DNS-subdomain prefix of at most 253 characters followed by `/`, then a name of at most 63 characters; at most 317 characters in total |
+| `workloadMetadata` | WorkloadMetadata | Optional. Labels applied to the generated workload object (the `TrainJob`), copied into the generated Workflow's `spec.jobTemplate.spec.workloadMetadata`. Use it for Kueue (`kueue.x-k8s.io/queue-name`) or any other integration keyed on the submitted object's labels. See [Workload object labels](#workload-object-labels) |
+| `workloadMetadata.labels` | map[string]string | Optional. At most 32 entries. Keys must be valid Kubernetes label keys and values valid Kubernetes label values; `app.kubernetes.io/managed-by` and any key under `nvcre.nvidia.com/` are rejected |
```
ndipebot: "At most 32 entries" reads as a flat cap, but when `gangScheduler` is set the resolved queue label is inserted into this same map and counts against `MaxLabels`, so the user's own budget is 31, and exceeding the cap fails rather than dropping a label.

`docs/api-reference/certification.md:128` already states both of those qualifications. Matching that wording here would keep the two pages consistent.

`docs/api-reference/job.md` is accurate as written and should stay as is, since the Job tier has no `gangScheduler` to insert anything.

ndipebot (commented):
I went through this closely.

The design holds up. The paired presence plus equality `XValidation` rules are the right way to make an optional field immutable, since a transition rule scoped to an optional field does not run when that field is added or removed. `BuildObject` as the single construction seam shared by Job reconciliation and dry-run rendering is the right call, and I agree with `PreserveSchedulingFields` preserving rather than repairing so an override that redirects a queue surfaces as a rejection instead of being silently overwritten.

Five things below, left inline where the code is. Only the first is a behavior bug; the rest are polish.

pkg/controller/workflow_deps.go
kaynetu: Fixed in 4cd84fe. Both call sites now go through `marshalJobSpecForDependencyClassification`, which nils `workloadMetadata` on a copy before marshalling.

Extended it to the runtime patch label and annotation maps as well. `suffixJobSpec` protects those for the same reason, so the same misclassification was reachable through a `runtimePatch` label. Real references inside a patch stay visible — a PVC `claimName` is still classified on.

Two testdata cases pin it. `workload-metadata-value-is-not-reference` covers the case you described. `runtime-patch-metadata-values-are-not-references` puts ConfigMap names in metadata at all three nesting levels and asserts they stay workflow-scoped, while `actual-pvc` in the same patch stays job-scoped.

The test harness was rewired to call the filtered marshal. Calling `classifyDependencies` directly would have made both goldens pass vacuously.

pkg/platform/transform.go
kaynetu: Took the second option — widened the constant's doc comment.

`ReasonGangSchedulingConflict` now says it marks a resolved Job template that is invalid after overrides. That explicitly includes workload metadata failing label validation with no gang scheduler configured.

A dedicated reason would be more precise. It would also add a second terminal reason to the Workflow status contract for a failure that is already fully described by the error message and field path. Happy to split it if you would rather have the reason carry that distinction.

api/v1alpha1/workloadrun_types.go
kaynetu: Added. `queueLabelKey` now carries the same reserved-key `XValidation` as `workloadMetadata.labels`.

Regenerated into the certifications, workflows, and workloadruns CRDs — the three that embed `GangSchedulerSpec`. `make manifests generate` produces no other drift.

New admission case `validation/workloadrun-gangscheduler-queue-label-key-reserved` pins that `nvcre.nvidia.com/queue` is rejected at submit time with the field named.

Documented on both API reference pages and both how-to guides.

pkg/platform/transform.go
kaynetu: Fixed. `ValidateResolvedJobTemplate` and `ValidateResolvedGangScheduling` now take the field path from the caller, the way `workload.ValidateLabels` already did.

Two exported constants: `JobTemplateWorkloadLabelsPath` and `WorkloadRunWorkloadLabelsPath`. Every call site was updated, including the ones in `pkg/render` and `pkg/workloadrun`.

The golden you pointed at now reads `spec.workloadMetadata.labels["kai.scheduler/queue"]`. Six `pkg/workloadrun` testdata goldens changed the same way.

docs/api-reference/workloadrun.md
kaynetu: Matched to the `certification.md` wording. The row now states the cap applies after merging and after the queue label is inserted, and that exceeding it fails rather than dropping labels.

Also spelled out that the queue label consumes one entry, so the user budget is 31 when `gangScheduler` is set.

Left `job.md` unchanged, as you said — the Job tier has no `gangScheduler`.

Voyagerroc-Lab (commented):
Reviewed the design/validation surface for declarative workload labels: immutable intent, deterministic per-key composition, reserved-key and Kubernetes syntax limits, shared post-Adapter application, dry-run/reconcile parity, and gang-scheduler conflict checks. The PR description also calls out the 300+ file/generated scope and the 173 golden cases; given the diff-size limit, I am recording this as a design-level review rather than an approval. The stated invariants and failure-before-Job behavior are coherent.

pkg/controller/workflow_deps.go
ndipebot: Fixed in `4cd84fea`, and more thoroughly than I flagged. `marshalJobSpecForDependencyClassification` is now used at both call sites, and `clearRuntimePatchMetadata` also strips runtime patch labels and annotations, which `suffixJobSpec` likewise keeps out of renaming. I had only caught the `workloadMetadata` half. The stale comment is gone too.

I like that `clearRuntimePatchMetadata` is scoped to just the label and annotation maps, so a real reference inside a patch still seeds classification. The `runtime-patch-metadata-values-are-not-references` case pins that with `actual-pvc`, which is the negative control I would have asked for.

pkg/platform/transform.go
ndipebot: Resolved in `4cd84fea` by widening the doc comment on `ReasonGangSchedulingConflict`, which now says it covers workload metadata failing label validation with no gang scheduler configured. That was the option I would have picked. One terminal reason with accurate docs beats threading a second reason through `launchPendingGroups`.

api/v1alpha1/workloadrun_types.go
ndipebot: Fixed in `4cd84fea`. `queueLabelKey` now carries the reserved-key `XValidation`, and `workloadrun-gangscheduler-queue-label-key-reserved` pins the admission rejection.

pkg/platform/transform.go
ndipebot: Fixed in `4cd84fea`. Split into `JobTemplateWorkloadLabelsPath` and `WorkloadRunWorkloadLabelsPath`, threaded through as a parameter. I checked all seven callers pass a path and that both WorkloadRun sites pass the WorkloadRun one. The persisted condition now reads `spec.workloadMetadata.labels`.

docs/api-reference/workloadrun.md
ndipebot: Fixed in `4cd84fea`. The row now states both the cap and the 31-label user budget, and `job.md` was correctly left alone.

ndipebot (approved):
All five of my findings are addressed in `4cd84fea`. I confirmed each in the code at `68aa16d9` rather than going off the commit message, and left the details on the original threads.

Verified locally at `68aa16d9`:

1. `go build ./...` clean.
2. `pkg/controller`, `pkg/platform`, `pkg/workload`, `pkg/workloadrun`, `pkg/render` all pass.
3. `go test ./cmd/integration/` passes.

The dependency classification fix went wider than what I raised, catching runtime patch metadata as well, and the new `classify-dependencies` cases pin both halves with a real reference left in as a control.

PR #358, opened 2026-09-20
feat: implement support for cordoned node selection

natherz97:
## Summary

This PR implements [ADR-081: Support for Cordoned Node Selection](https://github.com/NVIDIA/cluster-readiness-engine/blob/main/docs/designs/081-cordoned-node-selection.md). A Certification can opt-in to targeting cordoned nodes by adding a taintSelector entry for the node.kubernetes.io/unschedulable taint. The following 2 changes are implemented in this PR:

1. Node discovery: discoverTargetNodes will not exclude cordoned nodes that matched a taintSelectors entry for the node.kubernetes.io/unschedulable taint. If this taint is not present in taintSelectors, the existing behavior will apply which filters out cordoned nodes from the set of selected nodes.
2. Health monitor: if taintSelectors specifies the node.kubernetes.io/unschedulable taint, the cordon health check will be dropped for the run. This means clearing nodeHealthMonitor on the created Job even when the catalog jobTemplate sets it, not only skipping the nil default, since every catalog entry sets the node.spec.unschedulable == true expression explicitly. If this taint is not present in taintSelectors, the default monitor CEL expression will continue to be applied.

## Testing

The following test cases were added for cordoned node support:

TestTargetsCordonedNodes:
- nil-target: a nil target does not opt into cordoned-node targeting.
- empty-taint-selectors:  an empty taintSelectors list does not opt into cordoned-node targeting.
- unrelated-taint-key:  a taintSelectors entry for an unrelated key does not opt into cordoned-node targeting.
- unschedulable-taint-key: a taintSelectors entry for node.kubernetes.io/unschedulable with no effect set opts in.
- unschedulable-taint-key-with-effect:  a taintSelectors entry for node.kubernetes.io/unschedulable with an effect opts in.

TestDiscoverCordonedNodes:
- cordoned-node-opted-in: a cordoned node opted in via taintSelectors is included
- cordoned-node-not-opted-in-with-taint: the same cordoned, tainted node is excluded when there is no taintSelectors opt-in.

TestIntegration:
- cordoned-node-selected: a cordoned node opted in via taintSelectors gets a Job created against it with no nodeHealthMonitor.
- cordoned-node-selected-explicit-monitor: a cordoned-node opt-in clears nodeHealthMonitor even when jobTemplate sets it explicitly.
- workflow-explicit-health-monitor-preserved: a Workflow without the cordoned-node opt-in keeps its own explicit nodeHealthMonitor.

TestAWSH100CordonedNode
- aws/h100/cordoned-node:  a Certification can succeed against a cordoned node targeted via a taintSelectors entry for the unschedulable taint.

## Type of Change
- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation
- [ ] 🔧 Refactoring
- [ ] 🔨 Build/CI

## Component(s) Affected
- [ ] API / CRDs
- [X] Controller / Reconcilers
- [ ] Catalog / Workloads
- [ ] CLI (nvcrectl)
- [ ] Helm / Deployment
- [ ] Documentation / CI
- [ ] Other: ____________

## Testing
- [X] Tests pass locally
- [X] Manual testing completed
- [X] No breaking changes (or documented)

## Checklist
- [X] Self-review completed
- [X] Commits are signed off for the DCO (`git commit -s`)
- [X] `make manifests generate` run (if `*_types.go` was modified)
- [X] Golden files updated (if integration test output changed)
- [X] Documentation updated (if needed)
- [X] Ready for review


asivanadi0 (approved):
## Review (cordoned-node selection — ADR-081 implementation)

Tip reviewed: `500a75946473f122cacc8d6dcca55af18922742c`. First human review on this tip, after #359 in the same pass. CI green on this SHA (Lint, Test, Build, UAT Kind+KWOK, CodeQL, Docker Build, Verify, govulncheck, DCO). Approving.

### Context vs #348

Prior COMMENT on #348 blocked on decision-item 3: skipping the nil default alone left every catalog `jobTemplate`’s explicit `node.spec.unschedulable == true` in place, so Certification still HardwareFailureDetected on cordoned nodes. Merged ADR-081 now says clear `nodeHealthMonitor` on the created Job even when the catalog/Workflow set it. This PR implements that wording, not the earlier incomplete skip-default sketch.

### What lines up well

- **Discovery gate.** `discoverTargetNodes` only skips `Spec.Unschedulable` when `!targetsCordonedNodes(target)`. Opt-in is key-match on `corev1.TaintNodeUnschedulable` (effect optional), which matches the ADR example and the two valid selector shapes. With `taintSelectors` AND-matching, opt-in also narrows the run to cordoned nodes — the filter/mode-switch consequence called out on #348 is preserved in behavior.
- **Health monitor.** `createJobForGroup` clears `job.Spec.NodeHealthMonitor` on opt-in, else applies the default only when nil. That is the load-bearing fix for decision-item 3: catalog copies no longer keep the unschedulable CEL. Integration goldens cover (a) opt-in clears catalog-style explicit monitor, (b) non-opt-in preserves a custom non-cordon CEL — the right pair.
- **Tests.** `TestTargetsCordonedNodes` (nil / empty / unrelated / key / key+effect), `TestDiscoverCordonedNodes` (opted-in vs not), reconcile goldens, and `TestAWSH100CordonedNode` UAT (schedule onto cordoned node → Succeeded) are proportionate for an ADR-shaped feature.
- **Tolerations.** No new path; existing `adapter.SetTolerations` from `taintSelectors` is what the ADR relies on. Expected Job shows the unschedulable Exists/NoSchedule toleration under opt-in.

### Residual (non-blocking — ADR-documented)

1. **Clear-all vs strip-cordon-clause.** Opt-in nils the entire monitor, so a Workflow that both targets cordoned nodes *and* carries a custom non-cordon CEL (e.g. `gpu.unhealthy`) loses the custom check for that run. ADR Consequence 2 states this explicitly; Alternative 2 (Certification-level monitor override / selective strip) was punted. CodeRabbit’s “preserve explicit custom monitors” suggestion would be a follow-on design change, not a miss against ADR-081 as Accepted. Calling it out so it is not re-litigated as an implementation bug.
2. **Cordon race without taint.** A node with `Spec.Unschedulable` but not yet carrying `node.kubernetes.io/unschedulable` is still excluded under opt-in via `nodeMatchesTaints`. ADR Consequence 4 assigns that to the client; acceptable for v1.

No merge blockers on this tip.

kaynetu:
Reviewed `500a759`. The implementation follows ADR-081 cleanly: the same opt-in helper controls discovery and health-monitor removal, matching tolerations are propagated through the existing adapter path, and default cordoned-node exclusion remains covered. The explicit-monitor integration case also catches the important catalog-monitor path. Clearing all node-health monitoring is an explicit ADR decision, so I would not treat preservation of custom monitors as an implementation requirement here.

Locally, the controller and render package tests and all three new integration cases passed. GitHub CI also reports UAT success; I did not rerun Kind/KWOK locally. I found three non-blocking follow-ups around the waiting message, dry-run parity, and API documentation, detailed below.

1. Non-blocking: the [waiting message](https://github.com/NVIDIA/cluster-readiness-engine/blob/500a75946473f122cacc8d6dcca55af18922742c/pkg/controller/certification_controller.go#L47-L50) contradicts cordoned-node opt-in. With the unschedulable taint selector, matching cordoned nodes do count, but an empty discovery result still says “Nodes that are cordoned or unschedulable do not count.” This can happen while waiting for the required taint to appear. Please branch on `targetsCordonedNodes(&cert.Spec.Target)` and adjust both “No schedulable nodes” and the exclusion sentence.

2. Non-blocking: please apply the cordoned-node opt-in logic in [DryRunCreate](https://github.com/NVIDIA/cluster-readiness-engine/blob/500a75946473f122cacc8d6dcca55af18922742c/pkg/render/render.go#L392-L399) too. Shared discovery now admits opted-in cordoned nodes, but dry-run still retains—or injects—the `node.spec.unschedulable == true` monitor that the controller clears. I reproduced this using the PR's explicit-monitor fixture. Consequently, `--dry-run` validates a different Job configuration from the one actually created, which can produce different admission-policy results. A parity test would cover this gap.

3. Non-blocking: please document `target.taintSelectors` in the [Certification API reference](https://github.com/NVIDIA/cluster-readiness-engine/blob/500a75946473f122cacc8d6dcca55af18922742c/docs/api-reference/certification.md#L58), including the `node.kubernetes.io/unschedulable` opt-in and its side effect: it clears all node-health monitoring for the run, not just the cordon check. The concept and getting-started pages mention the feature, but the API reference omits this behavior.

PR #359, opened 2026-09-20
test: exercise release ref guards and follow attest wrappers

### LINKED ISSUE #351: test(releasepolicy): attest guard recognizers can be fooled by a commented-out exit and by a local wrapper
Follow-up to #341. That PR closes the real trust gap (`workflow_dispatch` at a `v*` tag minting the release signing identity through `attest.yml`), and I approved it. While mutation-testing the new guards at `c76e227` I found two places where the *tests* are weaker than their names suggest. Neither is an open hole in the tree today, so I did not hold up the merge, but both would let a future refactor reintroduce the class of bug #341 exists to prevent.

## 1. `runHasEffectiveReleaseTagRefCheck` is substring-based, so a neutered guard can still register as present

`test/releasepolicy/attest_guards_test.go` locates `exactReleaseTagRefCheck` in the `Resolve tag` step of `release.yml`, bounds the then-branch at the first ` fi`, and asks whether `exit 1` appears in that window. Appearing in the text is not the same as running.

I probed the helper directly against three mutations of the real step:

| Mutation of the `exit 1` in `Resolve tag` | `runHasEffectiveReleaseTagRefCheck` | execution table |
|---|---|---|
| replaced with `:` (the case @kaynetu raised) | rejects | rejects |
| replaced with `: # exit 1` | **accepts** | rejects, `attest_guards_test.go:824` |
| moved into an `else` branch | **accepts** | rejects, `attest_guards_test.go:813` |

Net protection holds, because `TestReleaseTagRefCheckFailsClosed` actually executes the step against accept env (`GITHUB_REF=refs/tags/v1.2.3`, `INPUT_TAG=v1.2.3`) and reject env (`GITHUB_REF=refs/heads/main`). The problem is that `TestAttestDispatchCallersRequireRefGuards` relies on the recognizer alone, so its verdict on a caller can be wrong even while the suite stays green overall.

What I would like:

1. Strip shell comments before matching, so a commented-out `exit 1` cannot satisfy the window.
2. Confirm the rejection is in the then-branch of the comparison rather than anywhere between the comparison and the first ` fi`.
3. Or drop the text matching entirely and have `TestAttestDispatchCallersRequireRefGuards` depend on the execution result, which is the check that cannot be fooled by rewriting the script.

## 2. `isAttestWorkflowCall` does not traverse a local wrapper

This is the CodeRabbit finding at https://github.com/NVIDIA/cluster-readiness-engine/pull/341#discussion_r4038927346, and it reproduces. The helper is basename matching on a single `uses`:

```go
uses = strings.SplitN(uses, "@", 2)[0]
return filepath.Base(uses) == attestWorkflowName
```

So a caller reaches `attest.yml` unguarded as long as it goes through one hop. I confirmed it with two probe files:

1. `probe-dispatch.yml`, a `workflow_dispatch` caller carrying only the repository gate and no ref guard, calling `uses: ./.github/workflows/probe-wrapper.yml`.
2. `probe-wrapper.yml`, a `workflow_call` with an unguarded `uses: ./.github/workflows/attest.yml`.

`TestAttestDispatchCallersRequireRefGuards` passes on that pair. It should fail.

No live exposure right now: `build-image.yml` is the only local workflow called by another local workflow, and it never reaches `attest.yml`. The fix is to resolve local `uses: ./.github/workflows/*.yml` targets recursively and treat a caller as an attest caller if any transitive callee is `attest.yml`, with the existing cycle guard extended to cover file-to-file edges.

## Why this is worth doing

The value of `test/releasepolicy/` is that it fails when someone weakens a guard, which is exactly the property I verified the rest of #341 has. These two helpers are the spots where that property does not yet hold.

MaxFreedomPollard:
## Summary

The dispatch policy can currently accept `: # exit 1` as a release-ref guard and miss an attest call hidden behind a local reusable workflow.

Exercise the release tag resolver with a matching tag, a branch ref and a different tag before crediting its shell guard. Follow local reusable-workflow calls when checking dispatch callers, with a visited set to stop cycles. Only the unconditional, failure-propagating `Resolve tag` step in `release.yml` can supply a shell guard.

Regression cases cover comments, quoted and unreachable exits, an exit moved to the wrong branch, guarded and unguarded wrappers, multi-hop calls, `.yaml` workflows and cycles. This changes tests only; it does not alter or dispatch a signing workflow.

## Related Issue

Closes #351

## Type of Change
- [x] 🐛 Bug fix
- [x] 🔨 Build/CI

## Component(s) Affected
- [x] Documentation / CI
- [x] Other: release-policy tests

## Testing
- [x] The new shell mutation cases reproduce four false positives with the old recognizer.
- [x] `go test ./test/releasepolicy -count=1`
- [x] `make ci` passes with Go 1.27.0, including lint, vet, build, unit tests and Kubernetes integration tests. Helm chart render tests were skipped because Helm is not installed locally.
- [x] No breaking changes.

## Checklist
- [x] Self-review completed
- [x] Ready for review


test/releasepolicy/attest_guards_test.go
```diff
@@ -563,42 +565,73 @@ func TestMainBranchAttestCallersPinExactRefGuards(t *testing.T) {
...
+// direct attest calls. Each file is visited once, including in cyclic graphs.
+func callsAttestThroughLocalWorkflows(workflows map[string]map[string]policyJob, uses string) bool {
+	seen := map[string]bool{}
+	var walk func(string) bool
+	walk = func(uses string) bool {
+		if isAttestWorkflowCall(uses) {
+			return true
+		}
+		const prefix = "./.github/workflows/"
+		if !strings.HasPrefix(uses, prefix) {
```
asivanadi0: **`$/` same-repo wrappers slip the walk.**

`callsAttestThroughLocalWorkflows` only accepts `./.github/workflows/`. GitHub’s recommended same-repo form is now `uses: $/.github/workflows/wrapper.yml` (resolves at the running commit; no checkout). Direct `$/.github/workflows/attest.yml` still hits `isAttestWorkflowCall` via basename, but a `$/` *wrapper* that then calls `attest.yml` returns false here, so `unguardedAttestCallers` omits an unguarded dispatch→wrapper→attest path — the same class of miss #351 item 2 closed for `./`.

Accept both prefixes, trim whichever matched, and add a `$/` row next to the existing one-wrapper / two-wrapper cases in `TestAttestWrapperReachability` / `TestWrappedAttestCallersRequireRefGuards`.

asivanadi0 (commented):
## Review (release-ref / attest wrapper recognition — #351)

Tip reviewed: `5b891e2c9fbb090bdb4cf21c161dd16d26676391`. First human design/code pass. Not approving yet.

### What lines up well

- **Execution over substring.** `runHasEffectiveReleaseTagRefCheck` now drives the allowlisted comparison through accept (`refs/tags/v1.2.3`) and reject (`refs/heads/main`, mismatched tag) via `runShell`, which is exactly the #351 option that cannot be fooled by `: # exit 1`, quoted `exit 1`, else-branch moves, or unreachable blocks. The mutation table in `TestReleaseTagRefRecognitionMutations` pins those four false positives the old window recognizer accepted.
- **Mandatory shell guard.** Restricting credited `Runs` to an unconditional, non-`continue-on-error` `Resolve tag` step on `release.yml` / `release-tag` closes the skipped-or-swallowed path; `TestLoadedReleaseTagGuardMustBeMandatory` covers it.
- **Wrapper reachability.** Recursive local walk with a per-file visited set, `.yaml` multi-hop, cycle-with/without-attest, and remote-vs-local negative cases are the right shape for #351 item 2. `unguardedAttestCallers` + the guarded / cancelled-bypass table matches how the live dispatch policy enumerates callers.
- **Scope.** Tests-only; no signing workflow change. Matches the PR claim.

### Blocking gap

1. **`$/` same-repo wrappers are not followed** (inline). GitHub now documents `$/.github/workflows/{file}` as the *recommended* same-repository reusable-workflow form (changelog 2026-07-30; docs: Reuse workflows). Direct `$/.github/workflows/attest.yml` is still caught by `isAttestWorkflowCall` (basename), but a `$/` hop to a wrapper that then calls `attest.yml` returns false and drops out of `unguardedAttestCallers` — same class of hole #351 item 2 asked to close for `./`. Accept both prefixes, normalize the basename, and add a regression case next to the existing `./` wrapper rows.

### CI note (non-blocking)

CI is thin on this tip: DCO + CodeRabbit only. `copy-pr-bot` is holding NVIDIA runner workflows pending vetting, so `go test ./test/releasepolicy` / `make ci` have not run on GH Actions for this SHA. Author reports local green; that is fine to note, not a substitute for runner CI once the PR is vetted.

### Residual (after the `$/` fix)

- Optional: strip a trailing `@ref` before the local-prefix walk the same way `isAttestWorkflowCall` does, so a hypothetical `./.github/workflows/wrapper.yml@…` does not fail closed for the wrong reason. Lower priority than `$/`.

Happy to re-review / approve once `$/` is in the walker and covered by a regression case.

test/releasepolicy/attest_guards_test.go
MaxFreedomPollard: Fixed in `33b12c5`. The walker now accepts both `./` and `$/` and removes the matching prefix before looking up the workflow. Regression cases cover single and multiple `$/` wrappers, mixed prefixes, and guarded versus unguarded callers. Five of the new cases fail with the previous walker.

`go test ./test/releasepolicy -count=1` and the full local `make ci` pass, including lint, vet, build, unit tests and Kubernetes integration tests. Helm rendering tests were skipped because Helm is not installed. NVIDIA runner validation is still separate from these local results.

asivanadi0: Verified on tip `33b12c5`: walker accepts both `./` and `$/` via `CutPrefix`, trims the matched prefix, and the new regression rows cover single `$/`, multi-hop `$/`, mixed `./`→`$/`, and guarded vs unguarded callers. Confirmed five of those cases fail under the previous `./`-only walker. Closing this thread.

asivanadi0 (approved):
## Re-review (tip `33b12c5`)

Prior blocking gap on `$/` same-repo wrappers is **FIXED**. Verified on tip `33b12c52968bb4695da771e1e9fa578e938ceeea`:

- `callsAttestThroughLocalWorkflows` accepts both `./.github/workflows/` and `$/.github/workflows/`, trims whichever matched, then walks.
- Regressions cover single `$/`, two `$/` hops, mixed `./`→`$/`, and guarded vs unguarded callers in `TestAttestWrapperReachability` / `TestWrappedAttestCallersRequireRefGuards`.
- Local check: those five new cases fail under the previous `./`-only walker and pass on this tip; `go test ./test/releasepolicy -run 'TestAttestWrapperReachability|TestWrappedAttestCallersRequireRefGuards|…'` green.

Closes prior feedback from review on `5b891e2`. Approving.

test/releasepolicy/attest_guards_test.go
```diff
@@ -563,42 +565,75 @@ func TestMainBranchAttestCallersPinExactRefGuards(t *testing.T) {
...
+
+// unguardedAttestCallers returns direct and transitive attest callers whose
+// dispatch workflow does not gate them on a recognized ref restriction.
+func unguardedAttestCallers(
+	t *testing.T, workflows map[string]map[string]policyJob, jobs map[string]policyJob,
+) []string {
+	t.Helper()
+	var callers []string
+	for name, job := range jobs {
+		if callsAttestThroughLocalWorkflows(workflows, job.Uses) && !jobOrAncestorHasRefGuard(t, jobs, name) {
```
kaynetu: Non-blocking: guards inside wrappers are not credited here. On `33b12c5`, I reproduced dispatch → wrapper → attest with `exactRepoAndMainIf` on the wrapper's signing job; `unguardedAttestCallers` still returns the outer caller because guard recognition only checks the dispatch workflow's jobs. This rejects a safe configuration rather than exposing signing, but a safe refactor could fail CI. Could we document that the outer caller must carry the guard, or add coverage and recognition for guards inside wrappers?

MaxFreedomPollard: Added recognition in `f5ad2d3` rather than documenting the limit. A reusable workflow evaluates its jobs' `if:` against the caller's `github` context, so a recognized guard on the wrapper job that leads to `attest.yml`, or on one of that job's needs-ancestors, now counts. An unrelated guard, a `!cancelled()` bypass, or a second unguarded path inside the wrapper still fails closed.
Nine regression rows cover these cases, and walkers that ignore wrapper guards, or credit any guarded job in the wrapper, fail them. `go test ./test/releasepolicy` and golangci-lint v2.13.2 pass on the package.

asivanadi0 (approved):
## Re-review (tip `9295c3df`)

Prior APPROVE on `33b12c5` is stale after `f5ad2d3` + main merge. Re-checked this tip.

### What changed since `33b12c5`

- **`f5ad2d3` — credit guards on the attest path inside local wrappers.** Addresses kaynetu’s non-blocking note: a reusable workflow evaluates job `if:` against the caller’s `github` context, so a recognized guard on the wrapper job (or its needs-ancestors) that leads to `attest.yml` should count. `unguardedAttestCallers` now walks via `reachesAttest(..., skip)` and skips jobs that already have a recognized guard before following their `uses:`, so a guarded hop does not mark a shared wrapper visited and hide an open hop.
- Fail-closed still holds for unrelated guards, `!cancelled()` bypass, partly-guarded wrappers, and shared-wrapper mixed hops — covered by the nine new regression rows in `TestWrappedAttestCallersRequireRefGuards`.
- Tip also includes a main merge only; PR files remain the two `test/releasepolicy` files.

### Still sound from prior review

- Execution-based `runHasEffectiveReleaseTagRefCheck` + mandatory Resolve-tag credit.
- `$/` + `./` local wrapper walk with visited set / cycle handling.

### CI

Tip checks green (Lint, Build, Test, UAT, CodeQL Analyze Go, Docker Build, DCO, etc.; Fern/collect skipped as expected). Not behind `main`.

Approving this tip.

kaynetu (approved):

